PR #15, opened 2020-09-11
change the storage impl

ChrisEddy:


rig.js
```diff
@@ -60,26 +60,30 @@ class RigManager {
...
-    this.localRig = new Avatar(o, {
-      fingers: true,
-      hair: true,
-      visemes: true,
-      // decapitate: selectedTool === 'firstperson',
-    });
-    this.scene.add(this.localRig.model);
-
-    await this.localRigQueue.unlock();
+    try {
```
avaer, line 63: Why try?

rig.js
```diff
@@ -100,28 +104,32 @@ class RigManager {
...
-      fingers: true,
-      hair: true,
-      visemes: true,
-      // decapitate: selectedTool === 'firstperson',
-    });
-    this.scene.add(peerRig.model);
-    this.peerRigs.set(peerId, peerRig);
-
-    await this.peerRigQueue.unlock();
+    try {
```
avaer, line 107: Why try? If this is meant to catch fetch errors, it should be scoped tighter.

rig.js
ChrisEddy: URL can bounce and was crashing site, uncaught promise rejection

rig.js
ChrisEddy: URL can bounce and was crashing site, uncaught promise rejection

rig.js
avaer: Then we should scope this down to the actual problematic fetch, rather than all errors we might see randomly.

ChrisEddy:
@avaer fixed

avaer (approved):
Looks good!

PR #16, opened 2020-09-11
File type detection support

avaer:


ChrisEddy:
Is there an implementation that goes with this? or just adding the source?

avaer:
I was initially considering how to detect the file type coming in when the user uploads, but right now we're doing it by file name extension, which probably is good enough.

I'm not sure this PR makes sense at the moment so closing.

PR #17, opened 2020-09-11
Flow support

avaer:


blockchain.js
```diff
@@ -0,0 +1,626 @@
...
+};
+const _bakeContract = async (contractKeys, contractSource) => {
+  const acctResponse = await flow.sdk.send(await flow.sdk.pipe(await flow.sdk.build([
+    flow.sdk.getAccount(contractKeys.address),
+  ]), [
+    flow.sdk.resolve([
+      flow.sdk.resolveParams,
+    ]),
+  ]), { node: flowConstants.host });
+  const seqNum = acctResponse.account.keys[0].sequenceNumber;
```
ChrisEddy, line 51: may want to check for undefined along this object path

blockchain.js
```diff
@@ -0,0 +1,626 @@
...
+  // set up ft
+  {
+    const acctResponse = await flow.sdk.send(await flow.sdk.pipe(await flow.sdk.build([
+      flow.sdk.getAccount(userKeys.address),
+    ]), [
+      flow.sdk.resolve([
+        flow.sdk.resolveParams,
+      ]),
+    ]), { node: flowConstants.host });
+    const seqNum = acctResponse.account.keys[0].sequenceNumber;
```
ChrisEddy, line 93: This is duplicated 8 times in the code, maybe break this out as it's own function if possible.

blockchain.js
avaer: As far as I can tell the APIs always return this format.

blockchain.js
avaer: Yes, that's a good call

ChrisEddy:
Just a couple review comments, file conflict needs resolve before merge.

PR #57, opened 2020-09-19
buttons

ChrisEddy:


planet.js
```diff
@@ -777,23 +777,34 @@ planet.connect = async (rn, {online = true} = {}) => {
...
         channelConnectionOpen = false;
+        presenceHost = '';
       } else {
-        planet.connect('lol');
-        button.innerHTML = `
-          <i class="fal fa-wifi-slash"></i>
-          <div class=label>Disconnect</div>
-        `;
-      }
+          const response = fetch('https://worlds.exokit.org/create', {
```
avaer, line 795: Probably should be a constant.

planet.js
```diff
@@ -777,23 +777,34 @@ planet.connect = async (rn, {online = true} = {}) => {
 window.addEventListener('load', () => {
   const button = document.getElementById('connectButton');
   if (button) {
-    document.getElementById('connectButton').addEventListener('click', (e) => {
+    document.getElementById('connectButton').addEventListener('click', async (e) => {
```
avaer, line 780: This could probably be moved outside of window load.

planet.js
```diff
@@ -15,7 +15,7 @@ import {makePromise} from './util.js';
 // import * as THREE from './three.module.js';
 // import { makeTextMesh } from './vr-ui.js';
 
-const presenceHost = `wss://${document.location.hostname}:4443`;
+let presenceHost = '';
```
avaer, line 18: Does this even need to be a variable up here anymore?

planet.js
ChrisEddy: fixed

planet.js
ChrisEddy: fixed

planet.js
ChrisEddy: probs fine

ChrisEddy:
this works now

planet.js
avaer: No it's not.

avaer: Is there a reason for this timing?

planet.js
```diff
@@ -10,13 +10,11 @@ import {
 } from './constants.js';
 import {XRChannelConnection} from 'https://2.metartc.com/xrrtc.js';
 import {loginManager} from './login.js';
-import {storageHost} from './constants.js';
+import { storageHost, worldsHost} from './constants.js';
```
avaer, line 13: Spacing.

edit.js
```diff
@@ -7384,8 +7393,10 @@ window.addEventListener('keydown', e => {
       break;
     }
     case 67: { // C
-      document.querySelector('.weapon[weapon="build"]').click();
-      buildMode = 'stair';
+      if (!keys.ctrl && document.pointerLockElement) {
```
avaer, line 7396: This breaks camera controls.

edit.js
```diff
@@ -7233,6 +7233,7 @@ const keys = {
   left: false,
   right: false,
   shift: false,
+  ctrl: false
```
avaer, line 7236: There is probably no need to track the ctrl key.

avaer (changes requested):
I think this breaks the controls, in addition to having binding races.

planet.js
ChrisEddy: fixed

avaer (approved):

Issue #72, opened 2020-09-23
Preact UI


avaer:
- Add Preact
- Integrate menus for tools as a 2D render on the page
- System for stringifying preact and rendering it with [exokit-render](https://github.com/exokitxr/exokit-render) and `vr-ui.js`
- System for tracking anchor callbacks for 3D renders so VR clicks work
- Copy 3D menus to 2D

Later:
- Split 2D and 3D menu systems to optimize for the respective form factor

ChrisEddy:
This is a pre bug issue. I'ma keep this but do smaller tickets to split up the work a bit.

avaer:
Well in progress.

PR #74, opened 2020-09-23
Inventory UIs.

ChrisEddy:


edit.js
```diff
@@ -46,6 +46,9 @@ import {GuardianMesh} from './land.js';
 import {storageHost} from './constants.js';
 import app from './app-object.js';
 import inventory from './inventory.js';
+import Inventory from './preact-ui/index.js';
+
+console.log(Inventory)
```
avaer, line 51: Probably no need to log this.

preact-xr-ui/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "preact-xr-ui",
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "preact build --src preact-xr-ui --dest preact-xr-ui/build/"
+  },
+  "author": "The Metaverse",
```
avaer, line 9: Lol.

Maybe set `private: true`

vr-ui.js
```diff
@@ -3,7 +3,7 @@ import * as THREE from './three.module.js';
 import {TextMesh} from './textmesh-standalone.esm.js';
 import easing from './easing.js';
 import * as icons from './icons.js';
-
+import { Inventory as InventoryHTML } from './preact-ui/components/Inventory.js';
```
avaer, line 6: Is this the built version?

avaer (commented):
This looks great on first class, we just need to make sure we're not totally breaking VR with this.

I'm guessing it also needs an `npm run build` script and Github Action.

avaer:
https://github.com/webaverse/app/issues/72

ChrisEddy:
no more Preact, no more GH actions, doing it vanilla.

.gitignore
```diff
@@ -1 +1,3 @@
 node_modules
+preact-xr-ui/build/
```
avaer, line 2: Not needed anymore?

components/helpers.js
```diff
@@ -0,0 +1,3 @@
+export const functionValueExtractor = async (fn) => {
```
avaer, line 1: What is the purpose of this?

edit.html
```diff
@@ -1,192 +1,19 @@
...
 <html>
-
 <head>
   <title>xrpackage | edit</title>
   <link href="index.css" rel=stylesheet>
   <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700&display=swap" rel="stylesheet">
   <link href="https://fonts.googleapis.com/css2?family=Muli:wght@400;600&display=swap" rel="stylesheet">
   <script src="https://kit.fontawesome.com/0735724151.js" crossorigin="anonymous"></script>
-  <!-- <script src="https://rawcdn.githack.com/ethereum/web3.js/a6ddec59e65116853435f203b25cb9c55824d084/dist/web3.min.js"></script> -->
+  <link href="./components/styles/index.css" rel=stylesheet>
```
avaer, line 9: If this is used for components, they should have their own tags for it. Otherwise this doesn't work for 3D.

index.css
```diff
@@ -10,7 +10,7 @@
 body {
   margin: 0;
   /* font-family: 'Open Sans'; */
-  font-family: 'Muli';
+  font-family: 'Muli' !important;
```
avaer, line 13: Probably shouldn't be `!important`

index.css
```diff
@@ -65,7 +65,7 @@ img:not([src]) {
   background-color: #FFF;
   align-items: center;
   user-select: none;
-  z-index: 1;
+  z-index: 10;
```
avaer, line 68: Why this magic value?

planet.js
```diff
@@ -777,24 +777,26 @@ planet.connect = async ({online = true, roomName: rn, url = null} = {}) => {
...
-    location.search = '';
-  } else { // connect case
-    const response = await fetch(`${worldsHost}/create`, {
-      method: 'POST'
-    })
-    if (response.ok) {
-      const json = await response.json();
-      // console.log(json);
-      location.search = `?w=${json.id}`;
+document.addEventListener('load',  () => {
```
avaer, line 780: Why this deep listener?

wallet.html
```diff
@@ -97,9 +97,11 @@ <h1 id=file-name></h1>
 }
 
 let file = null;
-bindUploadFileButton(document.getElementById('input-file'), f => {
-  file = f;
-});
+document.addEventListener('load', () => {
```
avaer, line 100: Why this listener?

components/Opaque.js
```diff
@@ -0,0 +1,4 @@
+const Opaque = (props) => {
```
avaer, line 1: For optimization reasons, it's not going to be a good idea to spread out little abstract components like this.

We can still do this, but I would collect them in a single file.

components/Login.js
```diff
@@ -0,0 +1,12 @@
+import { tryLogin } from '../login.js';
+
+window.addEventListener('load', async () => {
```
avaer, line 3: A component file should not be performing global initialization logic.

ChrisEddy:
dead branch

PR #80, opened 2020-09-26
New login

avaer:


ChrisEddy:
too much code to know if it works but syntax looks good

login.js
```diff
@@ -1,28 +1,67 @@
...
+      limit: 100,
+      script: contractSource.replace(/ARG0/g, '0x' + loginToken.addr),
+      wait: true,
+    }),
+  });
+  const response = await res.json();
+  const isBaked = response.encodedData.value;
+  if (!isBaked) {
+    const contractSources = await getContractSource('bakeUserAccount.json');
+    for (const contractSource of contractSources) {
```
ChrisEddy, line 27: contractSources can be null

login.js
```diff
@@ -1,28 +1,67 @@
...
+    method: 'POST',
+    body: JSON.stringify({
+      limit: 100,
+      script: contractSource.replace(/ARG0/g, '0x' + loginToken.addr),
+      wait: true,
+    }),
+  });
+  const response = await res.json();
+  const name = response.encodedData.value[0].value ? response.encodedData.value[0].value.value : 'Anonymous';
+  const avatarHash = response.encodedData.value[1].value && response.encodedData.value[1].value.value;
```
ChrisEddy, line 58: value[1].value.value pretty weird

login.js
```diff
@@ -32,17 +71,31 @@ async function pushUserObject() {
...
+async function finishLogin(newLoginToken) {
+  await storage.set('loginToken', newLoginToken);
+
+  loginToken = newLoginToken;
+
+  console.log('got user token', loginToken);
+
+  const loginForm = document.getElementById('login-form');
+  // document.body.classList.add('logged-in');
+  loginForm.classList.remove('phase-1');
```
ChrisEddy, line 92: To do: not a huge fan of the phase system. It works tho for now. Preact can handle this with some simple props now.

login.js
```diff
@@ -147,27 +167,57 @@ async function tryLogin() {
...
+  /* document.getElementById('unloadAvatar').addEventListener('click', async (e) => {
     e.preventDefault();
     e.stopPropagation();
     loginManager.setAvatar(null);
-  })
+  }); */
+
+  const userName = document.getElementById('user-name');
+  userName.addEventListener('click', e => {
+    userName.setAttribute('contenteditable', '');
```
ChrisEddy, line 180: All this can be preact soon.

ChrisEddy (approved):

login.js
avaer: That would be a serious error in the contracts deployment that we probably can't recover from anyway, so I'm fine with that crashing here.

login.js
avaer: Yeah. Unfortunately I don't know a better way to peel the data off, though we might be able to abstract these more in the future.

login.js
avaer: I agree, though that is not related to this PR.

PR #110, opened 2020-10-06
fix position undefined bug in uimanager

ChrisEddy:


avaer:
This is a merge of my code into a fix branch, so it's fine.

avaer (approved):

PR #111, opened 2020-10-06
fix position

ChrisEddy:


avaer:
I don't think meshes failing to load should be ignored. We should make sure a placeholder is loaded.

avaer (changes requested):
Should make sure a proper placeholder is loaded instead in `loadFileForWorld`.

avaer:
Is this one still valid?

ChrisEddy:
was a trash pr

PR #121, opened 2020-10-09
social and worlds update

ChrisEddy:


components/Social.js
```diff
@@ -1,8 +1,49 @@
-const Social = (props) => {
+const PeerCard = (props) => {
+  console.log(props)
   return `
-    <div>
-
+    <div class="twoD-social-peerCard">
+      <div style="width: 220px;">
+        <img class="twoD-social-peerCard-avatar" src="https://media.istockphoto.com/vectors/default-avatar-profile-icon-gray-placeholder-photo-vector-id844061224?b=1&k=6&m=844061224&s=612x612&w=0&h=adm3xmgWwC3YTVC2swWz02NDmtZ6vFaIw8a_XBGsOLk=">
```
avaer, line 6: Shouldn't do external URLs.

components/Social.js
```diff
@@ -1,8 +1,49 @@
...
+      </div>
+      <div style="width: 220px;">
+        <h2>Peer ID</h2>
+        <h1>${props.peerName}</h1>
+      </div>
+      <div style="width: 220px; text-align: center; padding-top: 20px;">
```
avaer, line 12: This should be in a style block.

PR #122, opened 2020-10-09
add allItems / Browse section

ChrisEddy:


components/App.js
```diff
@@ -158,7 +158,8 @@ export const toggleMenus = props => {
         activeTab: appState.menu.activeTab,
         inventoryItems: appState.menu.inventory.items,
         worlds: appState.menu.worlds,
-        peers: appState.menu.social.peers
+        peers: appState.menu.social.peers,
+        allItems: state.menu.browse.items
```
avaer, line 162: appState?

components/Browse.js
```diff
@@ -1,8 +1,46 @@
...
+  console.log(props)
+  let allItems = props.allItems || [];
+  return `
+    <div class="twoD-browse">
+      <div class="twoD-browse-list">
+        ${
+          allItems.map((value, index) => {
+            return BrowseCard({
+              id: value.id,
+              name: value.filename || '',
```
avaer, line 37: Can this ever be blank? If not it shouldn't have a default.

edit.html
```diff
@@ -10,6 +10,7 @@
   <link href="components/styles/menu.css" rel=stylesheet>
   <link href="components/styles/world.css" rel=stylesheet>
   <link href="components/styles/social.css" rel=stylesheet>
+  <link href="components/styles/browse.css" rel=stylesheet>
```
avaer, line 13: Can these CSS files be collapsed? It's already kind of not good that they are separated from the rendering.

ChrisEddy: maybe but not bad to keep separate

components/Browse.js
ChrisEddy: could be blank if a bug or something

avaer: But then the bug should be fixed or we should throw. What about value, can that be blank? What about id? Where does it end?

avaer: If this is a legit case we should document why this is possible. If not then there shouldn't be a default.

ChrisEddy: we should check that further up the logic not in the "dumb views", this is just to not error in any situation of using the view

edit.html
avaer: It is extra load time.

components/Browse.js
avaer: But id can't be blank?

edit.html
ChrisEddy: I mean... 200ms~ not that big of a deal imo right now, pre sure the JS takes another 7 secs to complete

xcvbxcvbxcvbxcvb

avaer: It's different on a Quest tethered to a phone.

ChrisEddy: https://www.npmjs.com/package/uglify-js recommend minifying everything into 1 file then

ChrisEddy: 2k line CSS files no fun

ChrisEddy: ill do later as a complete optimization, compression, minify.

avaer: I would just concat them. But yes.

PR #134, opened 2020-10-14
back button for trade

ChrisEddy:


components/Trade.js
```diff
@@ -41,7 +42,7 @@ const Trade = props => {
                   id: value.id,
                   name: value.filename,
                   preview: value.preview,
-                  selected: selectedItem === value.id,
+                  selected: selectedItem == value.id,
```
avaer, line 45: Why single equals? If it's the wrong type it should be parsed before being set on the state.

avaer (changes requested):
Small `selectedItem` type issue?

components/App.js
```diff
@@ -218,7 +218,7 @@ export const onclickBindings = {
   'twoD-inventoryCardTradeBtn': e => {
     const { menu } = getState();
     menu.trade.visible = true;
-    menu.trade.selectedItem = e.name
+    menu.trade.selectedItem = parseInt(e.name);
```
avaer, line 221: Should have `, 10` since the default of parseInt is crazy.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

avaer (commented):
Another small thing but then it's mergeable.

PR #135, opened 2020-10-14
restyle 2d menu, dark mode, no curves, etc

ChrisEddy:


avaer:
This is much better. It's probably good for merge since it's not a regression, but some design comments:

- Using a different font (3D uses Bangers); not sure what's appropriate but we should be consistent. There are three fonts in the UI now, if you include the avatar name tag.
- still a few lines/borders left around tabs and buttons that aren't present in 3d
- buttons not really aligned (will be replaced by panel)
Screen Shot 2020-10-14 at 9 56 07 AM

avaer (approved):

ChrisEddy:
buttons going to be taken out of cards when the preview panel goes in next, 👍🏻

PR #137, opened 2020-10-14
inventory preview panel

ChrisEddy:


avaer (approved):

avaer:
It's a good direction. It's still kind of hard to try use the panel on a smaller screen (macbook air), since it doesn't hold the full content on the right, and there is no indication it should scroll.

The content should probably collapse in a smaller bounding box.

Screen Shot 2020-10-14 at 6 22 42 PM

PR #181, opened 2020-11-25
Add boilerplate react UI setup with nav and view stubs

shawticus:
This adds boilerplate for a UI using es-react. No compile, no transpile.

Use "npx servor" as command to host from command line, it will update on code change (you will still need to refresh the page)


avaer:
Look like a good bootstrap. Thanks!

avaer (approved):

PR #207, opened 2020-12-05
Web Monetization

chrislatorres:
> the end goal is the user sets up their payment pointer on their "cryptoprofile", and from then on any of their owned NFTs that load trigger a payment stream in-world for supported browsers/WebMonetization enabled user agents

  pairs with https://github.com/webaverse/ethereum-bot/pull/1 and https://github.com/webaverse/ethereum-bot/pull/2 

webmonetizednft

webatest12


world.js
```diff
@@ -781,7 +782,31 @@ const _connectRoom = async (roomName, worldURL) => {
...
-world.addObject = (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
+world.addObject = async (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
+  (async () => {
+    const token = await contracts.sidechain.NFT.methods.tokenByIdFull(parseInt(contentId)).call();
+    const monetizationPointer = await contracts.sidechain.Account.methods.getMetadata(token.owner, 'monetizationPointer').call();
+    if (monetizationPointer && document.monetization && !document.querySelector("meta[name=monetization]")) {
+      const monetizationTag = document.createElement('meta');
+      monetizationTag.name = 'monetization';
+      monetizationTag.content = monetizationPointer;
+      document.head.appendChild(monetizationTag);
```
avaer, line 793: This tag should be removed when the object is removed.

world.js
```diff
@@ -781,7 +782,31 @@ const _connectRoom = async (roomName, worldURL) => {
...
+
+      document.monetization.addEventListener('monetizationprogress', ev => {
```
avaer, line 795: This listener should be cancelled when the object is removed.

world.js
```diff
@@ -781,7 +782,31 @@ const _connectRoom = async (roomName, worldURL) => {
 
 const objects = [];
 world.getObjects = () => objects;
-world.addObject = (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
+world.addObject = async (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
+  (async () => {
+    const token = await contracts.sidechain.NFT.methods.tokenByIdFull(parseInt(contentId)).call();
```
avaer, line 787: It might make more sense to put this as part of the backend metadata fetch. It would probably be much faster.

chrislatorres: Actually, could already pull owner from https://tokens.webaverse.com/16 for example -- would it be a good idea to add the `monetizationPointer` as well?

chrislatorres: b193246, 2c57e49, and https://github.com/exokitxr/exokit-backend/pull/34

world.js
chrislatorres: 0559424

world.js
chrislatorres: eb6d236

world.js
```diff
@@ -781,8 +824,8 @@ const _connectRoom = async (roomName, worldURL) => {
 
 const objects = [];
 world.getObjects = () => objects;
-world.addObject = (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
-  state.transact(() => {
+world.addObject = async (contentId, parentId = null, position = new THREE.Vector3(), quaternion = new THREE.Quaternion(), options = {}) => {
```
avaer, line 827: Heads up, this should probably not be `async` in the end.

[`contentId` need not be a token at all](https://github.com/webaverse/app/blob/a67a805227758078ddd329b488316d39007536a7/world.js#L858); it could also be a url in which case we can't look up monetization on our end.

It's also probably not a critical error for the monetization flow to fail and throw (we should still load the object).

chrislatorres: - e4ebb02 removes the added async and moves it to existing token backend call on trackedobjectadd event
- 9cd290c checks for pointer before pushing
- 0e6414a adds checking for [valid monetization pointer synax](https://paymentpointers.org/#syntax).

runtime.js
```diff
@@ -540,6 +540,13 @@ const _loadScript = async (file, {files = null, parentUrl = null} = {}) => {
     script = script.replace(r, function() {
       return arguments[1] + replacements[index++] + arguments[3];
     });
+    if (instanceId) {
+      script = script.replace("document.monetization", `window.document.monetization${instanceId}`);
+      script = script.replace("document.monetization.addEventListener", `window.document.monetization${instanceId}.addEventListener`);
```
avaer, line 545: Is this line needed? It looks like a superset of the one above.

Also, we shouldn't change  `document` -> `window.document`.

runtime.js
```diff
@@ -540,6 +540,13 @@ const _loadScript = async (file, {files = null, parentUrl = null} = {}) => {
...
+      script = `
+        window.document.monetization${instanceId} = window.document.createElement('div');
```
avaer, line 547: Probably `new EventTarget()` is more logical than a DOM element.

runtime.js
```diff
@@ -456,7 +456,7 @@ const _makeAppUrl = appId => {
   });
   return URL.createObjectURL(b);
 };
-const _loadScript = async (file, {files = null, parentUrl = null} = {}) => {
+const _loadScript = async (file, {files = null, parentUrl = null} = {}, instanceId) => {
```
avaer, line 459: `instanceId` is feels like it should be part of the `options`.

world.js
```diff
@@ -32,6 +32,63 @@ const localMatrix = new THREE.Matrix4();
...
+let pendingMonetizationStart = [];
+setInterval(() => {
+  if (pendingMonetizationStart.length > 0) {
+    pendingMonetizationStart.map((id, i) => {
+      const instanceId = pendingMonetizationStart[i].instanceId;
+      const object = pendingMonetizationStart[i].object;
+      const monetization = window.document[`monetization${instanceId}`];
+      const ethereumAddress = loginManager.getAddress();
+
+      const listener = new THREE.AudioListener();
```
avaer, line 44: The audio part seems like it should be part of the NFT. An NFT might want totally different behavior/to play as different sound.

world.js
```diff
@@ -32,6 +32,63 @@ const localMatrix = new THREE.Matrix4();
 const localMatrix2 = new THREE.Matrix4();
 const localRaycaster = new THREE.Raycaster();
 
+let pendingMonetizationStart = [];
+setInterval(() => {
```
avaer, line 36: Pending monetizations can just be queued synchronously, right? I don't quite understand this delay, unless it's intended for audio effects.

world.js
chrislatorres: That makes sense. Should there be any Webaverse UI/UX when and item is unlocked, or should we leave that to the NFT creators?

avaer: Monetization status should be viewable in the context sidebar, but there could also be in-world popup notifications for WM specifically.

I think it should be discoverable but not intrusive into author intent.

runtime.js
chrislatorres: d383152

runtime.js
chrislatorres: d383152

runtime.js
chrislatorres: 40f66f1

world.js
chrislatorres: There is a ~5ms delay between the script loading in that creates the custom `eventTarget` and the code running.. which is why this has to keep checking to see if the custom `eventTarget` exists yet.

39ca41e 

11msasync
lolasync

world.js
chrislatorres: > Monetization status should be viewable in the context sidebar

Here's a basic version, a money sign icon on the context sidebar for items when web monetization is enabled:  df3b8f4

monetizedicon

edit.html
```diff
@@ -18,6 +18,7 @@
       <img class="img primary open" src="assets/logo-circle.svg" id=location-icon>
       <img class="img secondary" src="https://preview.exokit.org/[https://raw.githubusercontent.com/avaer/vrm-samples/master/vroid/male.vrm]/preview.jpg" id=profile-icon>
       <img class="img tertiary" src="https://preview.exokit.org/[https://raw.githubusercontent.com/avaer/vrm-samples/master/vroid/female.vrm]/preview.jpg" id=item-icon>
+      <img class="img tertiary" src="https://user-images.githubusercontent.com/29695350/101466101-5d219100-3906-11eb-9f7b-ccbf00492f3c.jpg" id=item-monetized-icon>
```
avaer, line 21: Probably best to have this image in the repo.

weapons-manager.js
```diff
@@ -2205,7 +2206,9 @@ const _updateMenu = () => {
     objectMenuEl.classList.toggle('open', true);
     itemLabel.classList.toggle('open', true);
     itemIcon.classList.toggle('open', true);
-
+    if (document.monetization.state == 'started') {
+      itemMonetizedIcon.classList.toggle('open', true);
```
avaer, line 2210: We'll probably need to make this per-item once items are hooked in.

world.js
avaer: It sounds like this code should go wherever the conditions for its triggering change, not on an interval. If that is after the app script is loaded in the runtime then maybe this update should happen there?

https://github.com/webaverse/app/blob/111d3b7d762425d5c7b605a0b3bb7dfb9701a261/runtime.js#L473

world.js
```diff
@@ -833,11 +880,11 @@ world.addEventListener('trackedobjectadd', async e => {
...
   const {instanceId, parentId, contentId, position, quaternion, options: optionsString} = trackedObjectJson;
   const options = JSON.parse(optionsString);
+  const res = await fetch(`https://tokens.webaverse.com/${contentId}`);
+  const token = await res.json();
 
   const file = await (async () => {
-    if (typeof contentId === 'number') {
-      const res = await fetch(`https://tokens.webaverse.com/${contentId}`);
-      const token = await res.json();
+    if (typeof contentId === 'number' || /^\d+$/.test(contentId)) {
```
avaer, line 887: If `contentId` comes in as a string that is probably a bug elsewhere.

avaer (commented):
Done second pass 👍 .

world.js
chrislatorres: worked perfectly 
0f6f10d

edit.html
chrislatorres: 8f451f2

weapons-manager.js
chrislatorres: Yeah. Right now it is in-line with the current implementation, which is -- if monetization is enabled, then monetization is rotated between the items.. to the owner's monetization pointer.

world.js
chrislatorres: f479cbe

world.js
```diff
@@ -32,11 +32,26 @@ const localMatrix = new THREE.Matrix4();
 const localMatrix2 = new THREE.Matrix4();
 const localRaycaster = new THREE.Raycaster();
 
+let pendingAnalyticsData = {};
+setInterval(() => {
+  Object.keys(pendingAnalyticsData).map(item => {
```
avaer, line 37: `forEach` is preferred when the result of the map is not needed.

world.js
```diff
@@ -958,10 +992,10 @@ world.addEventListener('trackedobjectadd', async e => {
     const minimapObject = minimap.addObject(mesh);
     mesh.minimapObject = minimapObject;
 
-    if (token.owner.address && token.owner.monetizationPointer && token.owner.monetizationPointer[0] === "$") {
+    if (contentId && instanceId && token.owner.address && token.owner.monetizationPointer && token.owner.monetizationPointer[0] === "$") {
```
avaer, line 993: I don't think it ever makes sense for there to not be these ids; that would be a bug.

world.js
chrislatorres: 774e36a

world.js
chrislatorres: 502e348

runtime.js
```diff
@@ -540,6 +550,12 @@ const _loadScript = async (file, {files = null, parentUrl = null} = {}) => {
     script = script.replace(r, function() {
       return arguments[1] + replacements[index++] + arguments[3];
     });
+    if (instanceId) {
+      script = script.replace(/document.monetization/g, `document.monetization${instanceId}`);
```
avaer, line 554: That will need a `.`, because `.` has special meaning in a regex.

world.js
```diff
@@ -833,11 +891,11 @@ world.addEventListener('trackedobjectadd', async e => {
   const trackedObjectJson = trackedObject.toJSON();
   const {instanceId, parentId, contentId, position, quaternion, options: optionsString} = trackedObjectJson;
   const options = JSON.parse(optionsString);
+  const res = await fetch(`https://tokens.webaverse.com/${contentId}`);
```
avaer, line 894: I think `contentId` can be a URL, in which case this will fail.

world.js
```diff
@@ -797,6 +852,9 @@ world.addObject = (contentId, parentId = null, position = new THREE.Vector3(), q
 };
 world.removeObject = removeInstanceId => {
   state.transact(() => {
+    const index = pointers.findIndex(x => x.instanceId === removeInstanceId);
+    if (index !== undefined) pointers.splice(index, 1);
```
avaer, line 856: This won't work because `findIndex` returns a number (check against -1 instead).

world.js
```diff
@@ -32,6 +32,61 @@ const localMatrix = new THREE.Matrix4();
 const localMatrix2 = new THREE.Matrix4();
 const localRaycaster = new THREE.Raycaster();
 
+let pendingAnalyticsData = {};
+setInterval(() => {
```
avaer, line 36: Can we abstract this into an `web-monetization.js`? The code looks good but it seems separate from `world.js`.

avaer (commented):
Couple small things, then it's good to go 👍. Awesome work.

avaer:
Let me know if there are any questions about the conflicts.

world.js
chrislatorres: 14e3d33

world.js
chrislatorres: 50e067f

chrislatorres:
Fixed conflicts and tested again. All good 👍

web-monetization.js
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!document.querySelector("meta[name=monetization]")) {
+    const monetizationTag = document.createElement('meta');
+    monetizationTag.name = 'monetization';
+    monetizationTag.content = pointers[currentIndex].monetizationPointer;
+    document.head.appendChild(monetizationTag);
+    document.monetization.addEventListener('monetizationprogress', function monetizationProgress (e) {
+      const current = pointers[currentIndex];
+      const currentMonetizationPointer = encodeURIComponent(current.monetizationPointer);
+      const apiUrl = `https://api.metaverse.website/monetization/${current.contentId}/${current.ownerAddress}/${currentMonetizationPointer}`;
```
avaer, line 29: We can change this to `analytics.webaverse.com`.

avaer: `d-r2l88xz50c.execute-api.us-east-2.amazonaws.com`

avaer (approved):
8/8

PR #221, opened 2020-12-08
Fix avatar img

### LINKED ISSUE #220: Incorrect in-world profile picture
diffprofilepic

chrislatorres:
Fixes #220 
avatarimg

### DIFF
--- a/rig.js
+++ b/rig.js
@@ -146,9 +146,9 @@ class RigManager {
     const avatarMesh = new THREE.Mesh(geometry, material);
     avatarMesh.position.x = -0.5;
     avatarMesh.position.y = -0.02;
-    this.localRig.textMesh.add(avatarMesh);
 
-    this.localRig.textMesh.sync();
+    this.localRig.textMesh.remove(this.localRig.textMesh.children[0]);
+    this.localRig.textMesh.add(avatarMesh);
   }
 
   async setLocalAvatarUrl(url, filename) {
--- a/rig.js
+++ b/rig.js
@@ -128,6 +128,9 @@ class RigManager {
   }
 
   setLocalAvatarImage(avatarImage) {
+    if (this.localRig.textMesh.avatarMesh) {
+      this.localRig.textMesh.remove(this.localRig.textMesh.avatarMesh);
+    }
     const geometry = new THREE.CircleBufferGeometry(0.1, 32);
     const img = new Image();
     img.src = avatarImage;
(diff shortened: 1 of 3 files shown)


rig.js
```diff
@@ -127,6 +127,30 @@ class RigManager {
...
+    };
+    const texture = new THREE.Texture(img);
+    const material = new THREE.MeshBasicMaterial({
+      map: texture,
+      side: THREE.DoubleSide,
+    });
+    const avatarMesh = new THREE.Mesh(geometry, material);
+    avatarMesh.position.x = -0.5;
+    avatarMesh.position.y = -0.02;
+    this.localRig.textMesh.add(avatarMesh);
```
avaer, line 149: I suspect this will "paper over" the existing avatar mesh without removing it, and additional sets will further add more icons.

https://github.com/webaverse/app/blob/fbb507c4d0e2cc10cf3ef2251dabacf084c34b54/rig.js#L39

It probably makes sense to remove that icon rather than add ones on top.

rig.js
```diff
@@ -127,6 +127,30 @@ class RigManager {
...
+
+    this.localRig.textMesh.sync();
```
avaer, line 151: If we aren't changing the text mesh, there is no need to sync.

avaer (commented):
This looks functionally correct, but it might not be removing old avatar images.

rig.js
chrislatorres: 1a301bc

rig.js
chrislatorres: 1a301bc

rig.js
```diff
@@ -127,6 +127,30 @@ class RigManager {
...
+    const texture = new THREE.Texture(img);
+    const material = new THREE.MeshBasicMaterial({
+      map: texture,
+      side: THREE.DoubleSide,
+    });
+    const avatarMesh = new THREE.Mesh(geometry, material);
+    avatarMesh.position.x = -0.5;
+    avatarMesh.position.y = -0.02;
+
+    this.localRig.textMesh.remove(this.localRig.textMesh.children[0]);
```
avaer, line 150: I would be careful about depending on child order.

The way I usually do it is like `this.localRig.textMesh.avatarMesh = avatarMesh` and then it's by name.

avaer: Otherwise adding a child in a different order will explode this, which would be very surprising.

avaer (commented):
One small futureproofing and then it looks good.

rig.js
chrislatorres: 6d0860a and 537c46e

avaer (approved):
LGTM

PR #329, opened 2020-12-18
Add per-item monetization tracking

chrislatorres:
This PR adds monetization tracking for each item. Also now supports monetization tracking for all file types, not just js files/web bundles.

Note that the new analytics endpoint is not working right now.


runtime.js
```diff
@@ -40,6 +40,28 @@ basisLoader.detectSupport(renderer);
...
+  }
+
+  let monetization = document[`monetization${instanceId}`];
+  if (!monetization) {
+    document[`monetization${instanceId}`] = document.createElement('eventTarget');
+    monetization = document[`monetization${instanceId}`];
+  }
+
+  const ethereumAddress = loginManager.getAddress();
+  if (ethereumAddress && ethereumAddress == ownerAddress) {
```
avaer, line 55: `===`?

weapons-manager.js
```diff
@@ -1278,7 +1278,8 @@ const _updateMenu = () => {
     lastSelectedBuild = -1;
     lastCameraFocus = -1;
   } else if (highlightedObject) {
-    if (document.monetization && document.monetization.state == 'started') {
+    const monetization = document[`monetization${highlightedObject.instanceId}`];
+    if (monetization && monetization.state == 'started') {
```
avaer, line 1282: `===` ?

avaer (commented):
Just some small things.

web-monetization.js
```diff
@@ -26,7 +26,7 @@ setInterval(() => {
     document.monetization.addEventListener('monetizationprogress', function monetizationProgress (e) {
       const current = pointers[currentIndex];
       const currentMonetizationPointer = encodeURIComponent(current.monetizationPointer);
-      const apiUrl = `https://analytics.webaverse.com/monetization/${current.contentId}/${current.ownerAddress}/${currentMonetizationPointer}`;
+      const apiUrl = `https://analytics.webaverse.com/dev/monetization/${current.contentId}/${current.ownerAddress}/${currentMonetizationPointer}`;
```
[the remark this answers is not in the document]
chrislatorres: fdbec08

runtime.js
chrislatorres: 00f086a

weapons-manager.js
chrislatorres: ca584bb

runtime.js
```diff
@@ -40,6 +40,28 @@ basisLoader.detectSupport(renderer);
...
 basisLoader.detectSupport(renderer);
 
+const startMonetization = (instanceId, monetizationPointer, ownerAddress) => {
+  if (!monetizationPointer) {
+    return;
+  }
+
+  let monetization = document[`monetization${instanceId}`];
+  if (!monetization) {
+    document[`monetization${instanceId}`] = document.createElement('eventTarget');
```
avaer, line 50: I don't think this is very legal; `eventTarget` is not really a DOM node name.

If we want an event target the way to do it is `new EventTarget()`.

avaer: Though this works by chance because a DOM node called `eventTarget` is actually an `EventTarget` by inheritance, due to coincidence.

avaer (commented):
One small cleanup.

runtime.js
chrislatorres: 8a77ea2

avaer (approved):

PR #411, opened 2020-12-29
Update styling

chrislatorres:
## Before
beforeweba

## After
uiledgibility

## White on black legibility
Works on light or dark surfaces
loadoutp1
loadoutp12

### DIFF
--- a/index.css
+++ b/index.css
@@ -101,13 +101,13 @@ img:not([src]) {
   margin-left: -5px;
   padding: 2px 5px;
   color: #FFF;
-  background-color: var(--color);
+  background-color: #c4005d;
   z-index: -1;
 }
 .header .icon > .img.primary,
 .header .icon > .name.primary
 {
-  --color: #7986cb;
+  --color: #c4005d;
 }
 .header .icon > .img.secondary,
 .header .icon > .name.secondary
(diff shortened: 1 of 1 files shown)


avaer (changes requested):
During testing I noticed that this causes a large performance hit when hovering over items.

This is likely because of some opacity stacking, drop shadows, and other such effects which we deliberately do not use.

The design looks cool, but it probably can't go in until we investigate that.

avaer:
Testing on MacOS btw, these performance issues are sometimes OS related.

chrislatorres:
Looks like ` filter: drop-shadow` is notoriously bad for perf, going to find alternative

chrislatorres:
Updated to use `text-shadow` instead, and it works well against completely white backgrounds
legibility

avaer:
Keys are illegible.

https://cdn.discordapp.com/attachments/737801646364557394/793760125118578718/Screen_Shot_2020-12-30_at_3.39.02_AM.png

avaer:
Gradient to zero would work better.

https://cdn.discordapp.com/attachments/737801646364557394/793760326847954954/Screen_Shot_2020-12-30_at_3.39.57_AM.png

avaer:
Tab menu regressed.

https://cdn.discordapp.com/attachments/737801646364557394/793760483307290635/Screen_Shot_2020-12-30_at_3.40.35_AM.png

avaer:
Not this design's issue, but the username is illegible in this world. I needs a background.

https://cdn.discordapp.com/attachments/737801646364557394/793760599510482954/Screen_Shot_2020-12-30_at_3.41.04_AM.png

PR #483, opened 2021-01-02
use png instead of jpg for avatar pic transparency

chrislatorres:
## Before
ava2

## After
ava1


avaer:
I think this has broader effects, on the entire inventory. PNG is much slower to load than JPG, so I don't know if this is a good tradeoff.

Issue #611, opened 2021-01-18
App: interpolate avatars motion


avaer:
Lerp + slerp the rig, probably in `avatars.js`.

PR #619, opened 2021-01-18
Avatars input interpolation

avaer:
Fixes https://github.com/webaverse/app/issues/611.


avaer:
Core functionality should be done. Needs testing.

PR #663, opened 2021-01-21
Remove old react-ui code

avaer:
Was interfering with grep.


avaer:
Old code is in this branch: https://github.com/webaverse/app/tree/react-ui

Issue #637, opened 2021-01-19
Support UV Scrolling Animation


karunakaruna:
UV scrolling is a technique commonly used in games to convey motion with a material. Used for things like flowing water etc. 

Mozilla Hubs has a UV scroll component in it's blender exporter:
https://www.youtube.com/watch?v=0pCGswUlrfk
https://github.com/MozillaReality/hubs-blender-exporter

PR #696, opened 2021-01-23
Add initial uv scroll flow

avaer:
Fixes https://github.com/webaverse/app/issues/637.

Still needs some shader injection: https://github.com/mozilla/hubs/blob/6c49ce303ad7e030c80b301dbc3ce100cac7d9c1/src/components/uv-scroll.js#L36


avaer:
This works, bug behaves wrongly/glitchy compared to Hubs.

PR #810, opened 2021-02-09
lossless compression of /land-textures 

ChristopherTrimboli:
36.3mb > 15.8mb, used: https://tinypng.com/


avaer:
According to https://tinypng.com/ it is not lossless, so I am against this. These assets are not actually meant to be used directly, they are the high quality source. When we really use them we will run it through KTX/basis compression which will be many times better than this optimization.

I think these files should remain identical to how the artist provided them since that isn't harming anything.

ChristopherTrimboli:
I think we should keep the originals tucked away somewhere but for production web build, compression before they enter transit is ideal. 20mb shaved off these and can not see any difference in app. Currently we are sitting at 114 MB of total resources in client app, id like to see this get under 25mb.

avaer:
Ok that's fair. They will be in `assets` anyway.

Thanks!

avaer (approved):

PR #831, opened 2021-02-12
Mithril Menu UI

ChristopherTrimboli:


mithril-ui/menu/menu.css
```diff
@@ -0,0 +1,15 @@
+.Menu-container {
+  display: flex;
+  position: absolute;
+  z-index: 999;
```
avaer, line 4: I would be careful with random numbers like this. Can we make it only as high as we need it to be?

mithril-ui/App.js
```diff
@@ -0,0 +1,18 @@
+import {Menu} from './menu/Menu.js'
+import {appState, menuState} from './store/state.js';
+
+export const App = (initialVnode) => {
+  return {
+    oninit: (vnode) => {
+      console.log("Hello Mithrel", appState)
```
avaer, line 7: No need to log this and spam the console for everyone. Can you comment it out?

On your local dev env you can uncomment but simply not commit that.

avaer (commented):
Small things, the rest looks good.

avaer:
Let's merge these PRs gradually. As long as the code is kept working we can ship every day or two. It will stay behind the `M` key for now so it won't be intrusive.

Let me know when there is something ready to merge.

PR #837, opened 2021-02-12
Webpack.

ChristopherTrimboli:


contracts/addresses.js
```diff
@@ -0,0 +1,42 @@
+export default {
```
avaer, line 1: This should not be committed.

contracts/abi.js
```diff
@@ -0,0 +1,3586 @@
+export default {
```
avaer, line 1: No need to commit this.

app.js
```diff
@@ -392,3 +392,36 @@ export default class App {
     renderer.setAnimationLoop(animate);
   }
 }
+
+window.addEventListener("load", () => {
```
avaer, line 396: We cannot do this.

App.js is dynamically imported in the map and I think in that case we do *not* want to `bootstrapFromUrl`. But if we do it this way the bootstrap will happen and mess up the intended usage of `app.js`.

VOXLoader.js
```diff
@@ -29,7 +29,7 @@ import * as THREE from "./three.module.js";
 /**
  * @namespace
  */
-var vox = {};
+export var vox = {};
```
avaer, line 32: Why is the VOXLoader.js change necessary?

blockchain.js
```diff
@@ -6,8 +6,8 @@ import ethereumJsTx from './ethereumjs-tx.js';
 import {makePromise} from './util.js';
 import {storageHost, web3MainnetSidechainEndpoint, web3RinkebySidechainEndpoint} from './constants.js';
 const {Transaction, Common} = ethereumJsTx;
-import addresses from 'https://contracts.webaverse.com/config/addresses.js';
-import abis from 'https://contracts.webaverse.com/config/abi.js';
+import addresses from './contracts/addresses.js';
```
avaer, line 9: This should not be locally imported.

avaer (commented):
A few things to keep in mind when we pick this back up.

PR #855, opened 2021-02-13
action slot onclicks and some style tweaks

ChristopherTrimboli:


mithril-ui/menu/Inventory.js
```diff
@@ -47,14 +47,27 @@ const InspectPanel = (initialVnode) => {
 };
 
 const ActionSlots = (initialVnode) => {
+  const onSlotClick = (slot) => {
+    actionSlotsActions.setActionSlot({
+      id: slot.id,
+      image: !slot.image ? menuState.selectedItem.preview_url : null
```
avaer, line 53: I don't know if it makes sense as a toggle in the final version but this works for testing.

ChristopherTrimboli: agreed, mostly for testing

PR #877, opened 2021-02-15
initial wbn screenshot preview

### LINKED ISSUE #567: Preview: wbn
Support WBN object previews. This is easier now that the street is previewable and is already made of NFTs.

chrislatorres:
Fix #567 -- this gives a preview better than the current orange placeholder, it takes the first glb inside of the wbn and uses that for the screenshot

### DIFF
--- a/screenshot.html
+++ b/screenshot.html
@@ -248,17 +248,14 @@
       toggleElements(false);
       const screenshotResult = document.getElementById('screenshot-result');
 
-      const {url, hash, ext, type, dst} = parseQuery(decodeURIComponent(window.location.search));
+      let {url, hash, ext, type, dst} = parseQuery(decodeURIComponent(window.location.search));
 
       try {
-        const _loadGltf = async (passedUrl) => {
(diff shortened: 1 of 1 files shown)


screenshot.html
```diff
@@ -250,11 +251,14 @@ <h3 id=screenshot-header></h3>
       const {url, hash, ext, type, dst} = parseQuery(decodeURIComponent(window.location.search));
 
       try {
-        const _loadGltf = async () => {
+        const _loadGltf = async (passedUrl) => {
+          if (!passedUrl) {
```
avaer, line 255: Why would this happen? If some file is missing this doesn't seem like the place to make one up.

avaer: I would move this defaulting outside of this function or drop it altogether.

avaer (changes requested):
Just the one change, then I think this is good.

screenshot.html
chrislatorres: droped it

PR #879, opened 2021-02-15
Switch to mainnet

chrislatorres:


chrislatorres:
Tested locally, this looks good to go

blockchain.js
```diff
@@ -37,7 +37,7 @@ function _setMainChain(isMainChain) {
   }
 }
 // _setMainChain(!/^test\./.test(location.hostname));
-_setMainChain(/^main\./.test(location.hostname));
+_setMainChain(true);
```
avaer, line 40: Import the constant?

avaer (commented):
One thing and this is ready to go, but we should deploy the contracts first.

blockchain.js
chrislatorres: added https://github.com/webaverse/app/pull/879/commits/0981f3fe49e92719fd2576cfa7eedc5e560c11b3

Issue #509, opened 2021-01-08
Renderer: postprocessing support


avaer:
Bokeh as an effect looks great, and we definitely want to support the pipeline inside of parcels.

So we'll need to experiment and figure out how to make postprocessing pipelines work.

https://github.com/vanruesc/postprocessing

avaer:
https://github.com/mrdoob/three.js/pull/18846

PR #906, opened 2021-02-18
Pass

avaer:
Fixes https://github.com/webaverse/app/issues/509.


avaer:
Replaced with https://github.com/webaverse/app/pull/993, which supports shader passes.

PR #1051, opened 2021-04-03
Image dimensions by arg

avaer:


screenshot.html
```diff
@@ -248,8 +248,14 @@ <h3 id=screenshot-header></h3>
       toggleElements(false);
       const screenshotResult = document.getElementById('screenshot-result');
 
-      let {url, hash, ext, type, dst} = parseQuery(decodeURIComponent(window.location.search));
-
+      let {url, hash, ext, type, dst, width = 512, height = 512} = parseQuery(decodeURIComponent(window.location.search));
+      // make sure the image is square
```
avaer, line 252: That doesn't seem an accurate comment anymore.

screenshot.html
```diff
@@ -248,8 +248,14 @@ <h3 id=screenshot-header></h3>
...
+      let aspectRatio = width / height;
```
avaer, line 253: We want the exact size requested to come back. We shouldn't change the result dimensions.

The renders just need to fit in the requested dimension.

avaer:
The backend for this is deployed, just this PR does not support returning the requested size yet.

avaer:
Closing in favor of https://github.com/webaverse/app/pull/1270

Issue #1082, opened 2021-04-06
Animation : left / right hand animations flipped 


plankatron:
probably an issue with bone mapping , if you have an animation on the left hand of the avatar it will show up on the right hand. need programmer for debug assistance.

avaer:
Yeah, in a few cases we flip it.

Do you have an example that renders wrong? Just so we can repro + test.

plankatron:
have not tested everything but i noticed it primarily with the sword attack anims

avaer:
Are the fingers curled the wrong way or something?

Btw Github supports gif recordings + videos of screengrabs.

plankatron:
video is too big i will link to it from somewhere else .. 
http://7rnx.net/fleeky-pub/webaverse/teasdf0001-0617.mp4

avaer:
This looks like it was fixed.

PR #1101, opened 2021-04-10
Hand swap!

jerknose:
Fix for 1082. Not sure what else may need to be swapped. We'll cross those fingers when we come to them.

https://github.com/webaverse/app/issues/1082


avaer:
This looks good, but I'm concerned it's a hydra that fixed one bug and regressed something else.

Do all of the hand animations look correct once this is merged into `master`?

avaer (approved):

PR #1102, opened 2021-04-10
Add dotenv and default

shawticus:
This adds the necessary upgrades for automated deployment to beanstalk


avaer:
This isn't used yet right?

PR #1008, opened 2021-03-26
Animations from GLB

avaer:

PR #1133, opened 2021-04-18
Animations take 2

avaer:


avaer:
Replaces https://github.com/webaverse/app/pull/1008.

avaer:
Still need to split the arms and legs inputs so that we can make the top and bottom of an avatar point separately.

app-object.js
```diff
@@ -5,15 +5,15 @@ import {addDefaultLights} from './util.js';
 let canvas = document.getElementById('canvas');
 let context = canvas && canvas.getContext('webgl2', {
   antialias: true,
-  alpha: true,
+  alpha: false,
```
avaer, line 8: What triggered this change?

Note that we alpha in `app` into many places, like `https://webaverse.com` and the preview system. Se need alpha on the main canvas for these effects.

rig.js
```diff
@@ -680,11 +680,11 @@ class RigManager {
 
     this.peerRigs = new Map();
     
-    this.lastTimestamp = Date.now();
+    this.lastTimestamp = performance.now()
```
avaer, line 683: This is probably a good change, but are you sure nothing downstream is expecting integers?

avatars/avatars.js
```diff
@@ -53,1266 +31,92 @@ window.addEventListener('keydown', e => {
...
-    const _ensurePrerotation = k => {
-      const boneName = modelBones[k].name;
-      if (!preRotations[boneName]) {
-        preRotations[boneName] = new THREE.Quaternion();
-      }
-      return preRotations[boneName];
-    };
-    if (flipY) {
-      _ensurePrerotation('Hips').premultiply(new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(1, 0, 0), -Math.PI / 2));
+  const progress = (err)=>{
```
avaer, line 53: Can this be `Promise.all` rather than a custom callback counter?

avatars/avatars.js
```diff
@@ -53,1266 +31,92 @@ window.addEventListener('keydown', e => {
...
+      console.log(err)
     }
-    if (flipZ) {
-      _ensurePrerotation('Hips').premultiply(new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(0, 1, 0), Math.PI));
-    }
-    if (flipLeg) {
-      ['LeftLeg', 'RightLeg'].forEach(k => {
-        _ensurePrerotation(k).premultiply(new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(1, 0, 0), Math.PI / 2));
-      });
+    if (--count===0){
```
avaer, line 57: I don't think this is following `.eslintrc`: https://github.com/webaverse/app/blob/master/.eslintrc.json

app-object.js
abdullaBest: the canvas is always cleared and filled with one color, in the code I have not seen anywhere a custom cleaning of the canvas, you do not need the transparency of the canvas, it only wastes resources, are you sure you need it?

avatars/avatars.js
```diff
@@ -1366,878 +170,344 @@ class Avatar {
...
 
-    this._applyAnimation(now);
-    this._applyPose();
+    this._applyAnimation(timeDiff)
+    //this._applyPose()
 
-    if (this.getTopEnabled()) {
-      this._animateTop();
-    }
+    //if (this.getTopEnabled()) {
```
avaer, line 482: If we are going to merge this, IK needs to be enabled.

rig.js
abdullaBest: rework in the process, but this change only adds accuracy to the time, I'm sure there shouldn't be any problems below

avatars/avatars.js
```diff
@@ -1366,878 +170,344 @@ class Avatar {
...
-          }
-        }
-      }
+  setAnimation(name,weight){
+    if (this.active_animations>=this.animations.length){
+      return
     }
-    if (this.headTargetEnabled) {
-      this.outputs.neck.quaternion.premultiply(this.headTarget);
+    if (this.getAnimation(name)!==-1){
```
avaer, line 389: Why is there a two step finding of the animation?

avatars/avatars.js
abdullaBest: I try to make the code cleaner and more understandable, a lot of promises and async await make the code harder to understand, well, in general, we need a data loading system in the future

avatars/avatars.js
abdullaBest: I transferred this code, it was before me) you will need to go through the entire code and bring it to beauty

avatars/avatars.js
abdullaBest: additional optimization plans

avaer:
This branch stopped loading as of commit https://github.com/webaverse/app/pull/1133/commits/98637c03cbc9d17ca742719a373a425beda6f977.

image

Therefore I reverted that commit with https://github.com/webaverse/app/pull/1133/commits/30afe6973246d4652c2a81f1bc7dc283bc20e4dd.

The update should be brought back once the engine and animation can load it.

app-object.js
avaer: Yes we do need the transparency. It is used in the previews.

abdullaBest: this is the point that the canvas is completely filled in, therefore alpha is not needed, it is better to fix a black flash than to include alpha, there will be an increase in performance

avaer:
We didn't end up using this.

avaer:
The reason was that the animations delivered were targeted against the wrong rig, making them unusable for this system.

We can revisit when we can get a compatible animations set. For now we are baking FBX with the animation baker script.

PR #1309, opened 2021-06-12
Fixed portal object misidentification.

Benjamin-Dobell:
The inner mesh was being marked as a portal rather than the wrapper Object3D.


avaer:
Good catch!

PR #1325, opened 2021-06-25
add importmap to build.html

madjin:


build.html
```diff
@@ -37,9 +54,9 @@ <h1>XRPK unpack files</h1>
   </form>
 <script src="./bin/geometry.js"></script>
 <script type=module>
-import * as THREE from './three.module.js';
-import {GLTFLoader} from './GLTFLoader.js';
-import {GLTFExporter} from './GLTFExporter.js';
+import * as THREE from 'https://lib.webaverse.com/three.js';
```
avaer, line 57: Actually, this should be `'three'`, and similar for the rest. If import maps are supported that should work.

avaer:
```
import * as THREE from 'three';
```

avaer: https://github.com/webaverse/app/pull/1325/files#diff-1e1cae91d18fbf32b276399bcdce1fc2380580c17323539e9457d0888d3b7b26R13

avaer:
Some more imports normalization requested.

avaer:
Import maps have been removed in main.

PR #1431, opened 2021-08-19
Multiplayer!

AGoblinKing:
- [x] Enables integration with github.com/webaverse/dialog for multiplayer poses
- [x] Uses the eslint-prettier autoformatter, let me know if you don't like any of the styles and we can update the options

biz-gif

Depends on https://github.com/webaverse/dialog/pull/7


AGoblinKing:
Well back to draft, trying out a better way of doing the hashHost passing

avaer:
> Uses the eslint-prettier autoformatter, let me know if you don't like any of the styles and we can update the options
I don't think it's consistent with the ESLint rules, but also: I don't think this is the PR to do it.

We should do a full pass over to code to pretty it to either eslint or prettier, but we shouldn't keep multiple formats and format types, and we shouldn't make that change in an unrelated PR.

AGoblinKing:
I can take the linter out and undo any of its funkier changes. This should be enforcing the existing style guide though, so no file should change if it meets the lint requirements.

avaer:
> This should be enforcing the existing style guide though

Ah you're right, I'll clean up the eslint rules because they are not quite what I think we've been using for the code.

AGoblinKing:
I took prettier out. Should I revert those style changes to the files touched or keep them? My linter is mad at me for the revert per your earlier comment. 
image

AGoblinKing:
image
for comparison

avaer:
> I took prettier out. Should I revert those style changes to the files touched or keep them? My linter is mad at me with the old stuff per your reply.

Yeah we should fix that but not in this issue.

avaer:
One particularly nasty rule is the casing of hex literals.

AGoblinKing:
I use .prettierignore to avoid munging third party libraries. Lets talk about the style in another CL. I can revert the significant style changes here easily.

avaer:
> I use .prettierignore to avoid munging third party libraries.
It's not the libraries, it's the content we paste in to local source files from sources like design templates.

avaer:
I updated the `.eslintrc` to reduce the amount of code changes next time we run it.

AGoblinKing:
Okay that should be easier to read, minus RoomClient which greatly needed some formating.

constants.js
```diff
@@ -40,4 +40,21 @@ export const tokensHost = `https://${chainName}all-tokens.webaverse.com`;
 export const landHost = `https://${chainName}sidechain-land.webaverse.com`;
 export const web3MainnetSidechainEndpoint = 'https://mainnetsidechain.exokit.org';
 export const web3TestnetSidechainEndpoint = 'https://testnetsidechain.exokit.org';
-export const homeScnUrl = `https://webaverse.github.io/street/street.scn`;
+export const homeScnUrl = 'https://webaverse.github.io/street/street.scn';
+
+// Setup dialog URL, use hash if on githubpreview and its an okay hash
+const rGithub = /.githubpreview.dev$/;
```
avaer, line 46: Missing `\.`

constants.js
```diff
@@ -40,4 +40,21 @@ export const tokensHost = `https://${chainName}all-tokens.webaverse.com`;
...
+  /^[a-z][a-z0-9+\-.]*:\/\/([a-z0-9\-._~%!$&'()*+,;=]+@)?([a-z0-9\-._~%]+|\[[a-z0-9\-._~%!$&'()*+,;=:]+\])/.exec(
+    window.location.hash.slice(1),
+  );
+
+const isOKHASH = hashHost && hashHost[2] && rGithub.test(hashHost[2]);
+
+export const dialogUrl = isGH
+  ? isOKHASH
+    ? hashHost[2]
+    : window.location.hostname.replace('3000', '4443')
```
avaer, line 59: This would be better using a regex. We don't want to corrupt domains like `robot3000.club`.

AGoblinKing: it could affect subdomains of the githubpreview.dev only which seems unlikely. We're relying on an undocumented naming convention is the bigger issue imo.

constants.js
```diff
@@ -40,4 +40,21 @@ export const tokensHost = `https://${chainName}all-tokens.webaverse.com`;
...
+const isGH = rGithub.test(window.location.hostname);
+
+const hashHost =
+  /^[a-z][a-z0-9+\-.]*:\/\/([a-z0-9\-._~%!$&'()*+,;=]+@)?([a-z0-9\-._~%]+|\[[a-z0-9\-._~%!$&'()*+,;=:]+\])/.exec(
+    window.location.hash.slice(1),
+  );
+
+const isOKHASH = hashHost && hashHost[2] && rGithub.test(hashHost[2]);
+
+export const dialogUrl = isGH
```
avaer, line 56: Would prefer it as

```
isGH ?
isOKHASH
: 
'dialog.webaverse.com';
```

I don't know how to get the linter to do that yet, but since this is currently inconsistent with the codebase it's a little hard to read.

client/RoomClient.js
```diff
@@ -1,6 +1,6 @@
 import protooClient from './protoo-client/build.js';
 import mediasoupClient from './mediasoup-client/build.js';
-import bowser from'./bowser/src/bowser.js';
+import bowser from './bowser/src/bowser.js';
```
avaer, line 3: Was this file run through a linter or merely copied in.

Since we haven't started linting yet I'd prefer if the original was verbatim.

AGoblinKing: I made these changes to be able to read it, 100s of commented out lines of code. I can revert it and make the minimal changes.

world.js
```diff
@@ -106,36 +106,41 @@ world.connectRoom = async (roomName, worldURL) => {
...
-            },
-          },
-        }));
-        const pose = rigManager.getLocalAvatarPose();
-        channelConnection.send(JSON.stringify({
-          method: 'pose',
-          data: {
-            pose,
+
+    interval = setInterval(() => {
```
avaer, line 110: Why is being live no longer a requirement?

AGoblinKing: live wasn't defined in that scope

avaer: Good catch.

client/RoomClient.js
avaer: Oh I see. Yeah it's better to keep the delta clear, otherwise it will be harder to tell what we actually changed without going through individual commits or deeper code tree understanding tools.

constants.js
AGoblinKing: image
Is this correct?

client/RoomClient.js
AGoblinKing: okay, I reverted roomclient! The good news is I didn't change any functionality just deleted commented out code.

constants.js
```diff
@@ -40,4 +40,23 @@ export const tokensHost = `https://${chainName}all-tokens.webaverse.com`;
...
+    window.location.hash.slice(1),
+  );
+
+const isOKHASH = hashHost && hashHost[2] && rGithub.test(hashHost[2]);
+
+export const dialogUrl = isGH ? 
+isOKHASH ? 
+  hashHost[2] 
+  : 
+  window.location.hostname.replace('3000', '4443')
```
avaer, line 60: It also seems strange to assume 3000 here; there are better ways to add a port, like a window.location.host, which excludes it. Additionally any ports we use should probably be constants.

AGoblinKing: Its part of the hostname ie: https://agoblinking-app-wr9v5wqfvr4-3000.githubpreview.dev/

AGoblinKing: the 3000 is assumed because our index.js uses that 
image

AGoblinKing: location.host is with port location.hostname is w/o port.

avaer: Oh I see, in that case it could be done with a stronger regex.

AGoblinKing: I think the only edge case with 3000 is if someone had it in their name or project title.

 I'll narrow it to avoid those. Are there any others you're aware of?

avaer: I just think it should operate on a whitelist basis, not a replace basis.

```
/\.[0-9]+\.githubpreview\.dev$/
```

avaer: Or, what you have here works. 👍

avaer (approved):

constants.js
AGoblinKing: We'll have to do something fancy with ports and known URLs eventually that makes it easier to switch between dev and production.

avaer:
Keep in mind that we lose hosted rooms support with this change:

https://github.com/webaverse/app/pull/1431/files#diff-7aecd33c076eede220000d2edb73e82e0230c5fcb781bb22f8831b7e37b01b77L246

However, we can/should bring that back.

PR #1446, opened 2021-08-22
Add Sentry

softwarebycl:
This PR adds Sentry error monitoring and tracking.


avaer:
This won't work without a build system, since on Github Pages sentry will not be installed.

This needs a build system first (probably blocked on getting rid of import maps first), unless we can think of some other way like vendorizing all of Sentry or not using the lib.

avaer:
An alternative for now would be to vendorize it.

Issue #1375, opened 2021-07-13
Remove import maps


avaer:
At this point I think the least bad solution is to get rid of import maps and make all nfts import lib.webaverse.com assets, rather than trying to provide a convenient shortcut.

The non chrome browsers are fringe but the benefit to slightly cleaner code is probably not worth ignoring them, and education on import maps themselves (which are still new) is a job we don’t want to be on the hook for.

Note the approach of having a lib is relatively recent and changes a lot.

Fixing this would involve going to all public apps and all html entry points in them, and removing the import map, then patching all files to go to the lib. Then we would need to patch all of the first party nfts to use the lib as well.

PR #1452, opened 2021-08-23
Remove import maps

avaer:
This PR removes import maps from the `html` and `js` files.

External objects will need to be updated to support this.

Fixes https://github.com/webaverse/app/issues/1375.

### DIFF
--- a/CapsuleGeometry.js
+++ b/CapsuleGeometry.js
@@ -1,5 +1,5 @@
-import * as THREE from 'three';
-import {Vector3, Color} from 'three';
+import * as THREE from 'https://lib.webaverse.com/three.js';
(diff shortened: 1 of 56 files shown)


avaer:
Now this needs testing that all of the `html` files are correctly loading.

The external modules still need to be updated.

AGoblinKing (approved):
LGTM. I ran a codespace against it and didn't encounter issues. I only hit index.html, I'll test the others.

AGoblinKing:
image
Looks like editor.html is having issues with html-render

avaer:
> Looks like editor.html is having issues with html-render

Ah yup that needs to be patched!

PR #1463, opened 2021-09-02
Feature NEXT - WSRTC Integration

AGoblinKing:
This PR enables multiplayer using the WSRTC npm module. 

Largely ported what was in the feature-wsrtc branch but now using wsrtc via dependency instead of vendorization. Vite was able to handle serving the workers from wsrtc seamlessly.


AGoblinKing:
Getting a decoder error when trying to use the microphone, back to draft.

AGoblinKing:
https://github.com/webaverse/wsrtc/blob/main/ws-codec.js#L152 needs to be updated to use the import.meta.url and audio works with this PR.

rig.js
```diff
@@ -498,87 +496,66 @@ class RigManager {
...
+    const peerRig = this.peerRigs.get(player.id);
+    if (!peerRig) return;
+
+    const pose = player.pose;
+    const {hmd, leftGamepad, rightGamepad} = peerRig.inputs;
+
+    hmd.position.fromArray(pose.position);
+    hmd.quaternion.fromArray(pose.quaternion);
+
+    if (pose.extra.length === 0) return;
```
avaer, line 509: Why would this ever happen?

universe.js
```diff
@@ -243,7 +243,7 @@ const enterWorld = async worldSpec => {
     }
     if (room) {
       const p = (async () => {
-        await world.connectRoom(room, dialogUrl);
+        await world.connectRoom(room, worldUrl);
```
avaer, line 246: Is there still a need for `room`?

world.js
```diff
@@ -94,173 +89,183 @@ const _bindState = (state, dynamic) => {
 };
 _bindState(states.static, false);
 _bindState(states.dynamic, true);
+
+function makeid(length) {
```
avaer, line 93: This should probably be in `util.js`.

world.js
```diff
@@ -94,173 +89,183 @@ const _bindState = (state, dynamic) => {
...
+  buffer[2] = vec.z;
+
+  if (vec.w !== undefined) {
+    buffer[3] = vec.w;
+  }
+
+  return buffer;
+}
+
+function buffArgs(buffer, ...args) {
```
avaer, line 114: I could be wrong, but I suspect `...` is going to be less efficient than using an array.

If this is core loop function it's probably worth that change.

world.js
```diff
@@ -94,173 +89,183 @@ const _bindState = (state, dynamic) => {
...
+    });
 
-      const newAvatarUrl = avatarUrl || null;
-      const currentAvatarUrl = peerRig.avatarUrl;
-      if (currentAvatarUrl !== newAvatarUrl) {
-        rigManager.setPeerAvatarUrl(newAvatarUrl, avatarExt, peerId);
-        // updated = true;
-      }
+    const update = () => {
+      if (!connected) return;
```
avaer, line 244: I would have canceled the frame to reduce memory leakage but this is ok.

world.js
```diff
@@ -583,70 +589,19 @@ world.getNpcFromPhysicsId = physicsId => {
...
-    track.stop();
-  }
-  networkMediaStream = null;
-};
-
 const micButton = document.getElementById('key-t');
+
 world.toggleMic = async () => {
-  if (!animationMediaStream) {
+  if (!wsrtc.mediaStream) {
```
avaer, line 595: Maybe better to make this a method?

rig.js
AGoblinKing: This was happening early on where pose.extra was blank then later filled. I'll remove and test it.

avaer: Yeah, it just seems if that were to happen it might be a bug elsewhere, or a race condition/uninitialized state.

world.js
AGoblinKing:
```if (this.mediaStream) {
      throw new Error('mic already enabled');
    }```
    this seems to be the check wsrtc uses internally.

Comment (avaer) [reply world.js:594]: > this seems to be the check wsrtc uses internally.  

Yes, but `mediaStream` should be a private variable (to that library).

File: rig.js [hunk 7]
```diff
@@ -498,87 +496,64 @@ class RigManager {
...
-      localEuler.x = 0;
-      localEuler.y += Math.PI;
-      localEuler.z = 0;
-      peerRig.textMesh.quaternion.setFromEuler(localEuler);
-
-      peerRig.rigCapsule.position.copy(peerRig.inputs.hmd.position);
-    }
+  setPeerAvatarPose(player) {
+    const peerRig = this.peerRigs.get(player.id);
+    if (!peerRig) return;
```
avaer, line 501: Would there ever be a player without a rig? If that happens it seems like it's a bug with not setting a rig when we create a player.

I believe the fake rigs can be generated with `new Avatar(null, {})`, synchronously, so this error condition should never happen. IMO if this `if` statement throws an internal error it would be better than silently skipping.

avaer (commented):
Just the one change, the rest looks good. Great work!

rig.js
AGoblinKing: I removed this check after a bunch of testing without it.

world.js
```diff
@@ -94,173 +89,182 @@ const _bindState = (state, dynamic) => {
...
+    const name = makeId(5);
+    wsrtc.localUser.setMetadata({
+      name,
+    });
 
-    /* world.dispatchEvent(new MessageEvent('peersupdate', {
-      data: Array.from(rigManager.peerRigs.values()),
-    })); */
+    interval = setInterval(sendUpdate, 10);
+    wsrtc.enableMic();
```
avaer, line 213: Note that we had the `T` key for talk before, but it seems like that's changing now.

avaer (approved):

world.js
AGoblinKing: the T key should still toggle but I can have it not autoenable.

avaer: No it's ok, this actually makes more sense. We'll be deleting the old UI code anyway as we hook up React.

PR #1510, opened 2021-10-05
Fixed issue of using host instead of hostname

hanzlamateen:


avaer:
Thank you, good catch on this!

PR #1534, opened 2021-10-13
Updated Loaders to load GLTF, VOX, VRM from Metaverse-File-APi 

utf94:
The PR includes following 

**Additions**
1. Changed the Loader of GLTF to point to Metaverse-File-Loader
2. Changed the Loader of VOX to point to Metaverse-File-Loader
3. Changed the Loader of VRM to point to Metaverse-File-Loader 

**Removals**
1. WBN Loader removed <--- Outdated in our architecture.

### DIFF
--- a/screenshot.js
+++ b/screenshot.js
@@ -37,14 +37,14 @@ const _makeRenderer = (width, height) => {
   }));
   scene.add(cubeMesh); */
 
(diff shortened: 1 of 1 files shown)


avaer:
This is not a good name for the PR, and it has no description. I have to guess what it is trying to do from the code.

PR #1569, opened 2021-10-15
Animations.cbor baker

utf94:


avaer:
This looks correct, but was this supposed to run on the backend or the frontend? I think it could just be a `node` script.

abeersaqib:
> @abeersaqib Does this work in app when you replace `animations.cbor` after baking?

@avaer 
It had CBOR issues on using the CBOR from npm, so I switched back to our CBOR instead and made a mjs module out of it as well.

it only lacks reverse animations should we pass them as hard-coded?

avaer:
> it only lacks reverse animations should we pass them as hard-coded?
Whatever the avatar system expects in `animations.cbor` should be in our new `animations.cbor`. That seems to include all of the reverse animations.

PR #1595, opened 2021-10-19
shadow testing

Loryhoof:
Haven't managed to get CSM working correctly yet. Still on it.
Put simple threejs shadows together

For correct shadow display https://webaverse.github.io/atmospheric-sky/ is suggested. On shadows.scn


avaer:
These look good to merge once we fix the regression in avatars.

Issue #1598, opened 2021-10-20
App: configurable hair physics


avaer:
In `avatars.js` we can set the hair physics parameters in the VRM spring bone import override: https://github.com/webaverse/app/blob/f1fc9ceffe899daa66f55536813574547680be4e/avatars/avatars.js#L14

We should tweak this to look great for default avatars, or make it a `.scn` app type which specifies the hair physics parameters of a scene, to allow artists to configure this.

Ideally we expose at least everything `three-vrm` gives us, and we should definitely support wind.

abeersaqib:
@skaljac

PR #1619, opened 2021-10-22
Parameters for configurable hair physics added

skaljac:
This is solution for issue: https://github.com/webaverse/app/issues/1598

.scn file can now have parameter used for heir physics.
That parameter looks like this:

  "hair_physics":{
        "gravity_power_factor": 0.1,
        "stiffness_force_factor": 0.1
  }

Also there is a change in https://github.com/webaverse/metaversefile to support this

### DIFF
--- a/scenes/grid.scn
+++ b/scenes/grid.scn
@@ -1,7 +1,7 @@
 {
   "hair_physics":{
         "gravity_power_factor": 0.1,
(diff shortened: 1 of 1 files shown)


avaer:
The other PR is https://github.com/webaverse/metaversefile/pull/11/files; this should be linked in the PR description.

Issue #1617, opened 2021-10-22
App: render pass support


avaer:
For full screen effects like bloom/DOF, we will need to support render passes in our main loop. The regular THREE.js render pass approach should work AFAIK, so we should just set that up.

To test that it works, we can add a togglable color filter shader in the code, which will allow us to experiment.

The full screen shaders will later be user-activatable, so the step after this is to figure out the API for shaders to use (we will probably support ShaderToy fomat).

PR #1679, opened 2021-10-27
Post-processing

skaljac:
This is PR related to https://github.com/webaverse/app/issues/1617

Post-processing can be toggled via button 'k'


avaer:
Implemented the full version in https://github.com/webaverse/app/pull/1689

PR #1686, opened 2021-10-28
Wind parameters from .metaversefile

skaljac:
I have change everything commented in previous PR.


rig.js
```diff
@@ -59,6 +69,23 @@ const localRaycaster = new THREE.Raycaster(); */
   return geometry;
 })(); */
 const _makeRig = app => {
+
+
+    function initWindParameters(rig) {
+        var hairPhysics = rig.app.getComponent('hairPhysics');
```
avaer, line 75: Hair physics is not the best name, since all avatars should have hair physics. This is about wind.

rig.js
```diff
@@ -59,6 +69,23 @@ const localRaycaster = new THREE.Raycaster(); */
...
+    
+        if (!hairPhysics) {
+            return;
+        }
+
+        if (!hairPhysics.hasOwnProperty('windDirectionX')) hairPhysics.windDirectionX = 0.0;
```
avaer, line 81: Shouldn't the wind direction be gotten from the component?

avaer: Components have a `value` which you can get look at.

skaljac:
I have created new PR with fresh copy of webaverse:

https://github.com/webaverse/app/pull/1893

That is the reason for closing this PR

All requests are addressed there.

PR #1693, opened 2021-10-30
Per-scene render settings

avaer:
This PR allows per-scene render settings using the `metaversefile` `rendersettings` type:

```
{
      "content": {
        "fog": {
          "fogType": "exp",
          "args": [[255, 255, 255], 0.01]
        },
        "ssao": {
          "kernelRadius": 16,
          "minDistance": 0.005,
          "maxDistance": 0.1
        },
        "dof": {
          "focus": 3.0,
          "aperture": 0.00002,
          "maxblur": 0.005
        },
        "hdr": {
          "adaptive": true,
          "resolution": 256,
          "adaptionRate": 100,
          "maxLuminance": 10,
          "minLuminance": 0,
          "middleGrey": 3
        },
        "bloom": {
          "strength": 0.2,
          "radius": 0.5,
          "threshold": 0.8
        }
      }
    },
```

Depends on https://github.com/webaverse/metaversefile/pull/20

### DIFF
--- a/equipment-render.js
+++ b/equipment-render.js
@@ -45,8 +45,8 @@ class EquipmentRender {
     this.previewRenderer.autoClear = false;
     this.previewRenderer.sortObjects = false;
     this.previewRenderer.physicallyCorrectLights = true;
-    this.previewRenderer.outputEncoding = THREE.sRGBEncoding;
-    this.previewRenderer.gammaFactor = 2.2;
+    // this.previewRenderer.outputEncoding = THREE.sRGBEncoding;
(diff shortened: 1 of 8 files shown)


avaer:
We will support limiting the post processing per scene with user-overridable performance settings that guess the spec of the machine that the user is on and guess accordingly.

Something like...

- Low: No effects, just render()
- Medium: Adds shadows
- High: Adds AA
- Supernal: Adds postprocessing

avaer:
In the future LODs will be changeable this way as well.

PR #1697, opened 2021-10-31
Three-vrm exponential hair

avaer:
This PR makes [`three-vrm` hair](https://github.com/pixiv/three-vrm/blob/3dcb963c65d9cedcf2a40afc3aa165b8545af4b0/packages/three-vrm/src/springbone/VRMSpringBone.ts#L241) exponential, so it is frame-time independent. Due to being an IK chain it still looks different if your frame rate drops, but it should be a lot better.

We copy out their update method and hack it to make it exponential. Perhaps this could go upstream, in `three-vrm`, since it's more correct.

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -61,14 +61,13 @@ VRMSpringBone.prototype.update = (_update => {
     localMatrix2.multiply(this._getParentMatrixWorld());
 
     // several parameters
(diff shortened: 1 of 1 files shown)


avaer:
Related: https://github.com/webaverse/app/pull/1698/files

avaer:
We have much better effects now, including wind!

PR #1712, opened 2021-11-01
Adding Mocha Tests Action to test scenes and build instructions for CodePipeline

rob-webaverse:


test/loading/index.js
```diff
@@ -0,0 +1,300 @@
...
+
+  async setupRequestInterception(){
+    await this.page.setRequestInterception(true);
+    this.page.on('request',this.requestListener);
+    this.page.on('requestfailed', this.requestFailedListener);
+    this.page.on('requestfinished', this.requestFinishedListener);
+  }
+
+  async testScene(sceneUrl) {
+    await sleep(5);
```
avaer, line 236: Why the sleep?

avaer: If there is an actual condition we are waiting for, wait for that. Don't count on it being done in 5 seconds.

test/loading/index.js
```diff
@@ -0,0 +1,300 @@
...
+const {performance} = require('perf_hooks');
+const scenes = require('../../scenes/scenes.json');
+
+function sleep(seconds) {
+  return new Promise(resolve => setTimeout(resolve, seconds * 1000));
+}
+
+
+const ignoreErrors= [
+  /** Internal ThreeJS error sometimes gets caught in the extendTexture */
```
avaer, line 12: Why? We should fix that if it's happening.

avaer (changes requested):
Timing clean up and questions, but the core looks good.

test/loading.js
```diff
@@ -0,0 +1,43 @@
...
+      })
+
+      appTester.MochaIntercept = ()=>{
+        error = true;
+      }
+
+      try{
+        await appTester.run();
+      }catch(e){
+        //digest pupeteer crash error that comes up very rare.
```
avaer, line 24: Don't throw it away then, at least log it...

test/loading/index.js
```diff
@@ -0,0 +1,331 @@
...
+            if(self.lastCheckedAt - self.lastActivityAt > 10000){
+              clearInterval(outInterval);
+              resolve();
+            }  
+          }
+          self.lastCheckedAt = performance.now();
+
+          
+        }catch(e){
+
```
avaer, line 101: Throwaway catch is not good, we should log it.

test/loading/index.js
```diff
@@ -0,0 +1,331 @@
...
+  
+  MochaIntercept(){
+
+  }
+
+  PromiseIntercept(value){
+    let reportError = false;
+    try{
+    }catch(e){
+    }finally{
```
avaer, line 160: Missing error log?

test/loading/index.js
```diff
@@ -0,0 +1,331 @@
...
+      try{
+        self.stats.errors.push(value);
+      }catch(e){
```
avaer, line 163: Missing error log?

avaer (changes requested):
We shouldn't throw away errors, but at least log them so they are surfaced to devs.

avaer (approved):

Issue #1710, opened 2021-11-01
App: convert avatar url to y.js tracked


avaer:
Right now avatar `url`s are tracked by `rig.js` but they should be in the `Player`'s y.js `.state`, which is then observed to create the change in avatar.

That is necessary for network syncing avatar changes.

avaer:
Related: https://github.com/webaverse/app/issues/1709

avaer:
Done, it's all tracked in y.js now.

PR #1722, opened 2021-11-02
Unify world + player state

avaer:
Merge them under a single y.js instance for easy transactional app transplant. We can detect app moves across app managers at the time the y.js observer fires change events, and this should work regardless of network jitter.


avaer:
We use a prefixing mechanism to internally distinguish the keys.

avaer:
Related: https://github.com/webaverse/app/issues/1710

PR #1727, opened 2021-11-02
Update scenes.json

abeersaqib:
Pointed the scene to 

shadows.scn


avaer (changes requested):
This was just to cover up for https://github.com/webaverse/app/pull/1728

PR #1728, opened 2021-11-03
Incorrect start_urls fix

Loryhoof:


avaer:
If you actually look into vite.js loading you should be able to see that it is duie to not finding the .metaversefile, obviously it's due to the .nojekyll in that repo: https://github.com/webaverse/street/commit/128f9f80a1895cada5d4acfca07568691fe69997

avaer:
I also don't think we want the anime skybox here.

avaer:
Similar for skybox: https://github.com/webaverse/skybox/commit/6ca5137a8e0a37ea09b1fbef0325c655ad6c4378

avaer:
If that works we can close this PR.

avaer:
Should be fixed now.

PR #1734, opened 2021-11-03
Avatar first class citizens

avaer:
This PR makes avatar objects first class citizens in the world, following world object rules.

We use the automatic transactional app transplant logic to move the ownership of avatars between world and player context, so it is kept track of for multiplayer/world crossing. Additionally, we try to make the logic robust against player state conflicts (two players try to wear the same avatar).


avaer:
Note that the world and players have their own app managers, and the player avatar now lives in the player's app manager, not the world's.

PR #1736, opened 2021-11-03
Remove rig manager

avaer:
This PR moves the last of the logic that existed in `rig.js` to `avatars.js` (velocity/direction tracking).

Now the player object is the source of truth about a player, and there is no real manager of them. Each player has its own app manager which holds its apps, including the avatar app and wearables.

All of the Player objects are bound to the the y.js CRDT which can be replicated. The idea is that you simply update this to get multiplayer, since the rest of the state is derived from it.

Depends on https://github.com/webaverse/metaversefile/pull/21


avaer:
This PR also put the Player -> Avatar binding in a clean place, `avatar-binding.js`.

avaer:
`rig.js` remains in `deadcode/` for now.

avaer:
Where you used to do `rigManager.localRig`, we should now do `localPlayer.avatar`, which is the same `Avatar` object.

PR #1739, opened 2021-11-04
ccd and character controller

Loryhoof:


physx.js
```diff
@@ -612,6 +616,75 @@ const physxWorker = (() => {
       } : null;
     }
   };
+  w.groundRaycastPhysics = (physics, p, q) => {
```
avaer, line 619: I do not think we should be doing this to detect groundedness. Groundedness should be based on support restitution.

scenes/shadows.scn
```diff
@@ -15,39 +15,7 @@
...
-          "middleGrey": 3
-        },
-        "bloom": {
-          "strength": 0.2,
-          "radius": 0.5,
-          "threshold": 0.8
-        }
-      }
-    },    
+    },
```
avaer, line 18: Probs shouldn't be editing that here.

scenes/treehouse.scn
```diff
@@ -75,13 +43,13 @@
         0,
         0.7071067811865475
       ],
-      "physics": false,
+      "physics": true,
```
avaer, line 46: These do nothing.

scenes/treehouse.scn
```diff
@@ -61,12 +29,12 @@
         0,
         0
       ],
-      "start_url": "https://webaverse.github.io/skybox-anime/"
+      "start_url": "https://webaverse.github.io/atmospheric-sky/"
```
avaer, line 32: If we want to change this, it shouldn't be in this PR.

physx.js
```diff
@@ -511,14 +511,17 @@ const physxWorker = (() => {
     const quaternions = scratchStack.f32.subarray(index, index + maxNumUpdates*4);
     index += maxNumUpdates*4;
     const scales = scratchStack.f32.subarray(index, index + maxNumUpdates*3);
-    index += maxNumUpdates*7;
+    index += maxNumUpdates*3;
```
avaer, line 514: This is a good bugfix, I pushed it upstream: https://github.com/webaverse/app/commit/1f1b392d462ea86436b3c458e79303f1558a6705

character-controller.js
```diff
@@ -480,6 +480,17 @@ function update(timeDiff) {
   localPlayer.actionInterpolants.jump.update(timeDiff);
   localPlayer.actionInterpolants.dance.update(timeDiff);
   localPlayer.actionInterpolants.throw.update(timeDiff);
+
+
+  if(localPlayer.appManager.apps[0]) // Pushing updates to physicsManager
```
avaer, line 485: We can't assume that the `apps[0]` is the avatar. It is actually `localPlayer.avatar` which is the bound `Avatar` and you can get the `app` that of it that way. It may be `null`.

avaer:
This should be using axis locking to get a correct capsule simulation: https://gameworksdocs.nvidia.com/PhysX/4.1/documentation/physxguide/Manual/RigidBodyDynamics.html#axis-locking

avaer (changes requested):
I think we need to

1) Remove updates from the `simulate` step and do them in-place with methods
2) Use physx-level axis locking to achieve capsule uprightness without hacks

Loryhoof:
ok ser will fix

PR #1745, opened 2021-11-04
god rays example

skaljac:
This is work in progress.
Exact copy of three.js example.


avaer (changes requested):
This code looks fine, but it belongs in https://github.com/webaverse/metaversefile/blob/main/type_templates/rendersettings.js rather than `app`. All user post processing goes there.

avaer:
This has been abandoned.

PR #1747, opened 2021-11-05
App: add test multiplayer room

avaer:
This isn't much of a "gunroom" (currently just the old treehouse) but I needed a clean scene to test with multiplayer (created as a default room). We can update it with new art later.


avaer:
git push --set-upstream origin gunroom

avaer:
https://www.youtube.com/watch?v=Y70vcs3oV14

PR #1752, opened 2021-11-05
Initialize serverside multiplayer room

avaer:
We start up a multiplayer room along with the server for test purposes. This PR populates it with `gunroom.scn` for now.


avaer:
Requires another https://github.com/webaverse/wsrtc PR to take effect.

Issue #1681, opened 2021-10-27
App: do not track physics objects in physics manager


avaer:
Physics objects should be tracked by apps, not the physics manager (duplicated state). We should be able and to get rid of the physics manager collecting objects and simply look them up on the apps: https://github.com/webaverse/app/blob/cb3bb01e88c73e9ac40bb420f779059d2f766de5/physics-manager.js#L178

PR #1756, opened 2021-11-06
removed physicsObjects array from physics-manager.js

Loryhoof:
https://github.com/webaverse/app/issues/1681

### DIFF
--- a/metaversefile-api.js
+++ b/metaversefile-api.js
@@ -99,6 +99,9 @@ class App extends THREE.Object3D {
   getPhysicsObjects() {
     return this.physicsObjects;
   }
+  getPhysicsObject() {
+    return this.physicsObjects[0];
+  }
   activate() {
     this.dispatchEvent({
       type: 'activate',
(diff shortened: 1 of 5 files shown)


avaer (changes requested):
IMO `metaversefile-api` should gain the ability to find the exact physics object from the apps, rather than finding the matching app and then plucking the first physics object from there.

avaer:
Basically we need a `metaversefileApi.getPhysicsObjectByPhysicsId(physicsId)` which looks in the app mangers' apps for it.

metaversefile-api.js
```diff
@@ -99,6 +99,9 @@ class App extends THREE.Object3D {
   getPhysicsObjects() {
     return this.physicsObjects;
   }
+  getPhysicsObject() {
```
avaer, line 102: Can be deleted?

avaer (approved):

avaer (approved):

PR #1762, opened 2021-11-09
MetaMask + Discord integration - with Weba-Wallet

mujabmuneeb:
Added Login modal with two buttons, MetaMask, and Discord respectively.
Integrated Discord Login
Integrated Weba-Wallet storage module.

Flow: 
When the user logs in, App calls the setAddress setter to set the user's address and sendDataToWallet function to save the user's privateKey in Weba-wallet secure storage.
sendDataToWallet function opens hidden iframe and sends postMessage to weba-wallet along with user's privateKey.
Weba-wallet validates the app's origin receives privateKey, encrypts it, and saves it in localStorage.

For the next time, the user doesn't have to login again because on login App will request Weba-wallet for the user's privateKey using fetchWalletData('privateKey').
If Weba-Wallet validates App origin, it will decrypt privateKey and returns it. And by using that privateKey, App will fetch the user's address and set it.

[Flow Diagram](https://i.ibb.co/Yfgw5N8/Wallet.jpg)


constants.js
```diff
@@ -43,6 +43,7 @@ export const aiHost = `https://ai.webaverse.com`;
 export const web3MainnetSidechainEndpoint = 'https://mainnetsidechain.exokit.org';
 export const web3TestnetSidechainEndpoint = 'https://testnetsidechain.exokit.org';
 export const worldUrl = 'worlds.webaverse.com';
+export const discordAuthUrl = `https://discord.com/api/oauth2/authorize?client_id=684141574808272937&redirect_uri=${window.location.origin}%2Flogin&response_type=code&scope=identify`
```
avaer, line 46: Hm, I don't think this should be a constant, it is very specific to Discord logic.

`discordClientId: 684141574808272937` make sense to me here, but the URL synthesis probably belongs with the login code.

src/components/modal.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import "./modal.css";
+import PropTypes from "prop-types";
```
avaer, line 3: I said this before, but this is not the PR to introduce `PropTypes`. We can introduce it in another PR.

Note that because we run the devserver in production, this will actually slow down production if we include it without thinking.

avaer (changes requested):
Overall good, but needs some tweaks.

src/components/modal.js
avaer: We also don't need it in `package.json` in this PR.

index.js
```diff
@@ -49,7 +49,11 @@ function makeId(length) {
         req.originalUrl = u;
         next();
       }
-    } else {
+    }else if(/^\/login/.test(o.pathname)){
```
avaer, line 52: Why handle this route by redirecting? Since it is not doing anything, can we get rid of it?

src/User.jsx
```diff
@@ -0,0 +1,322 @@
...
+import {storageHost, accountsHost, tokensHost, loginEndpoint, discordClientId, walletHost} from '../constants';
+import {contracts, getAddressFromMnemonic} from '../blockchain.js';
+import {jsonParse, parseQuery, handleDiscordLogin} from '../util.js';
+import Modal from "./components/modal";
+
+function useComponentVisible(initialIsVisible, fn) {
+  const [isComponentVisible, setIsComponentVisible] = useState(initialIsVisible);
+  const ref = useRef(null);
+
+  const handleClickOutside = event => {
```
avaer, line 16: For this appear/disappear, can we use the existing infrastructure rather than adding more global event handlers?

avaer: https://github.com/webaverse/app/blob/b06ba4455f2f4ff1dfb627c2f78f722c5b8aba48/src/Header.jsx#L494

src/User.jsx
```diff
@@ -0,0 +1,322 @@
...
+
+    iframe.postMessage(JSON.stringify({'temp_token': sessionStorage.getItem('temp_token'), 'token': address}), walletHost);
+
+    window.addEventListener("message", (event) => {
+      if (event.origin !== walletHost)
+        return;
+    
+      setWalletData(event.data.data);
+
+      window.removeEventListener('message', ()=>{});
```
avaer, line 136: This isn't how you remove an events listener.

src/User.jsx
```diff
@@ -0,0 +1,322 @@
...
+      iframe.close();
+
+    }, false);
```
avaer, line 139: Could be `{once: true}`.

avaer (changes requested):
Cleanups requested

avaer:
This PR also needs a description.

index.js
mujabmuneeb: This will be removed after the new discord update. In which login has been removed.

mujabmuneeb:
> Is this a hardcoded password? If so it should at least be a constant.

It is now being handled by weba-wallet.

src/User.jsx
mujabmuneeb: > 

Fixed, updated removeEventListener code.

mujabmuneeb:
> This PR also needs a description.

Added description and updated Title.

src/User.jsx
```diff
@@ -0,0 +1,263 @@
...
+          setShow(false);
+        }
+        else {
+          setLoginError(String(error).toLocaleUpperCase());
+        }
+      }
+      else {
+        fetchWalletData('privateKey');
+      }
+  }, [address, setAddress]);
```
avaer, line 195: I think this `setAddress` is probably pointless?

avaer: Actually never mind, I can see how the way the code is structured it could be needed!

avaer (approved):

Issue #1690, opened 2021-10-30
App: handle render passes resize


avaer:
Right now it's not completely handled, causing resolution glitches on resize.

skaljac:
Solved

PR #1768, opened 2021-11-09
Render pass resize

skaljac:
Fix for issue:
https://github.com/webaverse/app/issues/1690


avaer:
I think this fails to take into account the pixel ratio? It is blurry on my 4K.

avaer (approved):
Thanks!

PR #1783, opened 2021-11-11
Camera pressure issue

### LINKED ISSUE #1731: App: camera pressure
Since the character controller is becoming good (such as not allowing tunneling), the camera is now becoming an annoyance because it doesn't feel very gamelike and is very easy to cheat.

We will re-enable minecraft-style mining at some point soon, and our camera is especially bad for walking through tunnels, where you can see neighboring veins and cheat.

To fix the camera, I think a good start is to backwards raycast from the player eyes to the camera each frame, and see if there is a hit. If there is a hit then we quickly exponentially move the camera in front of the obstruction to provide the illusion of a closed space.

Note that the camera should _not_ lock into place at the intersection distance automaticaly. If you are just moving your mouse through an obstruction, it should zoom in for the obstruction and then automatically zoom back out to the far distance when it is removed. Only when the user does a manual zoom adjustment (mouse wheel) should we lock the camera distance to the new one (which was possibly obstructed). This is important for good feeling gameplay.

skaljac:
Resolve #1731

Issue
https://github.com/webaverse/app/issues/1731

We have 4 raycasts from camera corners.
Based on that, we move camera towards the player.


camera-manager.js
```diff
@@ -1,11 +1,15 @@
...
 import {getRenderer, camera/*, orbitControls*/} from './renderer.js';
 import * as notifications from './notifications.js';
+import metaversefile from 'metaversefile';
 
 const localVector = new THREE.Vector3();
 
 const cameraOffset = new THREE.Vector3();
 let cameraOffsetTargetZ = cameraOffset.z;
+let cameraOffsetZ = cameraOffset.z;
+const raycaster = new THREE.Raycaster();
```
avaer, line 11: We should use our own physics raycasting solution, which is the actual physics.

I think the method used here will slow down the app like crazy...

avaer (changes requested):
Should be based on our physics, not the THREE.js scene.

avaer:
Here is the interface to our PhysX raycaster which can be used: https://github.com/webaverse/app/blob/1f1af692e449d536e2b72443eb379b9ccf6764e0/physics-manager.js#L198

camera-manager.js
skaljac: Ok, I will use physics for raycast in next PR

skaljac:
I have implemented physx racyast insted of THREE.raycast

avaer (approved):
This works great, thank you.

I have some improvements to suggest as a follow up but this is good enough to merge. Will put the rest in another issue.

avaer:
This has some conflicts to resolve before we can merge it.

avaer (approved):

PR #1787, opened 2021-11-11
Test Cases Logs - with Discord Bot Integration

mujabmuneeb:
Getting test-logs in an array, scene by scene.
Integrated Discord bot. (Need live Webhook URL)


test/loading/index.js
```diff
@@ -0,0 +1,308 @@
...
+          console.warn(e);
+
+        }finally{
+          retries--;
+        }
+      }, 1000);
+    });
+  }
+  
+  waitForNetworkIdle = ()=>{
```
avaer, line 95: I know this isn't used but I just want to say this will not work for scenes testing. Apps in scenes can totally succeed and still keep the network busy. E.g. loading an infinite list.

test/loading/index.js
```diff
@@ -0,0 +1,308 @@
...
+
+  async init() {
+    await this.waitForServerBoot();
+    this.browser = await puppeteer.launch({
+      headless: true, // change to false for debug
+      defaultViewport: null,
+      ignoreHTTPSErrors: true, 
+      dumpio: false,
+      args: [		
+      '--disable-gpu',
```
avaer, line 175: Btw if we ever use GPU machines we will need to remove this flag.

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: APT Update
+        run: sudo apt update -y
+      - name: Install Dependencies
+        run: sudo npm install -y
+      - name: Pre-Build
+        run: |
+              # Install apt-transport-https
+              sudo apt-get install -y apt-transport-https
+              # Use apt to install the Chrome dependencies
+              sudo apt-get install xvfb
```
avaer, line 43: I would list all of these on one line so install is faster/more parallel.

avaer (approved):

PR #1797, opened 2021-11-12
Homespace scene

skaljac:
Homespace scene should represent main scene for future release.

Homespace.glb is set up to use physics, while grass.glb is without physics component.

We also have Fox that user can control.

Two particle effects used for plant decoration.

Firefly simulation is done also with particles. For now fireflies only move avay from player, but later they will twinkle and be somewhat smarter.

### DIFF
--- a/scenes/homespace.scn
+++ b/scenes/homespace.scn
@@ -1,0 +1,88 @@
+{
+  "objects": [
+    {
+      "type": "application/rendersettings",
+      "content": {
+        "background": {
+          "color": [0, 0, 0]
+        },
+        "fog": {
+          "fogType": "exp",
+          "args": [[6.3, 4.35, 9.7], 0.001]
+        },
+        "dof": {
+          "focus": 3.0,
+          "aperture": 0.00002,
+          "maxblur": 0.005
+        },
+        "hdr": {
+          "adaptive": true,
+          "resolution": 256,
+          "adaptionRate": 100,
+          "maxLuminance": 10,
+          "minLuminance": 0,
+          "middleGrey": 3
+        },
+        "bloom": {
+          "strength": 0.25,
+          "radius": 0.5,
+          "threshold": 0.8
+        }
+      }
+    },
+    {
+      "type": "application/light",
+      "content": {
+        "lightType": "ambient",
+        "args": [[46, 113, 4], 0.1]
+      }
+    },
+    {
+      "type": "application/light",
+      "content": {
+        "lightType": "directional",
+        "args": [[255, 255, 255], 2],
+        "shadow": [50, 4096, 0.1, 1000, 0, 0.2]
+      },
+      "position": [6,10,0]
+    },
+	  {
+      "position": [0,0,0],
+      "start_url": "https://webaverse.github.io/night-skybox/"
+    },
+    {
+      "position": [0,0,-8],
+      "quaternion": [0,0.7071,0,0.7071],
+      "start_url": "https://webaverse.github.io/homespace_new/homespace.glb"
+    },
+    {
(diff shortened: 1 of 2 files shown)


yaserOSource:
Is the goal of this PR to add a new homespace scene? 

I would need a link to the original issue or story to be able to review this PR correctly. 

Otherwise, I tested the scene, and i could see the fireflies, they seem to be raining down, is this the intended effect? They appear at a certain height and then drop down.

The skybox looks pretty cool,  I wish it was generated directly from the shader rather than textures since you could see the image compression. 

Its also interesting there has not been any change in the scene environment map? Or is this being handled somewhere else?

skaljac:
I don't believe there is issue for this task. It was intended for site release.

Firefly system is still in development (that is separate gh project) and it will change how that looks.

Skybox is basic night skybox. It is done this way for performance reason. And yes, we can change that if it is needed.

yaserOSource (approved):
I would say this looks pretty good in that case, Im not seeing a huge amount of issues with it, aside from the aforementioned.


I would like to see how the scene would react if you applied the final skybox texture to the scene environment map, and with the overall lighting, it would look interesting or maybe not at all, we would need to see.

amnaarshed4224:
1-The bag collectable allows multiple collections? even after collecting one, There still is option to collect infinite numbers.

bag collectable can be picked up infinitely

2-The number of Particles/Fireflies  appear too much . The  frequency in my opinion should be reduced.
3-The particles/fireflies move but disappear after certain time. They should remain there. Not sure if fireflies are part of the home space?
4-The Collectable under the table, it doesn't gets collected whatever it is.

Collectable under table cant be picked up

Overall the homespace looks pretty cool and complete.

amnaarshed4224:
1-When Tried to collect the invisible collectables, The avatar simply falls down through the scene. (Though we can fly back)

https://user-images.githubusercontent.com/35926530/142292913-fd4edf67-9672-47d6-ba4e-85b0f66e077f.mp4

https://user-images.githubusercontent.com/35926530/142292916-94b775d1-a87e-4821-852b-64351fedb923.mp4


2-Whenever the avatar stumbles, it falls down through the scene.

https://user-images.githubusercontent.com/35926530/142292917-4d8c803a-5788-4eec-9ef8-540d5acb22d7.mp4


3-The staircase is very hard to climb, although climbable but its roundness and distance(height) of each step makes it difficult(needs jumping to climb each step)

https://user-images.githubusercontent.com/35926530/142292902-fada7698-9528-4b9f-8e56-d000d9ec7804.mp4

4-The avatar looks embedded into the stairs when tried to pass under it.

WhatsApp Image 2021-11-18 at 3 42 13 AM

5-The avatar stumbles when touches with wall with sword.(passes through the wall and reaches other room)

https://user-images.githubusercontent.com/35926530/142292911-a75d2693-0838-42c0-b8cd-cc516e80d356.mp4

6-Avatar falls down when it touches with the armor filled cupboard

https://user-images.githubusercontent.com/35926530/142292890-2a9068b8-4f78-473e-9138-a7f025d0b92e.mp4

7- The camera view due to congestion of rooms and furniture, stairs etc doesn't shows avatar properly.

https://user-images.githubusercontent.com/35926530/142292905-9a21df97-6c63-4056-b4ef-bc8f509d5c60.mp4

8-While flying/running, the avatar goes through the objects and in some cases falls down

https://user-images.githubusercontent.com/35926530/142292908-551bcc9e-7e3d-4f98-8eae-105ba8cf6de5.mp4

Overall the scene looks good, with proper lightning inside the house its great. rooms appear a  bit congested.























.

avaer:
This didn't make it in.

Issue #1536, opened 2021-10-13
App: retarget FBX animations to VRM


avaer:
The goal is to make our animations look great and not glitched.

Our animations right now are janky, because we use the quaternions from Mixamo FBX tracks to animate a VRM skeleton. In some cases this makes gameplay impossible, since the avatar cannot point a gun in the correct direction.

Note we do _not_ need to retarget names or the skeleton itself, we already do that successfully in `avatars.js`. This is purely about better proportion-preserving rotation maths when applying the bone transforms.

It seems `THREE` `SkeletonUtils` can retarget bone sets, the place to start is there and then we debug it: https://github.com/mrdoob/three.js/blob/master/examples/jsm/utils/SkeletonUtils.js

avaer:
animations loaded and applied in avatars.js: https://github.com/webaverse/app/blob/169ab5a4b2d7a5be2e6223514a0e42b902ea6663/avatars/avatars.js#L124
applying selected animations: https://github.com/webaverse/app/blob/169ab5a4b2d7a5be2e6223514a0e42b902ea6663/avatars/avatars.js#L1914

avaer:
Related: https://github.com/webaverse/app/issues/1537

avaer:
The code for doing retargeting is here: https://github.com/webaverse/app/blob/96a437ce5161671ba153d511ea695fa4e4c89e37/avatars/util.mjs#L707

However it was disabled in this commit: https://github.com/webaverse/app/commit/7385f45eaa7597e88c7ebdc4f1b88c1de47bb73e

It was disabled because retargeting all animations for all avatars might not work so well until we have a backend solution for doing so. The code is abstracted enough that it should work on the server.

PR #1815, opened 2021-11-15
Avatars retarget FBX -> VRM

avaer:
Currently attempting a live retarget.

Related to https://github.com/webaverse/app/issues/1536.


avaer:
I implemented retargeting which forces world quaternions relative to the source animation, while keeping positions from the destination skeleton.

The math on this should be correct, or at least the math is now broken out so we can experiment with different retargeting strategies.

This is only retargeting the idle animation right now and it is computationally intensive to retarget the animations. Thankfully it only needs to be done once per avatar and can easily be done offline, so we can add server-side support for computing the animations retarget for an avatar.

avaer:
We have the retargeting working with the idle animation, online in the app. However, we will need to find some other way to retarget the animations -- for example on the server at request time if we want to use this solution.

I am not convinced it is currently worth the effort and many extra bugs which integrating this will cause right now, so I will merge this PR (since it cleans up the avatars code) and we can try to optimize animations again later by reviving this code.

PR #1816, opened 2021-11-16
CCD

Loryhoof:
Needs QA (DO NOT MERGE)


amnaarshed4224:
Testing Specs
WIndows
chrome
Nividia GT730, 8gb ram, core2 quad 2.66 GHZ				


(1)Does FPS affect avatar movement in any way?

Yes and no.
-On one system The movement is jittery and there is lag in walking, running and even flying even if the key is kept pressed(30 fps).
-on one of the devices the avatar is running fine without jittering(videos attached)


(2)Can you go through any objects?

a)-WHile walking in majority of the cases the avatar was not able to pass through the objects(mirror, wall,sofas, lifts, benches etc)
b)-while running and flying the avatar can pass through the collectables
c)-also there is a bit of lag as the avatar gets stuck when it collides into certain objects such as stairs,steps.
d)-while flying the avatar sometimes passes through the objects, in one case it passed through the wooden wallls
e)-also when trying to pass through the objects the avatar appears to be a bit of embedded into the object(half in,half out),
f)-in some cases while flying the avatar can pass through the roof and in some it can't(wooden cabins: appears half in half out while trying to pass through the roof).
g)-the avatar while running is able to pass through tables in the the street screen that is probably when an object concides with a collectable(collectable and the object are in the same place)
h)-Avatar isnt collecting any collectables by pressing E in any of scenes.
i-)avater is not showing E in street scn when our current avatar approaches it.


(3) Does it break the game and when?
a)-Trying to ride any object/vehicle in any of scenes crashes.

(Others)
a)-The avatars find difficuilty in standing and walking on slopes(hills,dunes,steps etc)
b)-Trying to ride the vehicle in shadows doesn't works

https://user-images.githubusercontent.com/35926530/142085093-0c6d7237-6032-4dc6-be7a-3e7b98bf1534.mp4


https://user-images.githubusercontent.com/35926530/142085117-6e192a8d-4232-4544-ab95-1426e3ee2e4d.mp4


c)-The feet of the avatar do not appear on the ground in the latest update.

Loryhoof:
Hey, thank you for the review.

1.
_-On one system The movement is jittery and there is lag in walking, running and even flying even if the key is kept pressed(30 fps).
-on one of the devices the avatar is running fine without jittering(videos attached)_

Could you list the specs for the second device without jitter? Also the performance is pretty bad (7-10 FPS), how much do you get when not recording? I think GT730 is not going to be our target audience.

2. _avater is not showing E in street scn when our current avatar approaches it._
- This is probably a capsule size/height issue, will be fixed. You should be able to to activate it when in first person and looking at the avatar's feet.

3. _The avatars find difficuilty in standing and walking on slopes(hills,dunes,steps etc)_
- Could you elaborate on that? Do you mean jitter/wobble? Sliding down the hill? (for hills/dunes)
- Steps will be fixed, it's currently also a problem in the non-CCD version.

4. _Trying to ride any object/vehicle in any of scenes crashes._
- Vehicle code is currently outdated, I should've mentioned that, sorry.

amnaarshed4224:
1- The movement was not jittery on GT730, 8gb ram, 2.66ghz quad core. The performance I get at max with this machine is around 30-35fps. The jittering happened while I was accessing another device through TeamViewer. Yes I am aware the performance of gt730 is pretty low I will try to arrange a high end device soon.

2-Tried again with looking at avatars feet but still response is same as before. Tried from different angles too.

3-Yes the avatar feel as if it wobbles. In some cases it keeps sliding through the slopes too. Not sure if that's a bug or a feature.

PR #1820, opened 2021-11-17
CCD QA fixes

Loryhoof:

How does it work?
- We pass our player velocity into physX, which runs a simulation (we can set parameters like friction/gravity/bounciness etc.)
- We then call physX again to get the updated character capsule transform, then apply it to our avatar.

### Testing

2. Crouch walk down/up hills and steps (Heavy jitter? Fall through ground? Get stuck?)
3. Test switching scenes multiple times, while equipping weapons and changing avatars.

### NEW Testing Nov 19

2. Jump (Spacebar) actually didn't work before on flat surfaces, and I'm not sure how nobody noticed, but its fixed now please test on hills and flat areas.
3. Feet not stuck in ground anymore
4. Now that switching avatars works, please test if you can consistently change it and you spawn at the correct position (Where you held the `activate` button) before it would sometimes teleport you back to origin spawn (0,0,0)


amnaarshed4224:
1-Yes I am able to equip the collectables now. 

2-while walking on steps, the movements is not smooth enough, not bad but there feels a bit of wobble while climbing steps)
  -on dunes/hills(on top flat surface) the avatar even while standing keeps moving its feet without changing its position, also on 
    a slope, the avatar skids down side.

3-Crashed while trying to equip the lamp thing, firefly thing in darkness scene(Screenshot attached with log)
crashes while trying to pickup firefly collectable thing
Crashes while trying to pickup lamp collectale thing

  -Crashes when avatar falls while walking/running fast. It passes through the floor(dunes) and screen is blank then only 
    shadows are there no movement.

Issues still there:
1 -While running/ Flying, the avatar still passes through almost all objects(in my test cases, the 4 walled closed wooden cabin, tables, box walls of room, mirror etc.)
2-While flying avatar even passes through the dunes.
3-When avatar falls through the a place, it passes through the ground. I think its an issue related to speed, whenever avatar is moving with speed it passes through the objects.
4-Feet still not visible
5-Invisible collectibles? is that a thing?
6-Classroom Scene isn't working fine, the avatar appears below the scene(Screenshot attached)
7-Rides vehicle, it moves to but goes through all the objects even the hills, dunes. The movement of vehicle too is not working for up/down it moves in one direction based on cursor.(screenshot attached)
Vehicle passes thru objects

Screenshot 2021-11-18 at 12 41 18 AM

Loryhoof:
1. Being able to walk through the avatar is intended, otherwise you'd be able to move it around. Will need to change later. I have tested `activate` on https://staging.webaverse.com/ it works if you go in first person and look at the avatar's feet/lower body. This has to do with the height difference of each avatar, will be fixed.

2.Seems to be a physx grounded check problem, will look into that.

3. I believe lantern and firefly also break in `master`. If it doesn't add too much time please crosscheck `master` and `staging` cause some issues may not be related to this PR. (Though they'll need to be fixed either way)

4.Feet should be fixed next PR

5. I'm not sure what that is, please explain.

6. Just checked, happens in `master` too

7. Vehicles should probably be disabled for now, they don't have any physics based movement yet. I enabled them temporarily so it doesn't break. Physics based vehicles will be the next thing after CCD (Same as in `master`)

8. Currently the capsule is able to move through all objects' backfaces. As if the wall's collider was only front-sided and not double-sided. I will look into this and check if it's intended or not.

Thank you for the review. It does help alot.

amnaarshed4224:
1-I just checked the staging too but it doesn't shows again the E even while looking on lower body/feet can you look into it?

https://user-images.githubusercontent.com/35926530/142400972-ecc4a279-d4a9-4c12-9ba3-8e45eb79d324.mp4

5- I mean there are several collectables that are invisible, while they show E when reached a certain place. Is that a thing?

amnaarshed4224:
Device:
Mac m1

1- Walk works fine with normal fps(30-35). It starts jittering sometimes and the fps turn low(20-27) simultaneously. there is no jitter in some scenes where high fps(in Shadows scene(sand dunes), fps on same device are 40-55), that I guess is an issue related to effects in street scene. on low fps, jitter is caused.

https://user-images.githubusercontent.com/35926530/142597252-cd274e03-c19b-43d4-a49a-f8612245768b.mp4


https://user-images.githubusercontent.com/35926530/142598334-995c287f-113d-47a4-8409-16b09b1aeea6.mp4


https://user-images.githubusercontent.com/35926530/142598366-4a9ac609-460c-4d23-a5ac-3385dcabf3b8.mp4


2-pressing (Cmd+shift+f4) and (Cmd+shift+f5) to record  somehow turns walk to run(doesn't requires to hold shift key). Gets fixed when shift is pressed again.

3-Changing the avatar in shadows, darkness, grid scene makes avatar walk and gestures unstable(works fine with flying only, also works fine in app.webaverse.com). log and video attached

https://user-images.githubusercontent.com/35926530/142597957-3a0f757d-5fd6-40df-a41c-56a8e954c69c.mp4

4- Changing avatar in gunroom makes it invisible. Also another avatar appears at some distance as collectable.

https://user-images.githubusercontent.com/35926530/142597985-3ffbb053-b3d4-42f0-8fa0-d4365a01fd52.mp4


5-The collectable fox, in Tree house goes through objects and sometimes passes through the ground too.(Happens in the App.webaverse.co too. not related to this pr i guess)


https://user-images.githubusercontent.com/35926530/142598027-46bdcd3d-ddf8-468f-a18b-f818ae5e1eb2.mp4


6-in classroom scene avatar still appears under the surface. This issue is there in the production too.


https://user-images.githubusercontent.com/35926530/142598069-0ea24cd8-af6d-4fe7-b826-4ef705ba10f3.mp4

amnaarshed4224:
The Issue of jittering is solved.
Related to this PR, only speed IMO is a bit slow, should be optimized, as it appears to be in slow motion while walking/running.
Else is fine.

app-manager.js
```diff
@@ -543,6 +543,12 @@ class AppManager extends EventTarget {
                 physicsObject.position.copy(app.position);
                 physicsObject.quaternion.copy(app.quaternion);
                 physicsObject.scale.copy(app.scale);
+
+                if(app.appType === "vrm") {
+                  //const h = app.avatar ? app.avatar.height : 0;
+                  physicsObject.position.add(new THREE.Vector3(0, 1, 0));
```
avaer, line 549: App manager should not know about VRM.

character-physics.js
```diff
@@ -38,7 +38,7 @@ class CharacterPhysics {
     this.velocity.add(keysDirection);
   }
   applyGravity(timeDiffS) {
-    if (this.player.avatar) {
+    if (this.player) {
```
avaer, line 41: Isn't there always a player that this physics is tracking?

character-physics.js
```diff
@@ -55,7 +55,7 @@ class CharacterPhysics {
         .add(localVector2.set(0, -avatarHeight * 0.5, 0));
       const radius = 0.3 / 1.6 * avatarHeight;
       const halfHeight = Math.max(avatarHeight * 0.5 - radius, 0);
-      return physx.physxWorker.collidePhysics(physx.physics, radius, halfHeight, localVector, q, 4);
+      return physx.physxWorker.collidePhysics(physx.physics, radius, halfHeight, localVector, q, 10);
```
avaer, line 58: `collideCapsule` should be deleted and not used at all.

character-physics.js
```diff
@@ -80,6 +80,8 @@ class CharacterPhysics {
         this.rb.matrixWorld.decompose(localVector, localQuaternion, localVector2);
         localQuaternion.copy(this.player.quaternion);
 
+        const capsuleOffset = new THREE.Vector3(0, this.player.avatar.height/2, 0);
```
avaer, line 83: Should not allocate here, this is the main loop.

character-physics.js
```diff
@@ -165,8 +171,10 @@ class CharacterPhysics {
 
         localVector.add(this.sitOffset);
         localVector.y += 1;
-        localQuaternion.premultiply(localQuaternion2.setFromAxisAngle(localVector3.set(0, 1, 0), Math.PI));*/
+        localQuaternion.premultiply(localQuaternion2.setFromAxisAngle(localVector3.set(0, 1, 0), Math.PI));
       }
+      const feetOffset = new THREE.Vector3(0, 0.05, 0); // Or feet will be in ground, only cosmetical, works for all avatars
```
avaer, line 176: Why does this work for all avatars? Could you explain the math?

character-physics.js
```diff
@@ -203,7 +216,7 @@ class CharacterPhysics {
   updateVelocity() {
     if(this.player.avatar) {
       if(this.rb) {
-        physicsManager.setVelocity(this.rb, this.velocity /*this.velocity.clone() */);
+        physicsManager.setVelocity(this.rb, this.velocity.clone());
```
avaer, line 219: Should not allocate in main loop.

scenes/shadows.scn
```diff
@@ -15,39 +15,7 @@
...
-          "middleGrey": 3
-        },
-        "bloom": {
-          "strength": 0.2,
-          "radius": 0.5,
-          "threshold": 0.8
-        }
-      }
-    },    
+    },
```
avaer, line 18: Fine for testing but should not make it into final.

avaer (commented):
Some structural problems.

The biggest one is that `collideCapsule` still exists and is being used for avatar calculations. IMO the only reason to use that method now would be to do "grab mesh" intersection tests.

Issue #1814, opened 2021-11-15
Camera pressure 3x3 optimization


avaer:
The camera pressure is slightly annoying in small branch/pole/hill cases:

https://user-images.githubusercontent.com/6926057/141834878-7738b4f5-128f-4d82-b499-8b1df5b543d2.mp4

To clean this up, we could test 8 more squares around the camera and only zoom in if a sufficient number of those (e.g. 2) are also obscured.

Additionally, if we are doing multi-raycasts, this might be a good time to implement arrayed raycast support in [app-wasm](https://github.com/webaverse/app-wasm) so we only have to call into WASM once after setting up the sources array.

avaer:
Fixed.

PR #1829, opened 2021-11-18
Camera pressure - small object ignored

skaljac:
This PR is about camera pressure.
Now we are ignoring small objects that don't interfere with view.

Example is shown here:

https://user-images.githubusercontent.com/4487857/142434816-607f2eb8-25cc-4e52-8b28-85ec418b0237.mp4


skaljac:
Tree branches was intercepting rays from player to camera, and that was causing camera to jump in front of tree branch.
Now we are ignoring small objects that intercept rays so camera is more stable in its position.

skaljac:
This is the issue in question https://github.com/webaverse/app/issues/1814

yaserOSource (approved):
Looks good. I've tested it and it works.  Feels quite stable.

Areas with a lot of objects, like multiple trees and then rocks in various positions scattered around the player, can cause the camera to go forward backward, in quick succession, but other than that its good.

amnaarshed4224:
1-The Camera View has improved a lot, but the transition of camera from far to near when an object gets in the way is not smooth and appears very fast and sometimes disruptive.(I will post the video tomorrow due to an issue.)

Else is fine, the ignoring of small objects has made the view much more detailed.

abeersaqib:
Looks fine to me. @amnaarshed4224 Please post the video for reference.

camera-manager.js
```diff
@@ -121,29 +121,96 @@ const cameraManager = {
...
+    // Two rays from players position with small offset left and right
+    function checkSmallObjectIntersections(maxDistance)
+    {
+      let intersections = 0;
+
+      const localPlayer = metaversefile.useLocalPlayer();
+
+      let checkPoints = [];
+      
+      const sideways = new THREE.Vector3();
```
avaer, line 147: This is in the render loop, so it should not allocate.

Can be `localVector`.

camera-manager.js
```diff
@@ -121,29 +121,96 @@ const cameraManager = {
...
+      
+      // Left
+      checkPoints[0] = localPlayer.position.clone();
```
avaer, line 150: `clone()` is also an allocation.

camera-manager.js
```diff
@@ -121,29 +121,96 @@ const cameraManager = {
       return value1 + (value2 - value1) * amount;
     }
 
+    function checkRayIntersection(origin,dirQuaternion,maxDistance)
+    {
+      let res = false;
+
+      let collision = physicsManager.raycast(origin, dirQuaternion);
```
avaer, line 128: We could actually make the API support taking multiple raycasts at a time.

avaer: That should help with performance. Since this is run every frame, it might be worth it.

camera-manager.js
```diff
@@ -121,29 +121,96 @@ const cameraManager = {
...
     let hasIntersection = false;
 
     const localPlayer = metaversefile.useLocalPlayer();
 
+    let z = ( camera.near + camera.far ) / ( camera.near - camera.far );
+
     let cornerPoints = [];
     // Top left
-    cornerPoints[0] = new THREE.Vector3( -1, 1, ( camera.near + camera.far ) / ( camera.near - camera.far ) );
+    cornerPoints[0] = new THREE.Vector3( -1, 1, z );
```
avaer, line 192: More allocations.

camera-manager.js
```diff
@@ -121,29 +121,96 @@ const cameraManager = {
...
+    }
+
+    // Two rays from players position with small offset left and right
+    function checkSmallObjectIntersections(maxDistance)
+    {
+      let intersections = 0;
+
+      const localPlayer = metaversefile.useLocalPlayer();
+
+      let checkPoints = [];
```
avaer, line 145: This is also an allocation which I don't think is necessary, especially if we collapse the raycast into a single call.

avaer (changes requested):
The math looks fine, but we should optimize the memory allocation on this.

camera-manager.js
skaljac: Ok, I am working on API for multiple raycasts at a time.
And that will change a lot of this code.

skaljac:
Last commit is introducing raycast with array parameters as requested.

camera-manager.js
skaljac: I have removed all allocations from render loop in last commit.

camera-manager.js
skaljac: I have removed all allocations from render loop in last commit.

camera-manager.js
```diff
@@ -121,53 +126,102 @@ const cameraManager = {
...
-      let collision = physicsManager.raycast(rayStartPos, rayQuaternion);
-      if (collision) {
-        if (collision.distance <= -1 * newVal) {
-          if (newVal < (-1 * (collision.distance-0.15))) {
-            newVal = (-1 * (collision.distance-0.15));
-            hasIntersection = true;
-          }
+    // Check collision from player to camera corners
+    for(let i=0;i<4;i++) {
+      if ((collisionArray.hit[i] == 1) && (collisionArray.distance[i] <= -1 * newVal)) {
```
avaer, line 200: Probably should be `===`

camera-manager.js
```diff
@@ -121,53 +126,102 @@ const cameraManager = {
...
+          hasIntersection = true;
+          //console.log(i + " " + collisionArray.distance[i]+ " " + collisionArray.hit[i]);
         }
       }
     }
+
+    // Check collision from player pos and small offset to left and righ - to camera center
+    let offsetCollisionCount = 0;
+    for(let i=4;i<6;i++) {
+      if ((collisionArray.hit[i] == 1) && (collisionArray.distance[i] <= (-1 * cameraOffsetTargetZ))) {
```
avaer, line 212: `===`

camera-manager.js
```diff
@@ -121,53 +126,102 @@ const cameraManager = {
...
+        offsetCollisionCount++;
+      }
+    }
+
+    // Discard collision with small objects
+    if (hasIntersection && (offsetCollisionCount == 0))
```
avaer, line 218: `===

physx.js
```diff
@@ -612,6 +612,74 @@ const physxWorker = (() => {
       } : null;
     }
   };
+  w.raycastPhysicsArray = (physics, p, q, n) => {
+    if (physics && (n > 0 && n < 1000)) {
```
avaer, line 616: If we are going to guard then we should at least throw an error if it's violated.

avaer (commented):
Small changes requested.

physx.js
```diff
@@ -612,6 +612,74 @@ const physxWorker = (() => {
...
+
+      // Input parameters
+      for(let i=0;i<n;i++) {
+        // Position
+        p[i].toArray(scratchStack.f32, (i * 3));
+
+        // Quaternion
+        localVector.set(0, 0, -1)
+        .applyQuaternion(q[i])
+        .toArray(scratchStack.f32, (n * 3) + (i * 3));
```
avaer, line 627: I had to stop and think about why we are doing these offsets this way.

IMO it would be more clear if before this you declare an alias for these slots, like 

```
directions = new Float32Array(scratchStack.f32.buffer, scratchStack.f32.byteOffset + (n * 3), numDirections * 3);
```

Then this line would be

```
.toArray(directions, i * 3);
```

That way each array does not need to be aware of the previous arrays, which feels messy.

physx.js
skaljac: I have added throw new Error('raycastPhysicsArray error'); to solve this issue

physx.js
skaljac: Last commit is taking care of this issue

camera-manager.js
```diff
@@ -121,53 +126,100 @@ const cameraManager = {
...
-      rayDirection.subVectors(localPlayer.position, cornerPoints[i] ).normalize();
+    // Camera - Bottom right
+    initCameraRayParams(3,rayStartPos.set(1, -1, ( camera.near + camera.far ) / ( camera.near - camera.far )).unproject( camera ));
 
-      rayMatrix.lookAt(rayDirection,rayVectorZero,rayVectorUp);
-      rayQuaternion.setFromRotationMatrix(rayMatrix);
+    // Player postition - offset to left
+    rayStartPos.copy(localPlayer.position);
+    rayOffsetPoint.set(-1, 0, 0);
+    rayOffsetPoint.applyQuaternion(camera.quaternion);
```
avaer, line 183: I would have chained these variables for less typing.

Almost everything in three.js returns itself,

```
rayOffsetPoint.set(-1, 0, 0)
  .applyQuaternion(camera.quaternion);
```

avaer (approved):
Thank you, this seems like quite good code.

amnaarshed4224:
The camera view and transitions are working fine But there is one issue that needs to be addressed still

1-In congested places that are crowded with objects The camera view stirs. Kindly check it if it should be fixed i the current PR or a separate PR.

https://user-images.githubusercontent.com/35926530/143908722-b11e2716-ad2f-49c9-b13b-ada0c0d71f97.mp4

skaljac:
Last commit is about elimination of jitter when there is no camera movement.

https://user-images.githubusercontent.com/4487857/143954843-0cf74560-ecf2-42d9-be3d-cb491c10f188.mp4

amnaarshed4224:
The jitter is gone while there's no camera movement.
While passing through the back of objects it still stirs but I guess that is something that can be ignored for now

https://user-images.githubusercontent.com/35926530/144077711-24fcedb8-57d8-417d-b2ec-243d3259e93d.mp4

.

avaer (approved):

PR #1832, opened 2021-11-19
Blockchain integrated

mujabmuneeb:


src/Header.jsx
```diff
@@ -226,8 +227,26 @@ const User = ({address, setAddress, open, setOpen, toggleOpen}) => {
           setLoggingIn(true);
           try {
             const {address, profile} = await ceramicApi.login();
-            // console.log('login', {address, profile});
             setAddress(address);
+            if (address) {
+              (async () => {
+                var tokenUri = [];
+                const res = await fetch(`${apiBackend}/tokenids?address=${address}`);
```
avaer, line 234: This could be one request to get the uris, rather than O(n + 1) requests. Is there a reason for doing it this way?

avaer (commented):
Looks like it's headed in the right direction.

src/Header.jsx
mujabmuneeb: converted into single request. ✅

integration/web3/app.mjs
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    await contract.mint(await signer.getAddress(), tokenURI, {gasLimit: 200000});
+  } catch (err) {
+    console.log(err);
+  }
+}
+
+async function drop(tokenID) {
+  const claimableVoucher = new ClaimableVoucher({contract: contract, signer: signer});
+  const timestamp = Math.round(new Date().getTime() / 1000) + 1000;
```
avaer, line 30: This doesn't matter, but:

```new Date().getTime() = Date.now()```

avaer (approved):

avaer:
There is no description on this issue. Can you explain how it is treating the MetaMask and non-MetaMask cases?

avaer:
I reverted this PR because it seems other PR prerequisites are not met: https://github.com/webaverse/app/commit/9180a7d52f4eeb052c866cd42645fb170bbc4181

Can you please add a description about dependencies and merge order?

PR #1835, opened 2021-11-20
CCD

Loryhoof:

How does it work?

* We pass our player velocity into physX, which runs a simulation (we can set parameters like friction/gravity/bounciness etc.)
* We then call physX again to get the updated character capsule transform, then apply it to our avatar.

### Testing
1. Crouch walk down/up hills and steps (Heavy jitter? Fall through ground? Get stuck?)
1. Test switching scenes multiple times, while equipping weapons and changing avatars.

### Testing Nov 19
1. Jump (Spacebar) actually didn't work before on flat surfaces, and I'm not sure how nobody noticed, but its fixed now please test on hills and flat areas.
1. Feet not stuck in ground anymore
1. Now that switching avatars works, please test if you can consistently change it and you spawn at the correct position (Where you held the `activate` button) before it would sometimes teleport you back to origin spawn (0,0,0)

### NEW Testing Nov 20
1.Test if wall jitter still happens (as shown in avaer's video in `engine-2`

┆Issue is synchronized with this [Jira Task](https://webaverse.atlassian.net/browse/EN-983) by [Unito](https://www.unito.io)


amnaarshed4224:
1-The wall jitter still happens at a certain angle.

https://user-images.githubusercontent.com/35926530/142736480-406cec59-08b9-496f-aef2-14591353b127.mp4


https://user-images.githubusercontent.com/35926530/142736482-102601c7-eae2-4fb9-918f-0ca44706855b.mp4

2- building wall jitter while running increased


https://user-images.githubusercontent.com/35926530/142736496-dd70f1a4-b35b-4a11-9d12-f23b96ac39fd.mp4

4-climbing on steps not smooth yet

https://user-images.githubusercontent.com/35926530/142736500-69cde0a9-c12d-4f51-8d08-1ecfacd8c9d2.mp4


5- avatar looks embedded while flying needs solution too.

https://user-images.githubusercontent.com/35926530/142736519-c5894d87-3617-48d1-a9af-f7b8cdca534b.mov

physics-manager.js
```diff
@@ -24,6 +24,8 @@ const localQuaternion = new THREE.Quaternion();
 const localQuaternion2 = new THREE.Quaternion();
 const localMatrix = new THREE.Matrix4();
 
+const localMaterial = new THREE.Vector3();
```
avaer, line 27: I don’t think a material should be a vector3… it should just be the properties.

webaverse.js
```diff
@@ -297,7 +297,7 @@ export default class Webaverse extends EventTarget {
     const animate = (timestamp, frame) => { 
       timestamp = timestamp ?? performance.now();
       const timeDiff = timestamp - lastTimestamp;
-      const timeDiffCapped = Math.min(Math.max(timeDiff, 0), 30); // What is 100?
+      const timeDiffCapped = timeDiff;//Math.min(Math.max(timeDiff, 0), 30); // What is 100?
```
avaer, line 300: This is so that a very long frame (lag) does not junp several seconds into the future when the lag resolves.

PR #1839, opened 2021-11-22
Zjs networking for multiplayer

avaer:
Similar to y.js but contains game engine features we need: https://github.com/webaverse/zjs

┆Issue is synchronized with this [Jira Task](https://webaverse.atlassian.net/browse/EN-980) by [Unito](https://www.unito.io)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "packages/wsrtc"]
 	path = packages/wsrtc
 	url = https://github.com/webaverse/wsrtc.git
+[submodule "packages/zjs"]
+	path = packages/zjs
+	url = https://github.com/webaverse/zjs.git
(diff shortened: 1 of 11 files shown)


avaer:
Requires https://github.com/webaverse/wsrtc/pull/8.

avaer:
As far as I can tell, this library replacement slots in as a direct replacement for `yjs`; apps are correctly tracked with the new library since it is using the same API.

Will do more testing to make sure there are no subtle regressions.

avaer:
The only regression that I found was that over time we run out of history buffer.

That can be solved by using a ring buffer strategy; there is plenty of room for the history, but we eventually overrun the 1 MB that is allocated. We can loop around.

avaer:
We should also make sure the `zjs` module tests are passing.

avaer:
Added the ring buffer: https://github.com/webaverse/zjs/blob/51653241c7df2f21f25581ce3c927187aeb44cf0/z.mjs#L476

avaer:
I ran this code overnight and it did not overflow or crsah.

avaer:
Tests are fixed.

avaer:
https://github.com/webaverse/wsrtc/pull/8 was merged.

PR #1844, opened 2021-11-22
CCD

Loryhoof:

How does it work?

* We pass our player velocity into physX, which runs a simulation (we can set parameters like friction/gravity/bounciness etc.)
* We then call physX again to get the updated character capsule transform, then apply it to our avatar.

### Testing
1. Crouch walk down/up hills and steps (Heavy jitter? Fall through ground? Get stuck?)
1. Test switching scenes multiple times, while equipping weapons and changing avatars.

### Testing Nov 19
1. Jump (Spacebar) actually didn't work before on flat surfaces, and I'm not sure how nobody noticed, but its fixed now please test on hills and flat areas.
1. Feet not stuck in ground anymore
1. Now that switching avatars works, please test if you can consistently change it and you spawn at the correct position (Where you held the `activate` button) before it would sometimes teleport you back to origin spawn (0,0,0)

### Testing Nov 20
1.Test if wall jitter still happens (as shown in avaer's video in `engine-2`

### NEW Testing Nov 22
1.Test if wall jitter still happens (as shown in avaer's video in `engine-2`

* You could try the new prototype.scn for QA

┆Issue is synchronized with this [Jira Task](https://webaverse.atlassian.net/browse/EN-1768) by [Unito](https://www.unito.io)


amnaarshed4224:
1-The jitter has got a lot worse while walking/Running/Flying.


https://user-images.githubusercontent.com/35926530/142942399-08e7747b-3865-4979-8f10-13b7f2612923.mp4


https://user-images.githubusercontent.com/35926530/142942431-d01ecc26-8ca2-4a6d-b765-c097c84b2911.mp4


https://user-images.githubusercontent.com/35926530/142942413-95fdff6f-ac2f-4ada-9674-c437a5885392.mp4


https://user-images.githubusercontent.com/35926530/142942423-e6649bd6-829e-492d-b881-cb5873a23150.mp4


2-embedding of avatar in the new scene

https://user-images.githubusercontent.com/35926530/142942449-300d9a9c-24d8-4f1b-82ff-294424164309.mp4

3-sometimes after jumping the game kind of crashes as avatar starts jittering really bad. 

4- Changing avatar has issues

https://user-images.githubusercontent.com/35926530/142942785-14bda669-bf49-498f-a1f5-233e9e180371.mp4

5-Standing on slopes also has jitter

https://user-images.githubusercontent.com/35926530/142942893-e37b473c-4bc9-42d3-adc6-f95ab95719ca.mp4

Issue #1841, opened 2021-11-22
App: sit positions broken


avaer:
![image](https://user-images.githubusercontent.com/6926057/142857619-65ba8bb8-ea31-4618-9c76-fc6df5455032.png)

This regressed recently, probably due to massive avatar skeleton refactors.

┆Issue is synchronized with this [Jira Task](https://webaverse.atlassian.net/browse/EN-1765) by [Unito](https://www.unito.io)

avaer:
The fixes were merged, this needs verification before closing the issue.

Vianvolaeus:
Appears to be working fine. 

![image](https://user-images.githubusercontent.com/80061436/143258360-69952d29-17af-4b3f-8e05-b71d7528cafe.png)

avaer:
Thanks, I saw it working fine too.

PR #1845, opened 2021-11-22
Sit position fix

Loryhoof:
sitwip

┆Issue is synchronized with this [Jira Task](https://webaverse.atlassian.net/browse/EN-1769) by [Unito](https://www.unito.io)

### DIFF
--- a/character-physics.js
+++ b/character-physics.js
@@ -153,7 +153,7 @@ class CharacterPhysics {
           .decompose(localVector, localQuaternion, localVector2);
 
         localVector.add(this.sitOffset);
-        localVector.y += 1;
+        localVector.y += 1.4;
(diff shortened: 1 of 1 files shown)


Loryhoof:
https://github.com/webaverse/app/issues/1841

avaer (changes requested):
Both 1 and 1.4 are incorrect values.

I should work for all avatars and all sittables, with no constants.

Loryhoof:
changing it from 1.4 to `this.player.avatar.height` and changing the hovercraft's metaversefile's sitOffset to 0.5 works for all avatars. The dragon mount is using the sitbone so doesn't need sitOffset change.

https://github.com/webaverse/hovercraft/pull/3

avaer (approved):

PR #1856, opened 2021-11-24
Character controller 

abeersaqib:


app-manager.js
```diff
@@ -543,9 +543,16 @@ class AppManager extends EventTarget {
                 physicsObject.position.copy(app.position);
                 physicsObject.quaternion.copy(app.quaternion);
                 physicsObject.scale.copy(app.scale);
+
+                if(app.appType === "vrm") {
```
avaer, line 547: App manager should not be aware of app type. This code belongs in the loader.

character-physics.js
```diff
@@ -25,9 +25,12 @@ const localMatrix = new THREE.Matrix4();
 const zeroVector = new THREE.Vector3();
 const upVector = new THREE.Vector3(0, 1, 0);
 
+
 class CharacterPhysics {
   constructor(player) {
     this.player = player;
+    this.rb = null;
```
avaer, line 32: What is `rb`?

character-physics.js
```diff
@@ -63,14 +66,26 @@ class CharacterPhysics {
...
+
 
-        this.player.updateMatrixWorld();
-        this.player.matrixWorld.decompose(localVector, localQuaternion, localVector2);
+        this.rb.updateMatrixWorld();
+        this.rb.matrixWorld.decompose(localVector, localQuaternion, localVector2);
+        localQuaternion.copy(this.player.quaternion);
 
-        const collision = this.collideCapsule(localVector, localQuaternion2.set(0, 0, 0, 1));
+        const capsuleOffset = new THREE.Vector3(0, this.player.avatar.height/2, 0);
```
avaer, line 86: We should not allocate here, since I think it is part of the render loop.

character-physics.js
```diff
@@ -63,14 +66,26 @@ class CharacterPhysics {
...
+        localVector.add(capsuleOffset);
+        const collision = this.collideCapsule(localVector, localQuaternion2.set(0, 0, 0, 1)); // TODO: Replace with physicsManager.isGrounded restitution check
```
avaer, line 88: We should not collide physics here. This should be removed.

character-physics.js
```diff
@@ -156,6 +176,8 @@ class CharacterPhysics {
         localVector.y += 1;
         localQuaternion.premultiply(localQuaternion2.setFromAxisAngle(localVector3.set(0, 1, 0), Math.PI));
       }
+      const feetOffset = new THREE.Vector3(0, 0.05, 0); // Or feet will be in ground, only cosmetical, works for all avatars
```
avaer, line 179: More allocation.

character-physics.js
```diff
@@ -189,6 +216,36 @@ class CharacterPhysics {
...
+  updateVelocity() {
+    if(this.player.avatar) {
+      if(this.rb) {
+        physicsManager.setVelocity(this.rb, this.velocity.clone());
+      }
+    }
+  }
+  updateTransform() {
+    if(this.rb) {
+      const physicsObjects2 = [];
```
avaer, line 228: Allocation in the render loop?

character-physics.js
```diff
@@ -189,6 +216,36 @@ class CharacterPhysics {
       this.velocity.z *= factor;
     }
   }
+  updateVelocity() {
+    if(this.player.avatar) {
+      if(this.rb) {
+        physicsManager.setVelocity(this.rb, this.velocity.clone());
```
avaer, line 222: Allocation.

character-physics.js
avaer: I also don't think `isGrounded` is something you can ask after the simulation.

Grounded state must come from the simulation. After that, we don't know if we are truly grounded since the state has already been restituted and that knowledge has been lost.

avaer (changes requested):
More code issues to fix.

I suspect they will affect functionality (for the better) once fixed.

amnaarshed4224:
The CCD works fine within its domain. The issues that appear in it are purely fps based depending on each scene. I have attached few examples how fps affect the performance of avatar.

1-On 60 or more fps, everything works perfectly fine without any kind of issues/jitter. Just as fps drop, The issues such as jitter while walking/running appear.

-On 60 fps(no jitter issues.)

https://user-images.githubusercontent.com/35926530/143900719-2b82bbec-8bb6-464c-9c49-57947a055e03.mp4

-FPS drop in the same scene causing jitter

https://user-images.githubusercontent.com/35926530/143900827-c290043b-c9d2-48c2-bad3-c1c040f6cb7d.mp4
 
-No issue in scenes like prototype, grid where fps remain stable at above 60.

https://user-images.githubusercontent.com/35926530/143900981-deea7495-c0ad-43d7-8c56-fa0e0ba9a469.mp4

-Jitter drop in high requirement scenes causes issues.

https://user-images.githubusercontent.com/35926530/143901896-1fcd3a92-490b-439d-8874-d79e889cc9aa.mp4

IMO the issues related to this PR are resolved, The FPS related issues should be fixed separately. The CCD can be pushed to production.

ahadshams:
FPS has improved for me..
Wall jitter doesnt occur at high fps..

amnaarshed4224:
@avaer The CCD related issue are fixed. What is left is FPS related issues. Kindly review it for final verdict.

Loryhoof:
`on collision` function is still in use for the grounded check, haven't really gotten it to work without it. 
 It still works, but if it's necessary to change it I'll need to get help from the team.

PR #1857, opened 2021-11-24
Test Cases Updated to match new Module Structure (EN-1778)

mujabmuneeb:
Issue: The index file has been moved to module type and is thus renamed from index.js to index.mjs

Resolved: 
1. File call for index.js is changed to index.mjs (package.json & test/loading.js)
2. testCase.js is changed to testCase.mjs (test/loading.js) 
3. Before, we were only catching errors after app starts (only scenes test cases). Now I have fixed that and its returning app boot errors too (which is actually needed right now)
4. Converted discord messenger into a new function for re-usability.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "",
   "main": "index.mjs",
   "scripts": {
-    "start": "forever start index.js",
-    "dev": "node index.js",
+    "start": "forever start index.mjs",
+    "dev": "node index.mjs",
     "build": "vite build",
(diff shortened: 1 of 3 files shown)


avaer:
This is a really poor issue description.

image

ahadshams:
extremely poor !?!?
i could have never understood it by reading it

mujabmuneeb:
Sorry for the inconvenience, both title and description are updated now.

avaer (approved):

PR #1859, opened 2021-11-24
Add COOP headers to index.mjs serving

avaer:
This is needed for supporting `SharedArrayBuffer`, high-resolution `performance.now()`, and also for a strict renderer process separation for child iframes, which is important for the [kalidokit implementation](https://github.com/webaverse/app/pull/1788) to be able to perform asynchronous GPU (model inference) work.

The end result is that if we do it this way, we will have faster mediapipe (pose model) inference, which does not interfere with the main game engine loop.

This is due to how the browser process architecture works (https://developers.google.com/web/updates/2020/07/referrer-policy-new-chrome-default).

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -35,6 +35,9 @@ function makeId(length) {
 (async () => {
   const app = express();
   app.use('*', async (req, res, next) => {
+    res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
+    res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp');
+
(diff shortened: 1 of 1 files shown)


avaer:
In my testing I found that despite this PR we might _still_ need to follow the strict site isolation origin policy of TLD +1. I was able to hitch the render process of the parent from the iframe regardless, as shown in the profile..

avaer:
The fix therefore probably involves using another webaverse domain, like `webaverse.link`.

Issue #1633, opened 2021-10-23
App: SFX pack


avaer:
We should get an SFX pack and create a sound system to trigger them. Sound is easy but pretty important for the final product.

skaljac:
This sfx pack for anime games should be interesting for Webaverse:
https://assetstore.unity.com/packages/audio/sound-fx/anime-epic-combat-sound-effect-pack-206061


Maybe this is also interesting
https://assetstore.unity.com/packages/audio/sound-fx/vintage-anime-sfx-158863


Here we have some basic sfx sounds
https://assetstore.unity.com/packages/audio/sound-fx/eastern-sfx-74043


All of the assets can be seen here:
https://assetstore.unity.com/top-assets/top-paid?category=audio&q=anime&orderBy=1&page=1

PR #1864, opened 2021-11-24
Sound manager implementation,game sfx   WIP

skaljac:
This is related to issue:
https://github.com/webaverse/app/issues/1633

sound-manager.js implements sound subsystem for this game.

We now have sound for footsteps (walking and running), jump start sound, jump end sound.

Steps for testing:
walk around - footstep sound should play
run - footsteps should be faster
jump - there should be a sound for jump and jump end


amnaarshed4224:
Yes, The sounds are working fine. Just as it is supposed to be.

avaer:
Are we making sure to synchronize the footsteps perfectly with the animation?

Doing that should actually not be hard, since the animation loop is very well defined, and the time of the feet hitting the ground in the animation is consistent, purely based on the current game time.

avaer:
Loading all animations in avatars.js: https://github.com/webaverse/app/blob/a5346a446e519283317f746a2e7856c59c9c3fae/avatars/avatars.js#L223
Animation timing is based purely on the render loop current world time, so we can know perfectly when there is a foostep: https://github.com/webaverse/app/blob/a5346a446e519283317f746a2e7856c59c9c3fae/avatars/avatars.js#L1628

avaer:
Implementation notes:

- this looks like it's loading one copy of each sound, which means probably you can't play the same sound twice (the second hit will cut off the first one?)
- MP3 or OGG is probably better than WAV, especially as the amount of sounds increases.

avaer (commented):
Overall looks good.

skaljac:
I have change file format to .ogg
Also, as suggested we now have synchronization with steps for footstep sounds

avaer:
Maybe mp3 is better if we check the compatibility tables: https://caniuse.com/?search=ogg

skaljac:
I have changed file format to .mp3 it works fine in chrome, chrominum

amnaarshed4224:
This PR works Perfectly fine with sounds of footsteps while running/walking. The footsteps are too in synchronization with animation. So, its ready for merge if there are no code changes left.

amnaarshed4224:
This PR works Perfectly fine with sounds of footsteps while running/walking. The footsteps are too in synchronization with animation. So, its ready for merge if there are no code changes left.
1-One thing that should be in your notice is hovercraft and dragon have same footsteps sound right now. Might needs to be changed.

skaljac:
Ok, I need to change that in this PR

skaljac:
This problem with fly,jump,sit state of player and step sound is now fixed

amnaarshed4224:
Yes, Its working perfectly fine now.

avaer (approved):

PR #1866, opened 2021-11-25
Performance: no autoupdate [WIP]

avaer:
This PR improves performance massively by eliding matrix update operations. However, app fixes are needed before we can ship this performance boost.

## Context
THREE.js by default updates all matrices (local and world, composed from position/quaternion/scale) in a scene each frame. This is a very heavy CPU operation. It is the most expensive thing during rendering, and depends O(n) on the number of objects in the scene.

However, most of this math is pointless. We know when things are being updated in the code, so we can just call updateMatrix() at that time, rather than for all objects each frame (in _addition_ to the update we are already doing).

It is easy to turn off this feature in THREE; it is the `scene.autoUpdate` property on all scenes. I have added that in this PR, and it massively improves the performance and the profile.

However, the downside is that if you update P/Q/S but do not call the matrix update, the object will _not_ have the correct transform during rendering. The fix is to make sure that all cases of updates are followed -- once the transformation math is fully complete — by `Object3D.updateMatrixWorld()`. Then we would have a significant performance improvement while everything would still render correctly.

## Progress
This PR is not done; there are plenty of corner cases in our apps and internal objects that need to be fixed to update their matrices correctly with this PR, by adding `.updateMatrixWorld()` on the errant objects whenever they update. That should be done regardless of this PR.

Once everything renders correctly with this PR and it passes QA, we can merge it.

### DIFF
--- a/preview.js
+++ b/preview.js
@@ -287,6 +287,7 @@ window.onload = async () => {
         context,
       });
       const scene = new THREE.Scene();
+      scene.autoUpdate = false;
       const camera = new THREE.PerspectiveCamera(60, window.innerWidth / window.innerHeight, 0.1, 1000);
       camera.position.set(0, o.rig.height, -2);
       const target = new THREE.Vector3(0, o.rig.height / 2, 0);
@@ -337,6 +338,7 @@ window.onload = async () => {
         context,
       });
       const scene = new THREE.Scene();
+      scene.autoUpdate = false;
       const camera = new THREE.PerspectiveCamera(60, window.innerWidth / window.innerHeight, 0.1, 1000);
       camera.position.copy(center)
         .add(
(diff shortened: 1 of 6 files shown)


avaer:
There is an additional thing we can do _after_ we are done with fixing all of the transforms, which is setting `Object3D.DefaultMatrixAutoUpdate = false;`.

This would, in addition to not updating matrices eagerly in the scene, also make all objects not update their local matrix at all when the object's position/quaternion/scale changes. This would uncover additional corner cases that need an update slipped in, and the render will be broken until we fix them. However, that would add a second potential layer of performance boosts.

We should not attempt this part until the first part of the PR (`scene.autoUpdate`) is shipped. I commented out that line in the code for now: https://github.com/webaverse/app/pull/1866/files#diff-8a8d77cf15b2eaf50ec617c3603313fe363eaef34d8eb3c20ce12b607da55f04R11

avaer:
I am not currently working on this, but it should be a good basis for performance optimization.

Anyone who feels confident in updating matrices on our apps could take this to the finish line.

yaserOSource:
Testing this now

PR #1886, opened 2021-11-29
feat: add post-scene renderPass

Sirahi:
Added the postScene to render the gltj app or other objects after post-processing. The postScene render pass has to be enabled from renderSettings in scn file.
{
  "content": {
    "enablePostScene": true
  }
}
Gltj app or object can be added to postScene by setting renderPriority to postScene in components:
{
  "start_url": "https://webaverse.github.io/eyeblaster/",
  "component": [
    {
      "key": "renderPriority",
      "value": "postScene"
    }
  ]
}

### DIFF
--- a/post-processing.js
+++ b/post-processing.js
@@ -231,7 +231,7 @@ function setPasses(rendersettings) {
   composer.addPass(webaverseRenderPass);
   
   if (rendersettings) {
-    const {ssao, dof, hdr, bloom, usePostScene} = rendersettings;
+    const {ssao, dof, hdr, bloom, enablePostScene} = rendersettings;
     // console.log('got', ssao, dof, hdr, bloom);
     
     if (ssao) {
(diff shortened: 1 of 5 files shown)


post-processing.js
```diff
@@ -229,7 +231,7 @@ function setPasses(rendersettings) {
   composer.addPass(webaverseRenderPass);
   
   if (rendersettings) {
-    const {ssao, dof, hdr, bloom} = rendersettings;
+    const {ssao, dof, hdr, bloom, usePostScene} = rendersettings;
```
avaer, line 234: This can easily be confused with `usePostScene` above but it seems like it's not the same thing.

So can we choose a better name for the config param? `postScene`?

avaer (changes requested):
One suggestion.

avaer:
Let's go with `enablePostScene`, since that string is not used.

post-processing.js
Sirahi: Updated usePostScene to enablePostScene in post-processing. The usePostScene hook name in metaversefile-api will stay the same.

Sirahi: Updated the usage example comment as well.

avaer (approved):

PR #1901, opened 2021-12-01
Screenshot Preview with multiple iframes (Closing for future)

mujabmuneeb:
Added a new file (preview.js). There is a new function called generatePreview which receives url, ext, type, width, and height as params and returns blob object.


screenshot-iframe.js
```diff
@@ -0,0 +1,80 @@
...
+    if(hash) {
+        var tempUrl = `https://app.webaverse.com/${hash}.${ext}/preview.${type}`
+    }
+    else {
+        var tempUrl = `https://app.webaverse.com/screenshot.html?url=${reqUrl}&hash=${hash}&ext=${ext}&type=${type}&width=${width}&height=${height}`;
+    }
+
+    const u = url.parse(tempUrl, true);
+
+    const spec = (() => {
```
abeersaqib, line 39: what is the use of the function here ?

Can't we construct the url directly ?

screenshot-iframe.js
```diff
@@ -0,0 +1,80 @@
...
+    if(!iframeExists.length) {
+        var iframe = document.createElement('iframe');
+        iframe.width = '0px';
+        iframe.height = '0px';
+        iframe.name = 'screenshotIframe';
+        document.body.appendChild(iframe);
+    }
+
+    if(hash) {
+        var tempUrl = `https://app.webaverse.com/${hash}.${ext}/preview.${type}`
```
abeersaqib, line 31: why are you even constructing it this way ?

screenshot-iframe.js
```diff
@@ -0,0 +1,80 @@
...
+        }
+        spec = null;
+    })();
+
+    if (spec) {
+        const {url, hash, ext, type, height, width} = spec;
+        var ssUrl = `https://app.webaverse.com/screenshot.html?url=${url}&ext=${ext}&type=${type}&width=${width}&height=${height}`;
+    }
+
+    window.open(ssUrl, 'screenshotIframe')
```
abeersaqib, line 66: Have you tested it that it's not re-creating the iframe I don't think so it's tested properly?

abeersaqib (changes requested):

screenshot-iframe.js
```diff
@@ -0,0 +1,80 @@
+import {storageHost} from './constants';
+import url from 'url';
+
+function parseQuery(queryString) {
```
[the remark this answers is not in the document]
mujabmuneeb: Removed.

screenshot-iframe.js
mujabmuneeb: Yes we can, fixed it now.

screenshot-iframe.js
```diff
@@ -0,0 +1,80 @@
...
+        iframe.height = '0px';
+        iframe.name = 'screenshotIframe';
+        document.body.appendChild(iframe);
+    }
+
+    if(hash) {
+        var tempUrl = `https://app.webaverse.com/${hash}.${ext}/preview.${type}`
+    }
+    else {
+        var tempUrl = `https://app.webaverse.com/screenshot.html?url=${reqUrl}&hash=${hash}&ext=${ext}&type=${type}&width=${width}&height=${height}`;
```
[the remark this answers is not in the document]
mujabmuneeb: Updated it.

screenshot-iframe.js
mujabmuneeb: Yes, I have tested it by multiple function calls and it's not creating new iframes for every preview.

screenshot-iframe.js
```diff
@@ -0,0 +1,34 @@
+import {storageHost} from './constants';
+
+export const getScreenshotBlob = async (reqUrl, hash, ext, type, width, height) => {
```
abeersaqib, line 3: you are assuming here that hash is the second parameter this means if the user has to pass an argument for hash 
he will using like this?

getScreenshotBlob(undefined, 'some hash') ?
it looks not good to me 

Check for the first paramter whether it's a URL or hash and based on that construct the reqUrl

screenshot-iframe.js
```diff
@@ -0,0 +1,34 @@
+import {storageHost} from './constants';
```
abeersaqib, line 1: Change the filename to preview.js

screenshot-iframe.js
abeersaqib, line 3: Change name of function to generatePreview

abeersaqib (changes requested):

screenshot-iframe.js
mujabmuneeb: Done

screenshot-iframe.js
mujabmuneeb: Done

mujabmuneeb: Got it, fixed.

preview.js
```diff
@@ -1,476 +1,39 @@
...
+    if(!isValidURL(reqUrl)) {
+        reqUrl = `${storageHost}/ipfs/${reqUrl}`;
+    }
+    var ssUrl = `${window.origin}/screenshot.html?url=${reqUrl}&ext=${ext}&type=${type}&width=${width}&height=${height}`;
+    
+    window.open(ssUrl, 'screenshotIframe')
+
+    return new Promise(function (resolve) {
+
+    var f = function(event){
```
avaer, line 24: Messed up spacing.

mujabmuneeb: Fixed.

PR #1912, opened 2021-12-03
Rename metaversefile -> totum

yaserOSource:
ONE OF MANY

metaversefile -> totum

metaversefileApi -> totumApi

.metaversefile -> .totum

All Scenes have been set up to point to ALL forked repositories for testing purposes, this will be reverted to the webaverse repo on merge!

cd packages/totum and checkout hash 8ab8356b36fe8f8eb290c6b2db35cf86bd3bd5f6

QA Checklist:

Test each scene, try out all the items, can you see every single object in the scn list?
Go through FTUX, and interact with everything.

totum repo PR:

https://github.com/webaverse/totum/pull/41


src/Header.jsx
```diff
@@ -479,16 +479,92 @@ export default function Header({
...
-              open={open}
-              setOpen={setOpen} 
-              toggleOpen={toggleOpen} 
-              panelsRef={panelsRef} 
-              wearActions={wearActions} 
-              previewCanvasRef={previewCanvasRef}
-              game={game}
-            />
-            <World
+            <Tab
```
avaer, line 482: Why is this changed?

avaer (changes requested):
Was there a bad merge?

src/Header.jsx
yaserOSource: Will fix, will revert this to main branch

yaserOSource: It's the linting tool

yaserOSource:
> Was there a bad merge?

yes there was

yaserOSource:
PR updated

yaserOSource:
I will test it out some more before passing it through QA

deadcode/editor.js
```diff
@@ -2558,14 +2558,14 @@ const loadHash = async hash => {
...
-  const u = `${storageHost}/ipfs/${hash}/.metaversefile`;
+  const u = `${storageHost}/ipfs/${hash}/.totum`;
   loadedObject = await world.addObject(u);
   loadedObject.name = loadedObject.contentId.match(/([^\/]*)$/)[1];
 };
 const mintNft = async () => {
   const hash = await uploadHash();
   console.log('upload nft', hash);
-  window.location.href = `https://webaverse.com/mint?hash=${hash}&ext=metaversefile`;
+  window.location.href = `https://webaverse.com/mint?hash=${hash}&ext=totum`;
```
avaer, line 2568: I don't think this part is right. We should not be renaming the actual file IMO. This is literally the file type, which can stay as `metaversefile`.

yaserOSource: got it ok, no problem

PR #1927, opened 2021-12-04
No autoupdate fix - Performance part 1

yaserOSource:
This PR manually calls object.updateMatrixWorld. 

Object3D.DefaultMatrixAutoUpdate = false;. will be something i will add next after this PR.

This PR has a dependency on this totum PR, as such must be checked out in packages/metaversefile(soon to be called metaversefile/totum):
https://github.com/webaverse/totum/pull/42

There is also a dependency on all of the apps. 

All the relevant scene apps are pointing to my repository for ease of testing purposes, and will be merged into webaverse/[sceneapp] with a PR after code review.


avaer (approved):
Code looks good, though we shouldn't merge until the individual apps are merged.

avaer:
Found lots of bugs so far:

- glbb type positioned wrongly
image

- `E` does not update
image

- Chest does not animate at all
image

- Damage mesh does not move
image

I haven't even gotten started with testing, so it looks like there are many regressions to fix still.

yaserOSource:
image

glbb placement looks correct on mine

avaer (approved):
Code still looks good, but I recommend thorough testing.

That includes the screenshot system, since this touches it.

PR #1928, opened 2021-12-05
Clean up animations mappings

avaer:
This adds an abstract `animationMappingConfig`, which can be used by clients to parse the `animations` data array.

This is needed for AR support.


avaer:
This also removes our old `outputs` array, replacing it with `modelBoneOutputs`.

PR #1945, opened 2021-12-06
Performance Optimisation WIP

yaserOSource:
This is to start the next phase of optimisations on the App.

`THREE.Object3D.DefaultMatrixAutoUpdate = false;`

Starting with this line and we end up with an application that looks like this:

image

Working through each aspect ( models, cameras, objects etc ) and updating


yaserOSource:
Im also doing further profiling

PR #1946, opened 2021-12-07
Multiplayer fixes

avaer:
This fixes some cases where type info was missing from the Z.js bindings implementation, causing types errors when connecting in multiplayer.


avaer:
This should unlock multiplayer. There are still bugs to fix.

PR #1958, opened 2021-12-08
Animation baker windows support

avaer:


avaer:
To re-bake animations, window or not, we can now simply do:

```
node animations-baker.mjs
```

PR #1959, opened 2021-12-08
Add new jump animations

avaer:


avaer:
This also rebuilds `animations.cbor`

PR #1970, opened 2021-12-11
Login and modular fe

abeersaqib:
1. FE is converted to modular parts
2. Vite redirect added for Login

Code deployed at: https://staging.webaverse.com


index.mjs
```diff
@@ -53,6 +53,9 @@ function makeId(length) {
         req.originalUrl = u;
         next();
       }
+    } else if (/^\/login/.test(o.pathname)) {
```
avaer, line 56: `/^\/login$/`?

avaer (changes requested):
Can this be two PR's?

This has a very high likelihood of being reverted when a regression is discovered so let's make that clean to do so.

index.mjs
abeersaqib: $ will match with the end of the string. While our express don't end at login it has id & code as well.

abeersaqib:
> Can this be two PR's?
> 
> This has a very high likelihood of being reverted when a regression is discovered so let's make that clean to do so.

Sure

PR #1973, opened 2021-12-11
Integrated Monetization setup

abeersaqib:
will add description later


avaer (changes requested):
There's no description, so I don't know what this is supposed to accomplish.

lalalune:
Closing this PR because doesn't seem like any owner and isn't useful

PR #1974, opened 2021-12-11
[WIP] Screenshot iframe2

abeersaqib:
In App Preview service that open iframe from the weba.online. weba.online uses the dashboard worker to point to the weba.com 

It has it's own queuing systme and can use it based on the load.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -32,6 +32,7 @@ export const polygonVigilKey = `0937c004ab133135c86586b55ca212a6c9ecd224`;
 
 export const storageHost = 'https://ipfs.exokit.org';
 export const previewHost = 'https://preview.exokit.org';
+export const inappPreviewHost = 'https://app.webaverse.online';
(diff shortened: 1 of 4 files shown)


avaer:
Is this for https://github.com/webaverse/app/issues/1896 ?

abeersaqib:
> Is this for #1896 ?

Yes it's for the same PR. it requires worker to be fixed for the QA testing.

avaer:
We don't need to fiddle with CloudFlare Workers. If that's too hard, we can just start a server which will `node-http-proxy` and add the headers we need.

abeersaqib:
> We don't need to fiddle with CloudFlare Workers. If that's too hard, we can just start a server which will `node-http-proxy` and add the headers we need.

Yeah exactly. But i made it work thanks to your proxy suggestion.

avaer (approved):

Issue #1938, opened 2021-12-05
Naruto run double tap WASD


avaer:
Right now the naruto run is activated with double shift.

It should be activated with double tap of WASD keys _while holding shift_. This would feel more natural, while preventing accidental double taps of WASD while walking with precision, which would be annoying.

skaljac:
Implemented in https://github.com/webaverse/app/pull/1995

Issue #1918, opened 2021-12-03
NarutoRun + crouch


skaljac:
Naruto Run from crouch mode should not be possible.

Steps to reproduce:
Press C for crouch mode
Press Double shift to activate Naruto Run

This is how it looks

https://user-images.githubusercontent.com/4487857/144668920-1593486b-5b10-4ff9-9f24-135f2f8410a8.mp4

avaer:
Related: https://github.com/webaverse/app/issues/1938

skaljac:
Implemented in https://github.com/webaverse/app/pull/1995

PR #1995, opened 2021-12-13
Naruto run fix

skaljac:
This PR is solving 2 issues:
https://github.com/webaverse/app/issues/1938
https://github.com/webaverse/app/issues/1918

Steps for testing:
Case 1
Press C for crouch
Activate Naruto run with Shift and then double press on button W

* Naruto run should not be possible from crouch mode

Case 2:
Press and hold shift key
double press W or A or S or D and Naruto run should be activated

Double press on Shift key should not activate Naruto run


io-manager.js
```diff
@@ -64,8 +64,8 @@ ioManager.keys = {
   space: false,
   ctrl: false,
 };
-let lastShiftDownTime = 0;
-ioManager.getLastShiftDownTime = () => lastShiftDownTime;
+let lastWASDDownTime = [0,0,0,0];
```
avaer, line 67: I would label this with keys, not mysterious indices.

skaljac: Changing that now

avaer (approved):

amnaarshed4224:
It seems to work fine. A few clarifications:
-Holding +double tap WASD activates the Naruto run/fly/travel(Dragon, hovercraft). Once activated, we don't need to hold WASD key, just holding shift keeps avatar in Naruto mode. 
-While running when lets say shift+W is being already pressed, double tapping any other key A,S,D activates Naruto mode

Are these intentional?

avaer:
@skaljac ☝️

skaljac:
@amnaarshed4224 @avaer  From original issue it is not clear how exactly this should work. I taught that we should be able to change direction while we are in naruto run mode. This is only possible if we break naruto mode run only when shift key is up. I can easily change this behaviour if we need that. Please let me know what to do?

io-manager.js
```diff
@@ -296,13 +302,29 @@ ioManager.keydown = e => {
       if (!document.pointerLockElement) {
         game.menuVertical(-1);
       }
+
+      const now = Date.now();
```
avaer, line 306: Should use `performance.now()`, which is what we use elsewhere.

io-manager.js
```diff
@@ -296,13 +302,29 @@ ioManager.keydown = e => {
...
     }
     case 65: { // A
       ioManager.keys.left = true;
       if (!document.pointerLockElement) {
         game.menuHorizontal(-1);
       }
+
+      const now = Date.now();
+      const timeDiff = now - lastWASDDownTime.keyA;
+      if (timeDiff < 200 && ioManager.keys.shift) {
```
avaer, line 323: `200` should be a constant.

Issue #1909, opened 2021-12-02
App: footstep sound effect synchronization


amnaarshed4224:
Description: 
There is still lack of sync between footsteps sound and animation. It turns out it can be an fps dependent issue.
One example attached. I will attach more examples soon.

Related PR:
#1864 

https://user-images.githubusercontent.com/35926530/144501937-375b1924-0b10-4e21-9e6b-1c95a22b3056.mp4

ahadshams:
also @skaljac i noticed that naruto run has the same footsteps as normal run

amnaarshed4224:
being redone.

PR #1996, opened 2021-12-13
Footstep sound effect synchronization

skaljac:
Related issue:
https://github.com/webaverse/app/issues/1909

Steps for testing:
Footsteps for walk,run,naruto run should be synced with animation

In fly mode there should not be any sound (jump, jump end etc)


avaer:
IMO this needs a redo, without any hardcoded constants and simply watching the leg animation to trigger.

sound-manager.js
```diff
@@ -46,9 +46,11 @@ class SoundManager {
     }
     
     const isJumping = this.localPlayer.hasAction('jump');
-    //const isFlying = gameManager.isFlying();
-    
-    if (this.lastIsJumping && !isJumping) {
+    const isNarutoRun = this.localPlayer.hasAction('narutoRun');
```
avaer, line 49: Unused variable.

skaljac:
New approach to this issue is available here:
https://github.com/webaverse/app/pull/2025

That is the reason for closing

PR #1998, opened 2021-12-14
Part 2 of performance optimisation WIP.

yaserOSource:
avatars.js needs to be updated as well as shoulder transforms.

image

Changing this to :

image

Causes some issues in the avatar, looks like the animation is blending but the bones are not in their proper places.

Avatar z forward is also wrong.

Aside from this we can merge asap.

Requires Totum : `performance-optimisations` branch

update line 1435 on avatars.js to see the updated avatar with distorted bones.
```for (const k in modelBones) {
      const modelBone = modelBones[k];
      const modelBoneOutput = modelBoneOutputs[k];
      modelBoneOutput.updateMatrixWorld(true);
      modelBone.matrix.setPosition(modelBoneOutput.position);
      const modelBoneQuat = modelBone.quaternion;
      modelBoneQuat.multiplyQuaternions(
        modelBoneOutput.quaternion,
        modelBone.initialQuaternion
      );

      modelBone.matrix.makeRotationFromQuaternion(modelBoneQuat);
      modelBone.updateMatrixWorld(true);```

### CODE REVIEW

File: player-avatar-binding.js [hunk 1]
```diff
@@ -9,12 +9,16 @@ const avatarSymbol = 'avatar'; // Symbol('avatar');
 
 export function applyPlayerTransformsToAvatar(player, session, rig) {
   if (!session) {
-    rig.inputs.hmd.position.copy(player.avatarBinding.position);
-    rig.inputs.hmd.quaternion.copy(player.avatarBinding.quaternion);
-    rig.inputs.leftGamepad.position.copy(player.leftHand.position);
-    rig.inputs.leftGamepad.quaternion.copy(player.leftHand.quaternion);
-    rig.inputs.rightGamepad.position.copy(player.rightHand.position);
-    rig.inputs.rightGamepad.quaternion.copy(player.rightHand.quaternion);
+    rig.inputs.hmd.matrix.setPosition(player.avatarBinding.position);
```
avaer, line 12: Imo this would be better with `.compose`, or setting the positions/quats/scales and then `updateMatrixWorld(true)`.

In fact I think this has a bug, because `makeRotationFromQuaternion` completely overwirtes the matrix: https://github.com/mrdoob/three.js/blob/dev/src/math/Matrix4.js#L289

yaserOSource: good point

avatars/avatars.js
```diff
@@ -1551,12 +1520,18 @@ class Avatar {
     rightToe: 'Right_toe'
   }
   initializeBonePositions(setups) {
-    this.shoulderTransforms.hips.position.copy(setups.hips);
-    this.shoulderTransforms.spine.position.copy(setups.spine);
-    this.shoulderTransforms.chest.position.copy(setups.chest);
-    if (setups.upperChest) this.shoulderTransforms.upperChest.position.copy(setups.upperChest);
-    this.shoulderTransforms.neck.position.copy(setups.neck);
-    this.shoulderTransforms.head.position.copy(setups.head);
+    this.shoulderTransforms.hips.matrix.setPosition(setups.hips);
```
avaer, line 1523: This could use `.compose` instead of `.setPosition`.

avatars/avatars.js
```diff
@@ -1445,6 +1412,8 @@ class Avatar {
     
     modelBones.Root.traverse(bone => {
       bone.initialQuaternion = bone.quaternion.clone();
+      bone.updateMatrix();
```
avaer, line 1415: Probably not correct.

avatars/avatars.js
```diff
@@ -933,6 +933,7 @@ class Avatar {
 
     // height is defined as eyes to root
     this.height = getHeight(object);
+    console.log(this.height);
```
avaer, line 936: No need for this log.

player-avatar-binding.js
yaserOSource: ill rewrite this

yaserOSource: >or setting the positions/quats/scales and then updateMatrixWorld(true).

Does not work with THREE.Object3D.DefaultMatrixAutoUpdate  set to false

PR #2007, opened 2021-12-14
Full use/aim animation premultiply blend

avaer:
This PR makes the aim and use animations premultiply against the current animation, with idle base inverted.

All bones are unlocked, but all of the use animations need to be fixed as a result.


avaer:
This PR already looks far superior to `master` with no real drawbacks that I could find, so let's merge it as a checkpoint. The attack animations will be bad especially when running, but it's still better than what we had before.

PR #2015, opened 2021-12-15
Performance Optimisation Part 2 WIP

yaserOSource:
This PR enabled use of DefaultMatrixAutoUpdate = False 

Which should give a performance gain.

If you set DefaultMatrixAutoUpdate = true, the application should look fine and still work. 

Still WIP but this approach shold insure nothing is broken.

### DIFF
--- a/app-manager.js
+++ b/app-manager.js
@@ -293,17 +293,11 @@ class AppManager extends EventTarget {
           app.updateMatrixWorld(true);
           //bad patch for race-condition 
           setTimeout(()=>{
-            app.traverse(e => {
+            mesh.traverse(e => {
               e.updateMatrix();
               e.updateMatrixWorld(true);
             })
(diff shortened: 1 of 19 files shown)


yaserOSource:
https://jsfiddle.net/vreo0m2d/13/ 

JSFiddle to test DefaultMatrixAutoUpdate = False

yaserOSource:
image

Avatar animation is blended with the root backwards it seems and the arms are facing up, i think its 1 or 2 lines from being fixed

yaserOSource:
image
Avatar can sit on stuff just fine but again, its facing backwards

yaserOSource:
image
Highlight mesh works on certain objects but not everything, for example objects that have many nested children, like the chest or homespace, or stacks, seems to miss the highlight mesh.

PR #2025, opened 2021-12-16
New sound footsteps synced with foot position [WIP]

skaljac:
This is still work in progress.

Footsteps are now synced with ankle local position.

With this new PR old one https://github.com/webaverse/app/pull/1996 is not needed.


character-sfx.js
```diff
@@ -0,0 +1,107 @@
...
+
+    const boneFootLeft = this.player.avatar.foundModelBones['Left_ankle'];
+    const boneFootRight = this.player.avatar.foundModelBones['Right_ankle'];
+
+    let posLeft = 0;
+    let posRight = 0;
+
+    //console.log(boneFootLeft);
+
+    boneFootLeft.matrix.copy(boneFootLeft.matrixWorld).
```
avaer, line 55: Shouldn't mutate the bone here, it belongs to the avatar implementation...

character-sfx.js
```diff
@@ -0,0 +1,107 @@
...
+        premultiply(localMatrix.copy(this.player.avatar.modelBoneOutputs.Root.matrixWorld).invert());
+
+    boneFootLeft.matrix.decompose(localVector, localQuaternion, localVector2);
+
+    posLeft = localVector.z;
```
avaer, line 60: Why z? Shouldn't it be Y?

Either 1) this is a mistake, or 2) this is compensating for some weird rotation axis switch, which we should be multiplying out to do correct math, rather than switching axis here as a hack.

avaer (changes requested):
This is a good start, but I think it still has a ways to go.

Some code weirdnesses found and it doesn't work very well. I think there are some obvious bugs here, but this is roughly the right approach, it just needs to be fixed and optimized.

avaer:
This was bugged and has long been superseded by an animations-based footstep detection system.

Issue #1981, opened 2021-12-12
Flash of wrong position when grabbing items


avaer:
You can see how it pops up way too high at the end for one frame:

https://user-images.githubusercontent.com/6926057/145732790-af54bc77-c1ca-4414-8eec-7f2187f9cadc.mp4

skaljac:
Solved with
https://github.com/webaverse/app/pull/2061

PR #2027, opened 2021-12-17
Fix for flash of wrong position when grabbing items

skaljac:
Issue is described here:
https://github.com/webaverse/app/issues/1981

Steps for testing:
Grab item (for instance Pistol)
When player goes up observe is there a strange head movement

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -637,7 +637,7 @@ class UninterpolatedPlayer extends Player {
     
     this.actionInterpolants = {
       crouch: new BiActionInterpolant(() => this.hasAction('crouch'), 0, crouchMaxTime),
-      activate: new UniActionInterpolant(() => this.hasAction('activate'), 0, activateMaxTime),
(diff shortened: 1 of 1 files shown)


amnaarshed4224:
This PR serves its purpose. There is much improvement in avatar getting up while equipping collectibles. I have attached the video here. There's one thing I want to address. Its below the video.

https://user-images.githubusercontent.com/35926530/146585500-4738c1b6-5763-4156-a4cc-ce85011b559d.mp4


One thing I have observed is that, While trying to ride the hovercraft, the avatar rides it but the camera moves upside as if the avatar is getting up. Same happens while the avatar tries to equip things while flying, Though there's no animation for equipping objects while flying.

https://user-images.githubusercontent.com/35926530/146585570-d444ebf9-ea6f-48e0-ac96-0d93044c5064.mp4

avaer:
This PR is irrelevant because we have much better use animations now. 😸

PR #2036, opened 2021-12-19
Three.js local package 

avaer:
This PR makes `three` a local package in `packages/three`, implemented as a submodule.

This is so we can add our own patches to it, which is needed to perform synchronous buffer attribute updates.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "packages/three-vrm"]
 	path = packages/three-vrm
 	url = https://github.com/webaverse/three-vrm
+[submodule "packages/three"]
+	path = packages/three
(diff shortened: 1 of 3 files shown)


avaer:
We don't need to to this for now because we found a workable hack which does not require us to vendorize three.js (which has its own set of problems like breaking peer dependencies): https://github.com/webaverse/silsword/pull/1/commits/4156b8df561bd4e5587f4a3ad3fb524aebc8fa2d

In this hack we can simply delay the geometry upload so that we do one chunk per frame. In almost all cases, this simply works as intended, and the lag is at most one frame in the worst case, which is imperceptible on highly dynamic geometry.

Issue #2001, opened 2021-12-14
App: dropped object physics out of sync


avaer:
https://user-images.githubusercontent.com/6926057/145977048-a12543b2-8f85-4fd4-a625-9eddb0d6613a.mp4

When you pick up e.g. the silsword and then use the character UI to drop it, the object will no longer be collidable in its physics (we can confirm with tilde mode).

Additionally, we should place the object in front of the player a little above the floor, rather than leaving it in place where it was worn.

Soon we will activate true physics for the drop, so the item falls to the floor, but that is separate from this issue.

avaer:
I'm not sure if this happens for all items or only some, but we should test and fix as many of of the known cases as possible.

avaer:
I noticed that the object is being placed in front of the player supposedly, but the real mesh is not updated from the physics mesh.

![image](https://user-images.githubusercontent.com/6926057/145977450-27322f17-2cfa-4da4-97d0-103bb83d12c1.png)

PR #2047, opened 2021-12-20
Dropped object physics out of sync [WIP]

skaljac:
Related issue:
https://github.com/webaverse/app/issues/2001

So far fixed objects are:
Pistol (separate PR)
Potion

Objects that already works:
Sword

Left for fixing:
Helm


avaer:
I think the potion case here has long since been fixed in `master`. Therefore closing this PR.

PR #2057, opened 2021-12-22
added angularVelocity functions and updated physx wasm

Loryhoof:
~~moved all the vehicle controlling code to glb.js~~
Added setAngularVelocity function
~~New keys in ioManager.keys~~


metaversefile-api.js
```diff
@@ -480,6 +481,9 @@ metaversefile.setApi({
       },
     };
   },
+  useIO() {
+    return ioManager;
```
avaer, line 485: IMO apps should not be touching `ioManager`, that is a major can of worms.

If we need a new control mode then that can be handled here in `app` by setting the mode, rather than giving raw control to apps.

avaer (approved):

avaer:
This was not mentioned in the description, but this PR requires https://github.com/webaverse/app-wasm/compare/angularVel?expand=1

Note geometry.wasm has been updated since this PR was opened, so it needs a merge and rebuild.

avaer:
https://github.com/webaverse/app-wasm/pull/11 was merged. This PR needs a resolve.

avaer:
I couldn't get the helicopter hack to load with this branch.

image

Would be cool to get it working though.

avaer:
Note that modularization of glb.js components has started in https://github.com/webaverse/app/pull/2076. The rest of the default components need to be moved over.

avaer:
Closing stale PR.

Issue #1873, opened 2021-11-26
Animation : Grab Animations


plankatron:
here is a set of grab animations , they are a first draft , not a final draft. they will need coder support in engine to be able to test all of them. 

* grab_forward
* grab_down
* grab_left
* grab_right
* grab_up
https://drive.google.com/drive/folders/1BbOrhq70V2fJm10lwbUJnULviK-VU66s?usp=sharing

skaljac:
I have send request for access.

skaljac:
Implemented with
https://github.com/webaverse/app/pull/2061

Issue #1825, opened 2021-11-17
App: group objects cannot be found when worn


avaer:
![image](https://user-images.githubusercontent.com/6926057/142285449-ba08eea2-a85d-42b6-bdc8-a3da88fc77f9.png)

The lantern wearing (grabbing) crashes because the lantern is part of a group and independently controlled, but it is not part of the player's app manager. Therefore the wear logic cannot find the app by its instance id; it is not tracked anywhere besides the group.

To solve this we need to either:

1) make sure that only the top level group object is grabbed, it shouldn't know about the children, and the children can just get tehir state downstream
2) we can make a new app manager for these internal sub cases, collect these apps there, and then look there for lookup during the wear logic

My vote is for 1).

Lantern app: https://github.com/webaverse/lantern
Where the lookup fails: https://github.com/webaverse/app/blob/6e4c66a2de41cf72ec7fc6dcca75bcab54dacdd6/player-avatar-binding.js#L30
Where the lantern subapp is created: https://github.com/webaverse/metaversefile/blob/9fa1000110181c91d2f76a9b4fadc3a7c4187606/type_templates/group.js#L93

PR #2061, opened 2021-12-22
New grab animations

skaljac:
Related issues:
https://github.com/webaverse/app/issues/1981
https://github.com/webaverse/app/issues/1873
https://github.com/webaverse/app/issues/1825

In this PR we have new animations for item grab action.
Animations are: 
grab forward
grab down
grab up
grab left
grab right

Also "group objects cannot be found when worn" is already solved by Avaer but we just need one simple null check and lantern/fairy can be picked up / droped without any problem.

Steps for testing:
Prepare scene with pickable objects at different heights.
Check object grab / pick up in directly in front of player, on players left side, on players right side.
Also place pickable object above players head, and on the floor. 
Appropriate animation should be shown.

Plase take a look at video below for reference


skaljac:
https://user-images.githubusercontent.com/4487857/147165723-c9874aee-9904-4444-88c0-695ae258daf4.mp4

game.js
```diff
@@ -705,6 +705,61 @@ const _gameUpdate = (timestamp, timeDiff) => {
...
+    if (distance < currentDistance) {
+      currentDistance = distance;
+      currentAnimation = "grab_right";
+    }
+
+    if (localPlayer.getAction('activate')) {
+      localPlayer.getAction('activate').animationName = currentAnimation;
+    }
+
+    return {distance:currentDistance,
```
avaer, line 759: There is no need to return here; it counts as an allocation and isn't used.

skaljac: I have removed this return statement now

amnaarshed4224:
Issues:
1-The animation of grabing itself seems right but it seems to have some issues with direction of grab. (the direction in which hand moves to pick an object)
2-When avatar is facing opposite the object the direction of grab is totally opposite.
3-When tried to pick object diagonally same issue of direction persists
4-Trying to grab the item above or in any direction(below, left right), If the avatar is standing at bit distance, the hand doesn't seems to be going in the direction of the object.

https://user-images.githubusercontent.com/35926530/147277095-4f91888f-8856-4069-94c9-375934b4b851.mp4


https://user-images.githubusercontent.com/35926530/147277150-e7d5a661-d579-4473-bf2a-deb3bcab7c70.mp4

avaer:
Yeah I tested this and I think the grab directions might not account for object rotation.

skaljac:
I have change the way how this works. If the player is not facing item, he will not get E button.

I hope this is what ok now. If not please let me know what should I change.

https://user-images.githubusercontent.com/4487857/147289027-88c1296a-e0c0-4077-93ae-a45d029a1ecb.mp4

avaer:
Tried it and it works fine; it certainly looks a lot better than `master`.

We can continue improvements there.

PR #2073, opened 2021-12-24
Physics hijack removal

avaer:
The metaversefile implementation was hijacking the transforms in the overridden methods, causing downstream app problems.

This removes that hijack and additionally, correctly initializes the physics object local transform when creating a new physics object.


avaer:
Although I tested, I suspect this may cause problems in some apps in the scene. In those cases I would say it's a bug in the app, if it was relying on this old behavior.

PR #2088, opened 2021-12-27
Post Scene with configurable camera

Sirahi:
User can now add post scene with ortho or perspective camera in application/rendersettings
{
"content": {
"postPostProcessScene": {
"perspectiveScene": false,
"orthographicScene": true
}
}
}


avaer (approved):

world.js
```diff
@@ -301,8 +301,11 @@ const _getBindSceneForRenderPriority = renderPriority => {
     case 'low': {
       return sceneLowPriority;
     }
-    case 'postScene': {
-      return postScene;
+    case 'perspectiveScene': {
```
avaer, line 304: This can be renamed to `postPerspectiveScene` and similar, to indicate its actual order.

avaer:
Makes sense but recommend a small rename in the API.

avaer (approved):

PR #2096, opened 2021-12-28
[WIP] Test cases removed Certs folder to stop puppeteer pointing to local.weba

abeersaqib:


avaer (changes requested):
I don't really like this. Anyone running tests locally will corrupt their checkout as a result.

If the environment needs to change, we can use env vars and change behavior based on test mode or not.

abeersaqib:
Closing PR as created a new PR for this issue

PR #2109, opened 2021-12-29
Added isSceneLoaded() hook in metaversefile-api

Sirahi:


avaer (changes requested):
What if it's a scene change?

avaer (approved):

Issue #2070, opened 2021-12-24
Postprocessing colour banding


yaserOSource:
https://user-images.githubusercontent.com/51108458/147325503-e3c0fb23-a443-4384-845b-2a90beb6a14c.mp4

https://user-images.githubusercontent.com/51108458/147325534-22fdc470-6f9b-4f6e-8c7f-bfca6c86f136.mp4

You can see in both scenes that there is colour banding in the dark areas and also areas with light. The light emitted looks low quality and is represented by 2-3 tones when it should be a whole lot more detailed. 

Similarly the textures in dark areas also look distorted.

yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/147325777-8ab14105-e622-471e-9d44-2aa37e667dca.png)

yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/147325884-5bf2aa1b-a1ee-462b-b702-f4f43e59c0d8.png)

yaserOSource:
As far as I can tell this is a THREE issue

avaer:
Is it only with post processing? Does changing the framebuffer color depth affect it? What about changing the color encoding?

yaserOSource:
Good point, Let me test and get back to you

yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/147511775-daf75e0a-2452-4235-ab9f-ff3c46451712.png)

What happens when i disable post processing, everything except the encoding pass.

yaserOSource:
https://en.wikipedia.org/wiki/Colour_banding

yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/147511883-13d984cf-7235-4e9d-94ce-9696ac5d1717.png)

Without SSAO its less obvious but the bloom still causes some banding

yaserOSource:
Changing the encoding on the renderer doesn't change anything, same result, so will try the encoding pass.

yaserOSource:
I suggest turning off  SSAO and three bloom and trying to find alternatives that will behave

yaserOSource:
I will look at alternatives for bloom

yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/147712913-acc3161c-e21d-44be-b5a0-a51dd9442703.png)

Dual bloom is better for the bloom overall - more difficult to get right for every scene, and will definitely need settings per scene

[...]

PR #2112, opened 2021-12-30
Postprocessing colour banding fixes WIP

torchesburn:
https://github.com/webaverse/app/issues/2070


yaserOSource:
Just thought adding in new bloom would improve the bloom at least, which it does, but still some banding issues.


Likely the encoding is not right across the shaders, but the final encoding pass should have taken care of that, instead we get some colour banding issues

avaer:
This seems worth keeping around for later.

avaer:
Possibly related: https://github.com/webaverse/app/pull/2198

PR #2122, opened 2022-01-01
Use glmw to replace threejs' matrix calculation.

gonnavis:
Use [glmw](https://github.com/maierfelix/glmw) to replace threejs' matrix calculation.

The substitution of `multiplyMatrices()` and `compose()` is preliminarily completed.
But `compose()` is disabled, because I tested it on my computer, the performance of the glmw version is not even as good as the original one.

According to the profile and print data, the glmw version has a considerable speed improvement on the `multiplyMatrices()`.
Original vs glmw:
Time consumption ratio: 9.3% vs 3.4%.
Average time per execution: 0.00094 vs 0.00061.
image
image
image
image
But the strange thing is that I don't see a significant fps increase on my computer. So still need everyone to help with more tests to see if the actual performance is really improved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "dids": "^2.4.0",
     "express": "^4.17.1",
     "forever": "^4.0.1",
+    "glmw": "^0.1.3",
     "gpt-3-encoder": "^1.1.3",
     "json-6": "^1.1.4",
     "key-did-provider-ed25519": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "react-dom": "17.0.2",
     "smile2emoji": "^1.1.6",
     "stats.js": "^0.17.0",
-    "three": "^0.133.1",
+    "three": "https://github.com/gonnavis/three.js/tarball/WebaverseR133D1",
     "troika-three-text": "^0.43.1-alpha.0",
     "webm-writer": "^1.0.0",
     "wsrtc": "./packages/wsrtc",
--- a/renderer.js
+++ b/renderer.js
@@ -26,52 +26,53 @@ init().then((ready) => {
     const mat4b = mat4.create();
     const mat4aView = mat4.view(mat4a);
     const mat4bView = mat4.view(mat4b);
-    let a;
-    let b;
+    let ae;
+    let be;
     // let c;
 
     return function () {
       // your code
+      // return this
 
       // var result = cachedFunction.apply(this, arguments) // use .apply() to call it
 
       // more of your code
-      a = arguments[0].elements;
-      b = arguments[1].elements;
+      ae = arguments[0].elements;
+      be = arguments[1].elements;
 
-      mat4aView[0] = a[0]
-      mat4aView[1] = a[1]
-      mat4aView[2] = a[2]
-      mat4aView[3] = a[3]
-      mat4aView[4] = a[4]
-      mat4aView[5] = a[5]
-      mat4aView[6] = a[6]
-      mat4aView[7] = a[7]
-      mat4aView[8] = a[8]
-      mat4aView[9] = a[9]
-      mat4aView[10] = a[10]
-      mat4aView[11] = a[11]
-      mat4aView[12] = a[12]
-      mat4aView[13] = a[13]
-      mat4aView[14] = a[14]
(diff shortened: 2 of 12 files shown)


renderer.js
```diff
@@ -6,6 +6,101 @@ the purpose of this file is to hold these objects and to make sure they are corr
...
 import {minFov} from './constants.js';
+import {init, vec3, vec4, mat4} from 'glmw';
+
+// https://stackoverflow.com/a/62544968/3596736
+// Adding 'support' for instanceof Proxy:
+(() => {
+  var proxyInstances = new WeakSet();
+
+  // Optionally save the original in global scope:
+  window.OriginalProxy = Proxy;
```
avaer, line 17: We shouldn't do these global overrides. There should be no reason for it.

package.json
```diff
@@ -27,6 +27,7 @@
     "dids": "^2.4.0",
     "express": "^4.17.1",
     "forever": "^4.0.1",
+    "glmw": "^0.1.3",
```
avaer, line 30: Let's merge the WASM functionality we need into our own wasm build; we should keep the number of WASM modules to a minimum.

renderer.js
```diff
@@ -6,6 +6,101 @@ the purpose of this file is to hold these objects and to make sure they are corr
...
+  THREE.Matrix4.prototype.multiplyMatrices = (function() {
+    const mat4a = mat4.create();
+    const mat4b = mat4.create();
+    const mat4c = mat4.create();
+    const mat4aView = mat4.view(mat4a);
+    const mat4bView = mat4.view(mat4b);
+    const mat4cView = mat4.view(mat4c);
+
+    return function multiplyMatrices() {
+      if (Array.isArray(this.elements)) {
```
avaer, line 46: This replacement should probably happen at Object3D create time.

In which case we might need a truly vendorized THREE.js, if we don't want to create extra throwaway `.element` arrays.

renderer.js
```diff
@@ -6,6 +6,101 @@ the purpose of this file is to hold these objects and to make sure they are corr
...
+        // NOTE: Replace the ordinary array with Float32Array to support the direct overall assignment by calling the `set` method, instead of assigning the 16 elements one by one.
+        this.elements = new Float32Array(this.elements);
+      } else if (this.elements instanceof Proxy) {
+        // NOTE: `localPlayer.avatar.modelBones.Root.matrixWorld.elements` will be replaced with `Proxy` after loading is complete, causing an error in calling the `set` method.
+        // I guess it is React's binding mechanism that generates the `Proxy`. After removing the `Proxy` in the following way, the main game part seems to still work normally, but I am not sure what effect it will have on the UI logic.
+        this.elements = new Float32Array(Object.values(this.elements));
+      }
+
+      mat4aView.set(arguments[0].elements);
```
avaer, line 55: There should be no need to copy elements.

avaer (changes requested):
I think we can do a lot better than this.

We can allocate all of the `.elements` in a single pass on `render()`, copy the tree structure to WASM, and then do the entire multiplication of the tree in one WASM call.

The tree walk can be further optimized by setting dirty flags on Objecr3D `.add` and `.remove`.

gonnavis:
Now `Matrix4.elements` totally directly using memories belong to wasm, no any data copying between js and wasm.
The profiled time consumption ratio greatly decreased from master 9.3% and previous 3.4% to now 1.5%.
And the fps on my pc also noticeably increased from 111 to 125. 
image

Though still using a lot of `window.xxx` hacking and third party `glmw` lib.
And wasm must be initialized asynchronously, which brought me a lot of trouble.
I'll keep investigating.

renderer.js
gonnavis: Now no any data copying between js and wasm.

renderer.js
gonnavis: Forked threejs lib and modified.
Now can use `npm install https://github.com/gonnavis/three.js/tarball/WebaverseR133D1` to install modified threejs version.

But wasm must be initialized asynchronously prevent me initiate `elements` with wasm at start.

gonnavis:
Found that the `Proxy` used in three-vrm lib has the greatest impact on performance.
The current wasm implementation does not improve performance, just disabled Proxy as a side effect.

PR #2124, opened 2022-01-01
Diorama support

avaer:
Adds a new fully shaded diorama system, for chat, characters, npc, and playing out scenes.

Currently enabled with secret shortcut for testing.


nguoithulinh2021:
Vvvv

PR #2163, opened 2022-01-05
Character hups system

avaer:
This PR adds a new character "heads up player" system, intended to track player actions and play dioramas accordingly.

The idea behind this is that any occurrence worth having a diorama for (hits, jumps, chats, emotes, player joins) should be a true player action (`player.addAction(action)`), which is then picked up by the hups system to render the diorama. This design means that it should "just work" for the multiplayer case.


avaer:
This still needs to replace the old chat modes (world space bubbles) with the new hups.

PR #2164, opened 2022-01-05
healthControl.js added, with functions to damage player and track/display hp and mana

bru345:
You will need https://github.com/webaverse/npc/pull/1 PR to test the 'takeDamage' functionality.

### DIFF
--- a/avatars/healthControl.js
+++ b/avatars/healthControl.js
@@ -1,17 +1,33 @@
 
+import gameManager from '../game';
+import metaversefile from 'metaversefile';
 export class HealthControl {
-  constructor() {
+  constructor(isLocalPlayer) {
     this.health = 100;
     this.mana = 100;
+    this.isLocalPlayer = isLocalPlayer;
+    window.addEventListener('DOMContentLoaded', () => {
+      this.display();
+    });
   }
 
   takeDamage(damage) {
-    console.log('damage');
     this.health -= damage;
     if (this.health <= 0) {
       this.health = 0;
+      console.log('player is incapacitated')
       // play death animation etc
     }
+    // gameManager.playerDiorama.toggleShader();
+    // window.dispatchEvent(new MessageEvent('playerHit'))
+    this.display();
+  }
+
+  display() {
+    if (this.isLocalPlayer) {
+    document.getElementById('healthDisplay').innerHTML = `HP: ${this.health}`;
+    document.getElementById('manaDisplay').innerHTML = `MP: ${this.mana}`;
+  }
   }
 
   isDead() {
--- a/avatars/healthControl.js
+++ b/avatars/healthControl.js
@@ -1,0 +1,20 @@
+
+export class HealthControl {
+  constructor() {
+    this.health = 100;
+    this.mana = 100;
+  }
+
+  takeDamage(damage) {
+    console.log('damage');
+    this.health -= damage;
+    if (this.health <= 0) {
+      this.health = 0;
+      // play death animation etc
+    }
+  }
+
+  isDead() {
+    return this.health < 1;
+  }
+}
--- a/scenes/treehouse.scn
+++ b/scenes/treehouse.scn
@@ -171,7 +171,7 @@
         0,
         1
       ],
-      "start_url": "https://webaverse.github.io/npc/"
+      "start_url": "https://bru345.github.io/npc/"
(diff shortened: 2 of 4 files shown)


avatars/healthControl.js
```diff
@@ -0,0 +1,36 @@
+
+import gameManager from '../game';
```
avaer, line 2: We should use the full extension here (`game.js`) or it will not be portable.

scenes/treehouse.scn
```diff
@@ -171,7 +171,7 @@
         0,
         1
       ],
-      "start_url": "https://webaverse.github.io/npc/"
+      "start_url": "https://bru345.github.io/npc/"
```
avaer, line 174: https://github.com/webaverse/npc/pull/1 was merged. Can we switch this back to not be a fork?

src/tabs/character.js
```diff
@@ -29,11 +30,11 @@ export const Character = ({open, game, wearActions, panelsRef, setOpen, toggleOp
             <div className={classnames(styles['panel-section'], styles['name-placeholder'])} />
             <div className={classnames(styles['panel-section'], styles['main-stats'])}>
               <div className={styles['panel-row']}>
-                <h2>HP</h2>
+                <h2 id = "healthDisplay">HP </h2>
```
0xmetabad, line 33: nit: The other attributes in this file do not having spaces, it is standard to not have spaces here.

```html
<h2 id="healthDisplay">HP </h2>
```

src/tabs/character.js
```diff
@@ -29,11 +30,11 @@ export const Character = ({open, game, wearActions, panelsRef, setOpen, toggleOp
...
                 <progress value={61} />
               </div>
               <div className={styles['panel-row']}>
-                <h2>MP</h2>
+                <h2 id = "manaDisplay">MP</h2>
```
0xmetabad, line 37: Same spacing here, mentioned in other comment: https://github.com/webaverse/app/pull/2164/files#r779201048

avatars/healthControl.js
```diff
@@ -0,0 +1,36 @@
...
+      // play death animation etc
+    }
+    // gameManager.playerDiorama.toggleShader();
+    // window.dispatchEvent(new MessageEvent('playerHit'))
+    this.display();
+  }
+
+  display() {
+    if (this.isLocalPlayer) {
+    document.getElementById('healthDisplay').innerHTML = `HP: ${this.health}`;
```
avaer, line 28: There is a better way to do this. We generally don't want to access deep DOM like this, because the engine is imported in many other contexts besides this, which might not have this DOM, or even any DOM.

Therefore, we do bindings explicitly and there is an entire bootstrap procedure dedicated to that:
https://github.com/webaverse/app/blob/1e6b31cab9efa3e23941fcde552fff145cdbadc4/src/App.jsx#L22
https://github.com/webaverse/app/blob/277711fa98e160496af609bd1200876c2c22db61/webaverse.js#L134

This kind of canvas should be injected through that procedure. Maybe this is the character system's `bindCanvas()` method.

Structuring the code that way ensures that environment dependencies have a clear point of control, rather than being sprinkled around deeply in the code like this. It makes it much easier to reason about and test the code.

avatars/healthControl.js
```diff
@@ -0,0 +1,36 @@
...
+    document.getElementById('manaDisplay').innerHTML = `MP: ${this.mana}`;
+  }
```
0xmetabad, line 30: nit: This is not indented correctly.

src/tabs/character.js
avaer, line 33: Can be injected with a ref/binding procedure.

avatars/healthControl.js
avaer, line 2: This entire file doesn't really belong in `/avatars/`. That directory is for the avatar system and nothing more, no gameplay at all.

The code in this is more part of the "players/characters" system, in the same family as `character-controller.js` and `character-physics.js`. So we can either create a directory for that or move this file up one level. Your call.

avaer (changes requested):
Small changes requested, but it looks good.

avaer:
Note that the treehouse scene is left crashing without this PR.

image

avaer:
Abandoned.

PR #2165, opened 2022-01-05
Webawallet

abeersaqib:
* Introduces a new component `wallet.js` that is responsible for communicating with the Wallet
* Removed major login functionality from User.JSX and moved to weba-wallet

### DIFF
--- a/src/components/wallet.js
+++ b/src/components/wallet.js
@@ -4,9 +4,19 @@
 import {walletHost} from '../../constants';
 
 let pk = '';
+const nativeStrings = {
+  postMessage: 'function () { [native code] }',
+  generateKey: 'function generateKey() { [native code] }',
+  freeze: 'function freeze() { [native code] }',
+  sign: 'function sign() { [native code] }',
+};
 
 class Wallet {
-  async launch() {
+  constructor() {
+    this.launched = false;
+  }
+
+  async waitForLoad() {
     const self = this;
     let pubk = '';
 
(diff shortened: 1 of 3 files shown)


src/components/wallet.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+      };
+      window.addEventListener('message', f);
+    });
+  }
+
+  async walletPromise(action, data, waitForAction, timeOutPromise) {
+    const self = this;
+
+    const m = await this.sign({
```
0xmetabad, line 58: This is the signed message / signature. Can this be a more human-readable name? Instead of `m`, we call it `signature`?

src/components/wallet.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+    });
+
+    return new Promise((resolve, reject) => {
+      self.iframe.contentWindow.postMessage({
+        action: 'signed_message',
+        message: m,
+      }, walletHost);
+
+      const f = event => {
```
0xmetabad, line 71: Can we give this a more human-readable name?

src/components/wallet.js
```diff
@@ -0,0 +1,116 @@
...
+          window.removeEventListener('message', f, false);
+          resolve(event.data.data);
+        }
+      };
+      window.addEventListener('message', f);
+    });
+  }
+
+  async sign(message) {
+    const ms = 'function sign() { [native code] }';
```
0xmetabad, line 83: `pk` in this code works because `pk` is an established shorthand for private key. But "ms" usually means milliseconds, could we rename this to be more human-readable?

ceramic.js
```diff
@@ -10,43 +10,43 @@ import {ceramicNodeUrl, metaverseProfileDefinition} from './constants.js';
...
 const login = async () => {
   const address = await getMainnetAddress();
-  const threeIdConnect = new ThreeIdConnect();
-  const authProvider = new EthereumAuthProvider(window.ethereum, address);
-  await threeIdConnect.connect(authProvider);
-  const provider = await threeIdConnect.getDidProvider();
+  // const threeIdConnect = new ThreeIdConnect();
+  // const authProvider = new EthereumAuthProvider(window.ethereum, address);
+  // await threeIdConnect.connect(authProvider);
+  // const provider = await threeIdConnect.getDidProvider();
```
0xmetabad, line 16: Are we commenting this code out because we are re-adding DID/Ceramic back soon? Or will we not need these commented out lines and can delete them?

abeersaqib: We might. Avaer suggested to comment them.

src/components/wallet.js
abeersaqib: Addressed

src/components/wallet.js
abeersaqib: Addressed

src/components/wallet.js
abeersaqib: Addressed

avaer (approved):

avaer (approved):

PR #2167, opened 2022-01-06
Token loading from Sidechain  & Token Previews Added

abeersaqib:
* Fixed token loading from the Sidechain
* Fixed token previews getting blocked on login from Metamask

### DIFF
--- a/src/tabs/tokens.js
+++ b/src/tabs/tokens.js
@@ -8,51 +8,89 @@ import metaversefile from '../../metaversefile-api.js';
 import {tokensHost} from '../../constants';
 import {preview} from '../../preview.js';
 
-export const Tokens = ({userOpen, hacks, address}) => {
+export const Tokens = ({userOpen, loginFrom, hacks, address}) => {
   const [nftPreviews, setNftPreviews] = useState({});
   const [nfts, setNfts] = useState(null);
   const [fetchPromises, setFetchPromises] = useState([]);
 
   useEffect(() => {
-    if (address && !nfts) {
+    if (address && !nfts && loginFrom) {
       setNfts([]);
-
       (async () => {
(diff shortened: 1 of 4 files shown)


src/tabs/tokens.js
```diff
@@ -8,51 +8,89 @@ import metaversefile from '../../metaversefile-api.js';
...
-          },
-        });
-        const j = await res.json();
-        const {assets} = j;
-        setNfts(assets);
-        // console.log('got assets', assets);
+        if (loginFrom === 'metamask') {
+          const res = await fetch(`https://api.opensea.io/api/v1/assets?owner=${address}&limit=${50}`, {
+            headers: {
+              'X-API-KEY': '6a7ceb45f3c44c84be65779ad2907046',
```
avaer, line 23: Don't put API keys deep in code like this.

src/tabs/tokens.js
```diff
@@ -8,51 +8,89 @@ import metaversefile from '../../metaversefile-api.js';
...
+        }
       })();
     }
-  }, [address, nfts]);
+  }, [address, nfts, loginFrom]);
 
+  useEffect(() => {
+    if (nfts) {
+      for (const nft of nfts) {
+        if (!nftPreviews[nft.image_preview_url]) {
```
avaer, line 47: Can you explain how this routes to totum to the local iframe or some remote server to generate the preview? What is the flow like?

avaer (changes requested):
Needs clarification.

src/tabs/tokens.js
avaer: (yes, it was there before, but it should not have been)

avaer (approved):

PR #2170, opened 2022-01-06
Add title card to treehouse.scn

avaer:


avaer:
This does not currently render anything.

amnaarshed4224:
-The title card seems to work fine. It sometimes take longer than expected but i guess that is due to my connection speed. 
-Functionality wise, its fine but design wise it doesn't fits best with the app. The design needs a lot of improvement IMO.
-All other previous functionalities seems to work fine, no regressions so far.

https://user-images.githubusercontent.com/35926530/150836226-79dbf014-52bd-432e-92f9-e3231effa3d6.mp4

avaer:
We are doing this differently now, in HTML. We might bring back the shader version at some point if needed.

PR #2179, opened 2022-01-09
moon-base scene

Loryhoof:
I don't think it really fits our theme, but it might be a nice scene demo with some animated objects, flickering lights, dust etc.


avaer:
The more the merrier. I think we could actually fit it in somewhere, especially when we have dynamic loads walking along the street.

avaer (approved):

PR #2180, opened 2022-01-09
Fixed test cases and added HTTP_ONLY flag

abeersaqib:
The PR implements two important flags
- **SKIP_CERT** <--- so test cases can skip certificates and can run the test cases without certificate.

and fixes linting.

### DIFF
--- a/test/loading.js
+++ b/test/loading.js
@@ -19,20 +19,22 @@ export default () => {
     //Dispatch an event
     
     return app;
-  };`;
+};`;
+
+const PORT = process.env.PORT || 3000;
 
 async function setTestCase() {
   var cPath = path.join(__dirname, '..', 'public', 'testCase.mjs');
   await fs.writeFileSync(cPath, content);
 
-  process.env.SKIP_CERTS = true;
+  process.env.HTTP_ONLY = true;
 
   for (const scn of scenes) {
     var scenePath = path.join(__dirname, '..', 'scenes', scn);
     const data2 = fs.readFileSync(scenePath);
 
     var scene = JSON.parse(data2.toString());
-    var check = scene.objects.find(key => key.start_url === 'http://localhost:3000/testCase.mjs');
+    var check = scene.objects.find(key => key.start_url === `http://localhost:${3000}/testCase.mjs`);
 
     if (!check) {
       var data = {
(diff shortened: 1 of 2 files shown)


avaer (changes requested):
Let's not add any new flags to index.mjs.

test/loading.js
```diff
@@ -50,59 +51,64 @@ async function setTestCase() {
 
 describe('Running Pupeeteer', function() {
   describe('Loading Test Suite', function() {
-    it('Checking Scenes', async (done) => {
+    it('Checking Scenes', async done => {
       let error = false;
       const appTester = new LoadTester({
         slowMo: 0,
         host: 'http://localhost:3000',
```
avaer, line 58: Should use the actual port, which comes from the env.

test/loading.js
```diff
@@ -50,59 +51,64 @@ async function setTestCase() {
...
+      });
 
-      require('child_process').exec('git rev-parse HEAD', function(err, stdout) {
+      require('child_process').exec('git rev-parse HEAD', function(_err, stdout) {
         console.log('Last commit hash on this branch is:', stdout);
         appTester.addStatErr('HASH', stdout);
       });
 
-      await setTestCase();
+      if (process.env.CACHEGENERATOR) {
```
avaer, line 66: I don't think this is a good way to bootstrap vite cache. There must be a way to do it without booting a browser...

avaer:
This seems like 2 unrelated PRs.

test/loading.js
abeersaqib: Rectified it.

test/loading.js
abeersaqib: Removed will be added in the next PR.

test/loading.js
```diff
@@ -19,28 +19,31 @@ export default () => {
...
 
+  process.env.HTTP_ONLY = true;
+
   for (const scn of scenes) {
     var scenePath = path.join(__dirname, '..', 'scenes', scn);
     const data2 = fs.readFileSync(scenePath);
 
     var scene = JSON.parse(data2.toString());
-    var check = scene.objects.find(key => key.start_url === 'http://localhost:3000/testCase.mjs')
+    var check = scene.objects.find(key => key.start_url === `http://localhost:${3000}/testCase.mjs`);
```
avaer, line 37: Still old port.

PR #2187, opened 2022-01-10
Add player coordinates to debug stats (H key)

Loryhoof:
image

### DIFF
--- a/stats.js
+++ b/stats.js
@@ -3,6 +3,8 @@
  * @author jetienne / http://jetienne.com/
  */
 
+import metaversefile from 'metaversefile';
+const {useLocalPlayer} = metaversefile;
 
 export var Stats = function () {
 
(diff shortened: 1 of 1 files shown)


amnaarshed4224:
works fine

.

stats.js
```diff
@@ -41,15 +44,22 @@ export var Stats = function () {
...
 			frames++;
-		
-	        if (Date.now() > lastTime + 1000) {
 
+			var i = 0;
 
-				var i	= 0;
-				msTexts[i++].textContent = "FPS: "	+ Math.round((frames * 1000) / (Date.now() - lastTime));
+			// Update every frame
+			if(localPlayer) {
```
avaer, line 51: Is this update called even when the stats panel is closed? This has allocations and isn't free so I don't want to take away CPU from the app by default.

avaer (commented):
One question, but it looks good + good idea.

Loryhoof:
d
Only updating when panel is open

Loryhoof:
https://github.com/webaverse/app/blob/b5d9a8f569e2b51341268e0cfe02869d3b772dc7/webaverse.js#L294

avaer (approved):

Issue #2169, opened 2022-01-06
Asynchronous shader compile


avaer:
New THREE.js PR allows for the renderer to compile materials async: https://github.com/mrdoob/three.js/pull/19752/files

This should allow for more hitch-less loading of assets. It can be used by essentially all of the totum type loaders.

It would be especially helpful for the dioramas compile case, since we do not want to add to the scene load times, but we also want the dioramas for new characters to appear without compile lag.

avaer:
Related: https://github.com/webaverse/app/issues/2157

These two would probably fix most hitching issues, increasing the _perceived_ responsiveness and load time (though the load time will probably be the same, you'd be able to interact during it).

PR #2201, opened 2022-01-12
Asynchronous diorama compile

avaer:
This moves the shader/material compilation for dioramas to the new `renderer.compileAsync` mechanism, removing the hitch when you first open the diorama.

Fixes https://github.com/webaverse/app/issues/2169


avaer:
This also needs to handle the case where the scene fog changes, since that results in a material recompile.

PR #2217, opened 2022-01-17
Eye target update

avaer:
This PR fixes the eyes constantly following the eye target even after targeting is over.

The glitch was seen when animating and teleporting NPCs.


avaer:
This PR fixes the slerp and adds an `if` for the common case of no eye target, for efficiency.

PR #2235, opened 2022-01-20
Particle Integration

Noobulater:
Creates a particle system as a webaverse app

Allows us to add more effects using editor 

>temp link https://demo.mythicarchive.net?fx=1

### DIFF
--- a/world.js
+++ b/world.js
@@ -18,6 +18,20 @@ import {worldMapName, appsMapName, playersMapName} from './constants.js';
 import {playersManager} from './players-manager.js';
 import * as metaverseModules from './metaverse-modules.js';
 
+setTimeout(()=>{
+  const effectApp = metaversefileApi.createApp();
+  (async () => {
+    await metaverseModules.waitForLoad();
+    const {modules} = metaversefileApi.useDefaultModules();
+    const m = modules['specialEffect'];
+    await effectApp.addModule(m);
+
+    effectApp.position.set(-2,2,-2)
+    effectApp.updateMatrixWorld();
+  })();
+
+  scene.add(effectApp);
+}, 10000)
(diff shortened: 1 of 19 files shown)


world.js
```diff
@@ -18,6 +18,20 @@ import {worldMapName, appsMapName, playersMapName} from './constants.js';
...
+    await metaverseModules.waitForLoad();
+    const {modules} = metaversefileApi.useDefaultModules();
+    const m = modules['specialEffect'];
+    await effectApp.addModule(m);
+
+    effectApp.position.set(-2,2,-2)
+    effectApp.updateMatrixWorld();
+  })();
+
+  scene.add(effectApp);
```
avaer, line 33: We generally should not add things to the scene unless we are an app. And this is not an app, this is the world bootstrap.

world.js
```diff
@@ -18,6 +18,20 @@ import {worldMapName, appsMapName, playersMapName} from './constants.js';
 import {playersManager} from './players-manager.js';
 import * as metaverseModules from './metaverse-modules.js';
 
+setTimeout(()=>{
```
avaer, line 21: We shouldn't run this app by default right? We can preload the module but then not instantiate it. I imagine the effects system (accessed via totum APIs) would intantiate instances of the effects app as needed and coordinate.

world.js
Noobulater: Yea shouldn't be run by default. I'll adjust the PR, I was eager to show something working, but these changes are inappropriate for master

world.js
Noobulater: Correct, its just bootstrap.

avaer:
@Noobulater is not available to finish the work.

PR #2241, opened 2022-01-22
Depth render cleanup

avaer:
Cleans up ininity/"camera far" treatment of the depth buffer render pass.

Instead of hardcoding infinity at 30K, we now treat infinity as camera far + 1 for the purposes of depth.


avaer:
This cleans up some post processing effects that depend on infinite depth.

PR #2257, opened 2022-01-23
Sword Aim Fix WIP

yaserOSource:
This is still quite incomplete, but the start for the fix ( which I will not spend too much time on. ) 

I wanted to stop the interpolants from the legs with the sword_idle_side animation coming into effect so I just do a simple return here.

I've also deleted a bunch of commented out code, which will not be in the PR, I will reset everything back to the way it was.

https://user-images.githubusercontent.com/51108458/150660903-a18092b8-2d79-4ef1-95c0-af470d1175bd.mp4

Video shows no combat leg posture animating along with the walk, but the walk looks funny, this is not complete yet.


yaserOSource:
https://user-images.githubusercontent.com/51108458/150660947-19dc203b-6d9d-4e9a-94c0-4815bd006fe1.mp4

I also removed the jump interpolants as a test, and currently has the last frame of the run and looks like a cool mid-air jump. Not saying this will be my jump solution but I liked the effect.

avatars/avatars.js
```diff
@@ -1749,6 +1426,7 @@ class Avatar {
 
       if (this.velocity.length() > maxIdleVelocity) {
         this.lastMoveTime = now;
+        this.move = true;
```
avaer, line 1429: If this is a new avatars property, it should be in the constructor, and documented there.

yaserOSource:
https://user-images.githubusercontent.com/51108458/151267406-eeddde99-0b8e-4f6b-a945-3421175a0088.mp4

PR #2264, opened 2022-01-24
Editor:transform controller

yszhao91:
this first version use glb model to build transfrom controller
W，E，R，switch operate mode 
W - translate
E - rotate
R - scale

### DIFF
--- a/TransformControls.js
+++ b/TransformControls.js
@@ -1,12 +1,7 @@
-/**
- * @author arodic / https://github.com/arodic
- */
-
 import {
-	BoxBufferGeometry,
+	BoxGeometry,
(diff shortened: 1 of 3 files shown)


avaer:
Can we take over `WER`? Would the system be modal and you release control of the character?

yszhao91:
> Can we take over `WER` shortcuts like this? Would the system be modal and you release control of the character?

yes，call  transformControls.controller,.setMode(mode) function

After entering the editing mode, click the model to display the controller. If you click the unselected model, the controller will be hidden. Exiting edit mode also hides the controller

avaer:
Closing because nobody is driving this to completion.

PR #2270, opened 2022-01-25
Mirror API

avaer:
This adds mirror registration API in the metaversefile, which is used for avatar focus features.


player-avatar-binding.js
```diff
@@ -157,6 +164,30 @@ export function applyPlayerActionsToAvatar(player, rig) {
   // pose
   rig.poseAnimation = poseAction?.animation || null;
 }
+export function applyPlayerMirrorsToAvatar(player, rig, mirrors) {
```
yaserOSource, line 167: re mirrors sort: In the future would it be worth adding a collider to certain interactive objects and on collision we can say its the active Mirror or closest mirror?

avatars/avatars.js
```diff
@@ -2652,6 +2658,218 @@ class Avatar {
       }
     };
     _updateEyeTarget();
+
+    const _updateEyeballTarget = () => {
+      const leftEye = this.modelBoneOutputs['Eye_L'];
```
yaserOSource, line 2663: For this feature, getting the bones out of modelBoneOutputs is the main connection to avatars.js, would it be worth splitting this out?  and having a function that returns the bones to anything that requires it,  and that external object can handle the eye target updates for mirrors

avaer: > For this feature, getting the bones out of modelBoneOutputs is the main connection to avatars.js, would it be worth splitting this out? and having a function that returns the bones to anything that requires it, and that external object can handle the eye target updates for mirrors

Yes, I agree with that. I actually try to follow this design pattern with `Player`, which has e.g. sfx, physics, and the avatar separated out.

However, I think this issue is not isolated to the eyeball handling. Pretty much the entire `avatars.js` update loop can be broken out like this. The codebase is intentionally architected with strong logic isolation to facilitate these break-outs any time, so that "cleaning up" the code is mostly just moving the file it resides in. That is the point `_updateEyeballTarget`.

IMO we should clean up all of `avatars.js`, in a separate PR.

player-avatar-binding.js
avaer: We could indeed do it by distance to a collider.

But it might be more robust for the API to talk in terms of Object3D. Any user of this API can convert physics to an Object3D. Asking a client to generate physics just to get a mirror sounds like a worse experience.

avaer:
This looks so much better than the current mouselook that I think we might as well make the camera a mirror and use this system.

PR #2271, opened 2022-01-25
add `marchingCubes()`

koreex:
Added codes marching cubes based on the workflow from gonnavis


physx.js
```diff
@@ -1273,13 +1273,53 @@ const physxWorker = (() => {
...
+    scaleTypedArray.set(scale);
+
+    const outputBuffer = moduleInstance._doMarchingCubes(
+      dimsTypedArray.byteOffset,
+      potentialTypedArray.byteOffset,
+      shiftTypedArray.byteOffset,
+      scaleTypedArray.byteOffset
+    );
+
+    allocator.freeAll();
```
avaer, line 1299: You freed all memory so it is illegal to use it below.

physx.js
```diff
@@ -1273,13 +1273,53 @@ const physxWorker = (() => {
...
+
+    const head = outputBuffer / 4;
+
+    const positionCount = moduleInstance.HEAP32.subarray(head, head + 1)[0];
```
avaer, line 1303: `.subarray` aliases. If the user does not copy this, it will be corrupted next call. Should we copy?

physics-manager.js
```diff
@@ -253,6 +253,8 @@ physicsManager.simulatePhysics = timeDiff => {
   }
 };
 
+physicsManager.marchingCubes = (dims, potential, shift, scale) => physx.physxWorker.marchingCubes(dims, potential, shift, scale);
```
avaer, line 256: I think this needs a new namespace. It may be in the same binary but it is logically different from `physics`.

Should we make a new `geometry-utils.js` for this?

avaer (changes requested):
Needs some cleanup.

physx.js
koreex: `Allocator` frees memory allocated by itself. So we don't use memories allocated by `Allocator` below

physx.js
koreex: If `subarray()` is aliases, I think we need to copy to `Array`, not `TypedArray`

physics-manager.js
koreex: I'm wondering if I'm approved to make a new namespace. I think it should come from the design side of the project

physx.js
avaer: Why not copy to `TypedArray`?

physx.js
avaer: That makes sense.

physx.js
```diff
@@ -1273,13 +1273,55 @@ const physxWorker = (() => {
...
+    const faceCount = moduleInstance.HEAP32.subarray(head + 1, head + 2)[0];
+    const positions = [...moduleInstance.HEAPF32.subarray(head + 2, head + 2 + positionCount)];
+    const faces = [...moduleInstance.HEAP32.subarray(head + 2 + positionCount, head + 2 + positionCount +faceCount)];
```
avaer, line 1306: Why construct an inefficient splat array? Every caller is expecting a typed array.

scenes/grid.scn
```diff
@@ -63,6 +63,14 @@
...
       "start_url": "https://webaverse.github.io/grid/grid.glb",
       "dynamic": false
+    },
+    {
+      "position": [
+        -16,
+        0,
+        -40
+      ],
+      "start_url": "http://localhost:8000/marching-cube-example.git/"
```
avaer, line 73: We shouldn't include local files in the scenes...

physx.js
avaer: You could literally just call `.slice` instead.

physics-manager.js
avaer: Yes we can make a new namespace. I'll do it...

PR #2273, opened 2022-01-25
Vite-cache-issue

abeersaqib:
Added pre-compile approach to vite that lets vite to precompile all of the dependencies and then the new dependencies compilation takes very less toll on the server to present freshly compiled assets.


avaer (commented):
File removes/moves should be in a separate PR. Even though they are needed here, they are cleanups that can be done regardless.

If/when we revert this it should be as clean as possible, and the chance of reversion is high since it is a big technical change.

vite.config.js
```diff
@@ -8,6 +8,12 @@ export default defineConfig({
     metaversefilePlugin(),
     reactRefresh(),
   ],
+
+  optimizeDeps:{
+    entries: ['src/**/*.js', 'src/**/*.jsx','node_modules/**/*.js', '*.js', './**/*.jsx'],
```
avaer, line 13: Can we whilelist the dirs instead of using a megaglob `**`, to make sure we are not captuing e.g. locally checked out dev apps and files?

PR #2281, opened 2022-01-26
added vite compileable path

abeersaqib:
* PR implements the Vite pre-compile setup by introducing include/exclude paths
* Adds the option of force to server so cache chunks are not used.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -8,6 +8,17 @@ export default defineConfig({
     metaversefilePlugin(),
     reactRefresh(),
   ],
+  optimizeDeps:{
+    entries: [
+      'src/*.js',
+      'src/*.jsx',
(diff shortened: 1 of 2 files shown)


vite.config.js
```diff
@@ -8,6 +8,16 @@ export default defineConfig({
...
     reactRefresh(),
   ],
+  optimizeDeps:{
+    entries: [
+      'src/*.js',
+      'src/*.jsx',
+      'src/components/*.js', 
+      'src/components/*.jsx', 
+      'src/tabs/*.jsx',
+      '*.js'
```
avaer, line 18: What about `/avatars`? What are the consequences of not having that?

avaer (approved):

PR #2290, opened 2022-01-28
stop avatar aim legs influence

yaserOSource:
This is to fix the issue with aim walk being janky, now aim animation does not influence legs

https://user-images.githubusercontent.com/51108458/151588835-8b7db367-a2e1-4a49-b61e-121bb076bbf4.mp4

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -830,7 +830,7 @@ class Nodder {
 class Looker {
   constructor(avatar) {
     this.avatar = avatar;
-
+    this.move = false;
     this.mode = 'ready';
     this.startTarget = new THREE.Vector3();
     this.endTarget = new THREE.Vector3();
@@ -2025,6 +2025,9 @@ class Avatar {
 
       if (this.velocity.length() > maxIdleVelocity) {
         this.lastMoveTime = now;
+        this.move = true;
+      } else {
+        this.move = false;
       }
     };
     _updatePosition();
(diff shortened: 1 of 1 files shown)


avaer (changes requested):
The code looks about right but the animation does not.

Seem the avatar's focus is not correct? Face should be forward when aiming regardless of walk direction. The inaccuracy especially prominent on the left/right aim walk.

It could be a problem with the animation itself and we might need a new one.

avaer:
How well does this handle the couch case?

yaserOSource:
Looks like a frog during crouch, same as https://app.webaverse.com url

yaserOSource:
https://user-images.githubusercontent.com/51108458/151632701-f01baec0-8a80-4a65-a873-a4fd61a6a452.mp4

The face might be an animation change but should it not also just work the same as before, since the head should be controlled separately? or is it decided by hips forward?

avaer:
> Looks like a frog during crouch, same as https://app.webaverse.com url


Can't we use the same technique to keep  the legs stable when crouching?

avaer:
> The face might be an animation change but should it not also just work the same as before, since the head should be controlled separately? or is it decided by hips forward?

I think what is happening here (you are the expert on this code!) is the sword aim animation has a gradual torso twist and that is not being compensated for.

yaserOSource:
> > Looks like a frog during crouch, same as https://app.webaverse.com url
> 
> Can't we use the same technique to keep the legs stable when crouching?

I dont think I can do the same for the legs because there is no crouch locomotion.




> > The face might be an animation change but should it not also just work the same as before, since the head should be controlled separately? or is it decided by hips forward?
> 
> As far as I know the hips define the head, and identity twists along the route will result in the head being straight, which is how the rest of the walking works.
> 
> I think what is happening here (you are the expert on this code!) is the sword aim animation has a gradual torso twist and that is not being compensated for.

I will look at preventing the neck from updating.

yaserOSource:
https://user-images.githubusercontent.com/51108458/151640384-706e084c-ea2b-425c-867d-88df0b0e6129.mp4

avatars/avatars.js
```diff
@@ -2577,27 +2479,47 @@ class Avatar {
             })();
             if (!isPosition) {
               if (useAnimation) {
-                const src2 = useAnimation.interpolants[k];
+                const idleAnimation = _getIdleAnimation('walk');
+                
+                let src2 = useAnimation.interpolants[k];
+                
+                if(this.move && k.includes("Leg") || this.move && k.includes("Foot") || this.move && k.includes("Toe") || this.move  && k.includes("Hips"))
```
avaer, line 2486: This is a weird duplicated condition.

avaer: It could be made simpler with better layering of the statements.

yaserOSource: will update this to cleaner code

yaserOSource: Updated

PR #2298, opened 2022-01-29
Add A* path-finding `PathFinder`.

gonnavis:
Added A* path-finding feature.
Still a lot of todo, such as refactor, use plane instead of capsule to perform rise collide detection, let `generateVoxelMap()` and A* algorithm can handle ramp/cliff situations, bind with character/pet's CCT, etc.
But the core logic/workflow works now. Made draft pr first, keep working on it.

scene completely loaded:
image
rise voxels:
image
generate voxel map:
image
run A* until found destination:
image

Some voxels, such as these under stones, will not `rise`, may need using `fall` or other approaches to check collision detection too.
image
image

2d standalone version: https://raw.githack.com/gonnavis/app/8c19f95efa754b32dd82453ad6b7ecc375403295/a_star_standalone/index.html

---UPDATE---
~~Sibling PR: https://github.com/webaverse/totum/pull/59~~
~~https://github.com/webaverse/app/pull/2298#issuecomment-1031682713~~
https://github.com/webaverse/app/pull/2298#issuecomment-1034508782


gonnavis:
Handled ramp/cliff situations.
image
image

avaer:
Thanks! Small conflict resolution needed.

gonnavis:
Hello @avaer , where do you think the code for the A* algorithm is better?
`app/webaverse.js`, `totum\type_templates\glb.js`, `metaverse-utils`, or somewhere else?

I placed it in `app/webaverse.js` just for fast testing, so made draft pr before.

gonnavis:
Supported two layers of terrain/scene now.
Should be able to extend to support multi-layers.

QQ截图20220202043259
QQ截图20220202043231

webaverse.js
```diff
@@ -41,6 +41,357 @@ import WebaWallet from './src/components/wallet.js';
 // const leftHandOffset = new THREE.Vector3(0.2, -0.2, -0.4);
 // const rightHandOffset = new THREE.Vector3(-0.2, -0.2, -0.4);
 
+window.isStart = false;
```
avaer, line 44: We should not hack `window` like this. It's ok for testing but we should clean it up before merging.

avaer:
> Solved. But , where do you think the code for the A* algorithm is better placed?

Maybe a new file called `npc-utils.js`? It doesn't really fit with the existing things.

gonnavis:
> Maybe a new file called `npc-utils.js`? It doesn't really fit with the existing things.

Ok, I see. Thx!

webaverse.js
gonnavis: Ok, I'll refactor.

gonnavis:
Refactored into `npc-utils.js`'s `class PathFinder`.
Although there are still many places that can be optimized, such as merge result path points and support multi-layers, but may worth as a first step.

Usage 1: All automatically:
```
const pathFinder = new PathFinder({debugRender: false,
  width: 71, height: 71, voxelHeight: 2, lowestY: -0.1, highestY: 15, highestY2: 30});

await pathFinder.init();
// => "PathFinder auto inited."
```

Usage 2: Manually:
```
const pathFinder = new PathFinder({debugRender: true,
  width: 71, height: 71, voxelHeight: 2, lowestY: -0.1, highestY: 15, highestY2: 30});

pathFinder.rise();
pathFinder.rise2();
pathFinder.generateVoxelMap()
```

Sibling PR: https://github.com/webaverse/npc/pull/2

npc-utils.js
```diff
@@ -0,0 +1,528 @@
...
+
+  generateVoxelMap() {
+    this.isRising = false;
+    this.isRising2 = false;
+
+    for (let z = -(this.height - 1) / 2; z < this.height / 2; z++) {
+      for (let x = -(this.width - 1) / 2; x < this.width / 2; x++) {
+        const currentVoxel = this.getVoxel(x, z);
+
+        const leftVoxel2 = this.getVoxel2(x - 1, z);
```
avaer, line 161: This is a mess of `if` cases. Is this something that could be abstracted? Like...

```
leftVoxel2 = getVoxelDelta(-1, 0, 0);
rightVoxel2 = getVoxelDelta(1, 0, 0);
```

avaer:
IMO the whole pathfinder concept should be based around the NPC, not the scene. The NPC asks for a path to a target, and the pathfinder obeys. That should have nothing to do with the scene's configuration, and it should be available everywhere in the world with no setup step at all.

I don't think we need to do any pre-calculation for this pathfinding. PhysX should be fast enough to compute the entire voxel set every frame if needed. This can be done in one call, like `computePathfindingVoxels(Vector3 center, float size)`.

avaer:
Rough sketch of the idea:

```
// npc.js
useFrame(() => {
  const pathfinder = usePathfinder();
  
  const range = 20;
  const waypointResult= new Float32Array(Math.ceil(range * Math.SQRT2) * 3); // waypoint vectors, max possible is diagonal?
  const numWaypoints = pathfinder.getPath(localPlayer.position, new THREE.Vector3(66, 2, -108), waypointResult);
  const nextWaypoint = new THREE.Vector3().fromArray(waypointResult, 0); // the place we should ideally go this frame
  const target = new THREE.Vector3().fromArray(waypointResult, (numWaypoints - 1)*3); // the last waypoint
});
```

gonnavis:
OK, thanks! I'm turning it into real-time.

gonnavis:
Turned it into real-time, generating voxel map on demande, no need precomputing anymore.

Caching already generated voxels for speed.

Output `waypointResult`, which is the last successfully found path, to let `npcPlayer` can keep going after `localPlayer` go to somewhere that can't found path.

Handled some more edge-cases.

But still using two layers of voxel map structure, need manually set `lowestY` `highestY` `highestY2` for specific scene, not purely auto-suitable for arbitrary scene now. But tested that very robust in treehouse and some other scenes, if properly set these args.

Usage:
```
const pathFinder = new PathFinder({voxelHeight: 2, lowestY: 0.1, highestY: 15, highestY2: 30});

let target = npcPlayer; // target: Object3D
function update() {
  if (isInit || localPlayerFarawayPrevDest) {
    const start = npcPlayer.position.
    const dest = localPlayer.position.
    const isFound = pathFinder.getPath(start, dest);
    if (isFound) target = pathFinder.waypointResult[0];
 } ese if (npcPlayerCloseToTarget) {
  if (target._next) {
    target = target._next;
  }
}
```

Todo:
-Don't using two layers of voxel map structure, auto-suitable for arbitrary scene.
-Dispose some oldest voxels when cached too many.
-Clean `debugRender` related codes for performance.
-`cost` changing based on height bias, instead of all `1`.
-Refactor `generateVoxelMapLeft/Right/Btm/Top()` etc into common function.

gonnavis:
Combined waypoints into lines, now output much lesser waypoints.
Thus by the way solved npc jitter back and forth and can just walk can't run problems.

gonnavis:
Should supported scenes with any number of layers, not limit to pre-defined two layers structure.

npc-utils.js
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    return this.isFound;
+  }
+
+  simplifyWaypointResultX(result) {
+    if (result?._next?._next) {
+      if (
+        result.position.x === result._next._next.position.x && // check whether in one line
+        Math.sign(result.position.z - result._next.position.z) === Math.sign(result._next.position.z - result._next._next.position.z) // check wheter in one layer // TODO: simplifyWaypointResultXZ() should need check wheter in one layer too.
```
avaer, line 149: Unfinished todo.

gonnavis: Yeah, it's a little hard and need a lot test, I'm keep working on this.
But I feel it's rare case or even will never appear due to the A* algorithm, I'll confirm.

npc-utils.js
```diff
@@ -0,0 +1,551 @@
...
+const colorPath = new THREE.Color('rgb(149,64,191)');
+const colorPathSimplified = new THREE.Color('rgb(69,0,98)');
+
+class PathFinder {
+  constructor({voxelHeight = 1.5, heightTolerance = 0.6, detectStep = 0.1, maxIterdetect = 1000, maxIterStep = 1000, maxVoxelCacheLen = 10000, debugRender = false}) {
+    /* args:
+      voxelHeight: Voxel height ( Y axis ) for collide detection, usually equal to npc's physical capsule height. X/Z axes sizes are hard-coded 1 now.
+      heightTolerance: Used to check whether currentVoxel can go above to neighbor voxels.
+      detectStep: How height every detecting step moving.
+      maxIterdetect: How many steps can one voxel detecing iterate.
```
avaer, line 27: I still don't understand `maxIterdetect`. Why is it needed? Isn't the max already known from `maxIterStep`?

npc-utils.js
```diff
@@ -0,0 +1,551 @@
...
+      maxIterStep: How many A* path-finding step can one getPath() iterate. One A* step can create up to 4 voxels, 0 ~ 4.
+      maxVoxelCacheLen: How many detected voxels can be cached.
```
avaer, line 29: Why does the cache matter? Why does the user control it? Will it change functionality to change this value?

npc-utils.js
gonnavis: Every voxel need use `overlapBox()` to recursively `detect()` to decide it's `position.y`, `maxIterdetect` is used to limit this recursion.
It's very different from `maxIterStep` which is used to limit A* algorithm's `step()` recursion.

npc-utils.js
gonnavis: Every voxel need to know whether it can go to -X/+X/-Z/+Z ( left/right/btm/top) 4 neighbor voxels, based on their `position.y`.
And every voxel's `position.y` is decided by recursively `detect()`, which is sometimes a time-consuming operation.
So I cached their `position.y`, `canLeft/Right/Btm/Top` and other calculated infos for better performance.
Then I think we can't infinitely cache them all, so used `maxVoxelCacheLen` to limit the cache size.
`maxVoxelCacheLen` will not change functionality at all, users can control it to a suitable value to just prevent occupy too many memory.

But I just noticed that after refactored the algorithm from predefined two layers to support arbitrary layers, the cache seems not reused by the algorithm it self, just affecting `debugRender`.
I'll confirm and try to:
1. Reuse the cache.
2. Delete the cache and related codes at all, if performance is ok.

gonnavis: Fixed "not using cache" bug. https://github.com/webaverse/app/pull/2298/commits/39b30e236fe3ab74654dcc0a654ed54ac231c243

And tested that if not using cache, the `detect()` will run much more times thus the performance will drop a lot.

npc-utils.js
gonnavis: Done.
https://github.com/webaverse/app/pull/2298/commits/fdc03437cce000d9e8b4a2f6c964ed8b5f753e25
https://github.com/webaverse/app/pull/2298/commits/1d728d0d42736d829a384a3e28dec995be72cfa3

avaer:
Needs merge.

gonnavis:
Ok, just notice the conflict, let me solve it.

avaer (changes requested):
Looks like this is allocating voxels indivdually, and each frame.

There is a caching solution to try to allocate less, but IMO the whole approach should be a flat array in WASM. Any JS tree/allocated voxel objects approach will scale poorly with the number of players/npcs.

We can still cache things in a C++ object if needed, but in C++ at least it will not cause GC problems when scaling to multiple players.

avaer:
Functionality wise this PR is testable, it's just not usable in production due.

Let's do it this way: If we fix the conflicts we can merge this PR and start testing, and move to WASM in another PR.

avaer:
IMO `PathFinder` could be a WASM class, and we call things on it, just like our `Scene`.

lore-ai-scene.js
```diff
@@ -0,0 +1,7 @@
+class LoreAIScene {
+  addCharacter() {
+    return new EventTarget();
```
avaer, line 3: Posterity: this is a stub that will be replaced from downstream fork.

avaer (approved):

PR #2300, opened 2022-01-30
Fixed renderer stats ['H' panel] cause by post-processing.

ohmed:


renderer.js
```diff
@@ -44,6 +44,23 @@ function bindCanvas(c) {
   // renderer.gammaFactor = 2.2;
   renderer.shadowMap.enabled = true;
   renderer.shadowMap.type = THREE.PCFSoftShadowMap;
+
+  // reload renderer WebGLInfo.reset function
```
avaer, line 48: What is being reloaded?

renderer.js
```diff
@@ -44,6 +44,23 @@ function bindCanvas(c) {
...
+  // in order to continue chaning frames value
```
avaer, line 49: What is a chaining frames value?

renderer.js
```diff
@@ -44,6 +44,23 @@ function bindCanvas(c) {
...
+  // in other case [.autoReset = false] it'll broke Avatars hair rendering [not sure why]
```
avaer, line 50: This is worrying. Why would the stats affect the rendering? This should be debugged.

avaer (changes requested):
Looks like more debugging needed, because the code is mysterious and relying on unexplained comments.

renderer.js
ohmed: Default renderer.info.reset method

renderer.js
ohmed: Changing renderer.info.frames number.
Is increased each .info.reset() call.
If not called automatically [renderer.info.autoReset = false] it'll result in renderer.info.frames not updating and will cause issues.

ohmed:
I dig a bit deeper you can get to this
image
frame is used to update skinned meshes [if frame the same, update is skipped].
and it's probably related to PostProcess or smth.
Should I continue or just contact somebody who did this part of code [rendering head etc] ?

renderer.js
avaer: That's not really "reload". Reloading the page to collect stats is something very different, so this comment is confusing.

renderer.js
avaer: Where is the chain?

avaer:
☝️ 
Links are more useful than screenshots. I found it by searching myself, but that's pure luck: https://github.com/mrdoob/three.js/blob/fadb6ab1cfcf917e8287e2cca6292aae6ef15793/src/renderers/WebGLRenderer.js

avaer:
I don't think this is the right direction, since the research was not done and this code causes mysterious problems.

We can re-open when we have a better understanding of the code and what to fix.

renderer.js
avaer: > If not called automatically [renderer.info.autoReset = false] it'll result in renderer.info.frames not updating and will cause issues.

What issues?

PR #2305, opened 2022-01-31
new Jump animations WIP

yaserOSource:
Adding manual jump logic with jump, fall and land animations


yaserOSource:
https://github.com/yaserOSource/app/tree/jump-action

I also wanted to use actions to get the correct states for jump / fall / land, so there is some progress with that on this branch.

PR #2309, opened 2022-02-01
Ragdoll support [WIP]

avaer:
Depends on https://github.com/webaverse/app-wasm/pull/23

[image](https://user-images.githubusercontent.com/6926057/152674352-aff3964b-c77c-449e-b0f5-52f72bd65fb2.mp4)

Press `H` to see the bones. Then, press `B` to toggle ragdoll.

## Motivation

This is for supporting dynamic character deaths.

We could do static death animations, but those look so bad that I didn't even want to bother tuning it when we can do actual ragdolls with Physx.

## What's done

- Create a parallel physical bone mesh to the avatar's skinning mesh. This is the visualization of what PhysX sees, using capsule bones. This is what you enable when you press `H`. The bones also have a "tail" so you can visualize the direction that it is facing. You can think of it like a plane with a tail. The plane faces toward `(0, 0, -1)`.

image
image

- Bind the ragdoll bones to the animated skin bones.
- Copy the skeleton to PhysX with a serialization procedure
- Hook up the joints simulation in PhysX
- Support anti-self-collision based on "collision groups", so the skeleton of the avatar does not collide with the capsule.
- Copy the PhysX simulation results back to the bones visualization.
Note: this copy is _not_ exact -- we only copy the quaternion, so the simulation and the actual ragdoll can be out of sync a bit, when the bones are violating constraints and being torn apart. But it makes our avatars less warped and prone to noise coming in from the matrix feedback loop, which can be magnified to give stretches and other mesh corruptions as the errors accumulate.
- Try to rotate the bones perfectly in one direction (from toes to head) to optimize skeleton simulation stability.

This PR also fixes some bones binding bugs (e.g. incorrect bones mappings for Toes), because they were getting in the way of testing this PR.

## Research

Much of the logic/math comes inspired from https://github.com/diharaw/physx-ragdoll
PhysX geometry document https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/Geometry.html
PhysX joints document https://docs.nvidia.com/gameworks/content/gameworkslibrary/physx/guide/Manual/Joints.html

### Bug 1

The avatar will often immediately be pulled backwards and twitch when it is ragdolled, meaning either:

1) the math on the relative joints setup is wrong, and the constraints are being violated immediately and the avatar must immediately do a violent correction
2) the joint bindings math is so wrong that it is contradicting itself, causing the bones to pull on each other with increasing ways
3) our constants are bad in terms of timestep fidelity, materials restitutions, iteration counts, or some other details of the solver, which can be updated with physx parameters
4) maybe we need to restructure our joints binding to not be a "bone worm", but instead just do it more arbitrarily, like https://github.com/diharaw/physx-ragdoll/blob/32c0e5634889ad88726fbfd7a804cbf91010b9bc/src/main.cpp#L512

### Bug 2

We do not have correct human constraints on the joints. Right now it is just a simple limit that you do not deviate too much from the initial pose, but that is not how human joints work. We might want to copy some constants from here: https://github.com/diharaw/physx-ragdoll/blob/32c0e5634889ad88726fbfd7a804cbf91010b9bc/src/main.cpp#L649. But this is just something that we should tweak to look good with our avatars.

### Bug 3

We do not compute the parallel bone skeleton very efficiently, but it is a common copy operation (potentially per-avatar, and per-frame). We should calculate the space we need for the skeleton buffer, allocate that memory from the physx WASM using `Module._malloc`, and then write the data there directly.

### Bug 4

We are not garbage collecting/destroying skeletons once they are done simulating (e.g. the VRM object is destroyed).

### Bug 5

Ragdoll physx bones lookup is only handled for the local player, and is not yet a good API for accessing player bones. That is why there is a hack in `getAppByPhysicsId`: https://github.com/webaverse/app/pull/2309/files#diff-67940e2695eb5cc95d3e4f085f352d728d210cdf367aa25c36ba980feff83120R852 and `getPhysicsObjectByPhysicsId`: https://github.com/webaverse/app/pull/2309/files#diff-67940e2695eb5cc95d3e4f085f352d728d210cdf367aa25c36ba980feff83120R880. We should get rid of this hack and cleanly handle the physics code paths for all avatars by going through e.g. the world `AppManager` and remote player `AppManager`, which has the VRM apps which have these `ragDollMesh`.

## Todo

- Fix the bugs
- Fix a memory inefficiency, where we constantly create new buffers for the skeleton each frame: https://github.com/webaverse/app/pull/2309/files#diff-fcb2093ac11d9b04a179d3dedbef5c0aae4d14b7f71db90e483becca4170b7ccR990
- Tweak the bone shapes and constants to make it fit the avatar, instead of just their spine
- Add velocity support to the skeleton so we can make the avatar fly when the get hit

This ragdoll test mode can be shipped before we start hook it up to user apps in combat.


avaer:
I don't know if this is important but this is what this code is about: https://github.com/webaverse/app/pull/2309/files#diff-fcb2093ac11d9b04a179d3dedbef5c0aae4d14b7f71db90e483becca4170b7ccR876

image
image

PR #2312, opened 2022-02-02
H panel inaccurate info fix

Sirahi:


avaer (approved):

avaer:
Reverted because it caused avatars decapitation corruption.

PR #2315, opened 2022-02-02
[WIP] New Compile Architecture

abeersaqib:
* [x] Achieve the plug out of totum
* [x] App should be compilable after totum plug out
* [x] App should be loadable from parcer, serve or any other server and should open fine


avaer:
This PR changes 617 files and it's not clear what it's doing, with no forward progress.

Closing this because it's not clear what this is or if it will ever merge.

PR #2316, opened 2022-02-03
Jump logic with new jump animations WIP

yaserOSource:
Jump on the spot if no x,z movement, otherwise jump forward in the direction of movement

Go into fall mode after 1000 jumptime

After nearing ground, play land animation, vary between hard and soft landing animation based on jumptime ( WIP)

https://user-images.githubusercontent.com/51108458/152265657-69ab570c-8dc9-419d-9cf2-e8b265666165.mp4


yaserOSource:
https://user-images.githubusercontent.com/51108458/152266806-c19e88f4-dec3-4ca5-ad4e-3b0e2d1ecd11.mp4

yaserOSource:
https://user-images.githubusercontent.com/51108458/152268665-5fa81300-8978-4379-81d0-157f987766c4.mp4

yaserOSource:
https://user-images.githubusercontent.com/51108458/152282592-387b4334-9833-41b6-914c-5b9c198c5f77.mp4

yaserOSource:
https://user-images.githubusercontent.com/51108458/152293904-13b18aa5-d771-4f6b-8848-cb7dd3f0deb5.mp4

hard landing cleanup

yaserOSource:
https://user-images.githubusercontent.com/51108458/152294065-d4f378e3-8b69-4190-b2f1-6442262d9b09.mp4

soft landing test

yaserOSource:
https://user-images.githubusercontent.com/51108458/152464299-765df012-809a-458d-9ab1-5a462518fe58.mp4

Soft landing update with default horizontal blend on the legs, and influence from hipPos

avaer:
It's getting there 👀

One suggestion I would make is a 50% chance of flipping the jump animation, so both feet are used for variety. The brain stops buying it if it's the same foot.

Flipping can be accomplished in the bake, as we already do for reversals.

yaserOSource:
> It's getting there 👀
> 
> One suggestion I would make is a 50% chance of flipping the jump animation left/right, so both feet are used for variety. The brain stops buying it if it's the same foot.
> 
> Flipping can be accomplished in the bake, as we already do for reversals.
thanks

ive got an animation with the other foot jumping for that reason too, ill be putting that in for sure

yaserOSource:
I will use the reversal in the animation baker

yaserOSource:
Do you know why mirrored mixamo animations dont work? 

Seems to look fine in blender but doesnt play in the app

yaserOSource:
https://user-images.githubusercontent.com/51108458/152497474-565580f5-1256-42c5-84bd-23d6da67574c.mp4

Removed the hip influence from the soft landing as it caused the feet to fall through the ground but the overall animation is not as good.

yaserOSource:
https://user-images.githubusercontent.com/51108458/152501287-81ab0c2e-295c-4310-b9c0-64452dc295f3.mp4

with hips influence from landing animation, as you can see its more bouncy and realistic but feet go through floor,i will fix this another way.

yaserOSource:
https://user-images.githubusercontent.com/51108458/153054318-c176700f-7750-456c-b120-c37323cdd502.mp4

jump forward kicks with left and right foot and switches, better timing with hard landing and soft landing no longer goes through floor

yaserOSource:
-hard landing roll to stand if they continue to move
-soft landing from standing jump
-soft landing from jump forward

yaserOSource:
hard landing to roll (if there is additional movement) or just land on hands and feet

https://user-images.githubusercontent.com/51108458/153191860-90e7f792-2836-426c-8fb1-f5fa848c9123.mp4


https://user-images.githubusercontent.com/51108458/153191839-601b40fd-fb15-48ec-a28a-7086ca5652fc.mp4

yaserOSource:
- landing while pressing aim needs a solution, so directional landing can occur

yaserOSource:
- Case where there is a roll and also some airtime after the roll and it defaults into a jump forward, needs to be fixed

amnaarshed4224:
The new animation do look really good. These few thing i wanted to give my feedback on:
1-The bump at the end of jump animation seems a bit different and separate as compared to the whole other jump animation. Also this bump is same at all kind of terrains, like on smooth surface, down the slope, upward slope.. that seems a bit odd. Also i guess the transition from when avatar's upper half goes forward at end of jump, to normal position is fast and jerky, maybe it is making it seem odd.

https://user-images.githubusercontent.com/35926530/153488049-afa92cbd-5f6c-48c1-ba9f-da2adb3bede4.mp4

2-While walking and running, if we jump, the bump at the end happens even if we keep running. That seems a bit odd too.

https://user-images.githubusercontent.com/35926530/153486839-e87d7d6e-8ffa-4655-b966-d670378415b0.mp4

3-The rolling animation is really good but on uneven surfaces the avatar sometimes doesn't completes the animation. That doesn't seems right.

https://user-images.githubusercontent.com/35926530/153487315-b600b1e9-3e90-495f-a2e4-634982b703dc.mp4

scenes/treehouse.scn
```diff
@@ -63,116 +63,6 @@
       ],
       "start_url": "https://webaverse.github.io/skybox-anime/"
     },
-    {
```
avaer, line 66: Probably shouldn't have this deletion in here.

avaer (changes requested):
This code is exacerbating the lack of factoring in `avatars.js` by confusing states with animations.

Basically, the avatars animation update loop is confounding the concept of collecting environmental data into a state engine, and actually rendering out that state engine (i.e. the actual bone kinematics). These are basically unrelated concepts but they are interwoven into the `avatars.js` code and controlled by `this.variables`, which is basically turning it into a straightforward, but global variable spaghetti code. That really is not going to work when we try to do things like multi-state animations and rolls which get into deeper logic and `if` statements across multiple frames. It makes the code really hard to get right.

IMO the next step in this PR is to come up with inter-lerpable animation chains that execute in a well-defined state update + render pipeline, thus clarifying the distinction between state and output.

Then, a jump could be naturally encoded as detectable state transitions from grounded to air to roll, which once set are smoothly interpolated out during the bones render pass.

avaer:
The main visual problem with the current code is the lack of lerps between the states. But when each state is its own separate class in a chain, with evaluation functions, then lerping between the action classes should be very straightforward. Just evaluate both bones (quaternions) and slerp.

yaserOSource:
> The main visual problem with the current code is the lack of lerps between the states. But when each state is its own separate class in a chain, with evaluation functions, then lerping between the action classes should be very straightforward. Just evaluate both bones (quaternions) and slerp.

Yes exactly there is no lerping between the jump stages and various outcomes ( soft / hard/ movement after land )

avaer:
> Yes exactly there is no lerping between the jump stages and various outcomes ( soft / hard/ movement after land )

But I bet the only reason this isn't already in the current code is the code was too spaghetti to add a `.slerp`.

avaer:
> I recognised that jumpState and others like it were updated outside of avatars.js and thought i could have conditions within it, as a way of controlling the flow of the animation ( what to p[lay next )

Yeah, but that is being entangled with the rendering code, because each action sometimes maintains its own state, and depends on other actions, etc.

There should be a clear separation between all of this state.

avaer:
Here is a rough sketch of the above ideas.

image

yaserOSource:
> Here is a rough sketch of the above ideas.
> 
> image

So i need to remove these inner conditions and actually have them updated elsewhere similar to jumpState

ahadshams:
I dont know what that comment means.. afaik @plankatron has been available for any tweaks / modifications required

avaer:
This PR is no longer being worked on.

0reo:
Hi @avaer, @ahadshams  asked me to work on an animation system and sent me this PR to look at.

I was wondering if there's any specific information regarding this that I should be aware of(outside of the notes/diagrams here), and if there are any specific issues currently relevant to this(either jumping specifically, or the animation system as a whole).  I can just dump all the issues referencing animation into a milestone and start reading, but it'd be good to know if there are any in particular that I should look at/ignore.  I'm also going to be talking with textametron, so if he's got all the relevant info then I'll await his response.

avaer:
IMO there is code + data (animations) worth keeping here.

Ideally this PR wouldn't be closed, but we can fork to a branch in this repo.

PR #2332, opened 2022-02-08
[WIP] Implementing new react game-play UI.

ohmed:
New UI is still build using React, but more clear and with improved structure.
Creating new components & porting some old components.

- Making wireframe UI and logic.
- New structure with some clean up.


avaer:
Gonna close this PR since it's gotten too big.

To keep in spirit with our iterative process, we should be shipping parts of the UI every 2 days tops; it's basic React.

This has been in wip for over a week and balooned into something with no end in sight. Additionally, part of the qualifications of being able to ship PR is that is done fast enough to demonstrate our ability to support it in production, which this PR isn't meeting.

It's cool to use this code in future PRs, but we will not be able to integrate PRs which take this long.

PR #2335, opened 2022-02-08
Hpanel accurate info with draw calls

Sirahi:


avaer:
From the author, since the PR is missing a description:

> I figured out the issue. I was calling info reset in the stats which only gets called when h panel is open since you added the check game.debugMode in render loop. So when h panel was off, not resetting info was causing rendering issues

avaer (approved):
One more try at fixing the bug. Thanks @Sirahi!

PR #2338, opened 2022-02-09
WIP: Add geometry utils and function to generate basic chunk geometry

koreex:


avaer:
Closing because it's old code with no description.

PR #2347, opened 2022-02-10
tweaking idle.fbx arm positions and also pushing animations.z updated…

plankatron:
pushing requested changes


avaer:
Is it worth folding the arms in more?

image

PR #2348, opened 2022-02-10
Combat mode and combat animations WIP

yaserOSource:
_todo_
- Equipping sword is an animation and puts the character in loose combat stance (sword out but not actively aiming)
- Aiming with the sword should enable aiming walk/run animation, with 4 way movement (f/b/strafe)
- When not holding right-click(aim) player should otherwise run forward in any dir with the weapon out
- 4 combo animation weapon

This should apply for all weapons.

_Ideas_
- Target lock mode that also utilises aim mode


yaserOSource:
im recreating this branch and cleaning up the conflicts


its not done enough to test no, but it will be today

Issue #2343, opened 2022-02-09
app: Switching avatar styles multiple times breaks the app


alisaad673:
If we switch the currently available avatar styles multiple times( in this case 5 times), the app breaks.

https://user-images.githubusercontent.com/35926530/153225351-a59d60d6-d07f-46c2-baa6-913548c5e63f.mp4

![33](https://user-images.githubusercontent.com/35926530/153226340-86388cbf-433c-4956-bec2-a16d6772d0f4.PNG)

Issue #2341, opened 2022-02-09
app: Turning on multiple avatar styles isn't handled 


alisaad673:
Currently, if we turn on one avatar style, lets say "sprite sheet" and then we switch to the other avatar style "optimized avatar", it results in a mixed up avatar of both styles. Switching from one style to other should disregard the previous style.

https://user-images.githubusercontent.com/35926530/153217040-03717087-d3ab-40b2-9fdf-40d110ecff57.mp4

Similarly, after above steps, if we now switch back again to "sprite sheet" doesn't really switches the style, it instead throws aside the avatar with sprite sheet style and carries on with previous mixed avatar.

https://user-images.githubusercontent.com/35926530/153218762-9c81185c-73a7-4071-96dd-8dd2c008988d.mp4

Issue #2339, opened 2022-02-09
app: No option to go back to default avatar style in option panel


alisaad673:
Right now we have two of the avatar styles implemented but we do not have an option to go back to original default avatar. Only way is to reload the app.

avaer:
I think the way we will do it is instead of deleting the old avatar when we switch, we will leave the old avatar in place.

amnaarshed4224:
Or we can add another option for default avatar style as a 5th option. Can we?

avaer:
Maybe we are talking about different things. I was talking about the ability to wear an avatar in-world, and still having access to the previous avatar.

If this is about changing avatar render modes, but not changing the avatar, then I think we should just finish implementing all of the settings in the options panel. There is no need for a 5th option. The default will be auto-detected from one of the existing four options.

PR #2351, opened 2022-02-10
Performance settings

### LINKED ISSUE #2321: Persistent avatar settings
We have an avatar auto-optimization pipeline coming online, and it can be tested in the options panel. (Ignore the UI)

image

We want to make this setting persistent 

## Requirements

1. Make the avatar setting persistent across reloads (apply directly on load, without the unoptimized avatar ever being rendered). Can use `localStorage`
2. Make the setting apply to all avatars in the scene (all [totum](https://github.com/webaverse/totum/) VRM)
3. Bugfix corner cases, like the dioramas not rendering
4. Support some sort of heuristic for setting the performance mode if it was not previously set, or at least don't always use the MToon HQ mode, since that is a performance killer -- one option is we default to the crunched avatar in all cases
5. (Performance optimization) use the crunched version of the avatar for depth rendering in the post processing stack, and the diorama, to collapse draw calls into one

0reo:
~~Here's what I have so far for the avatar quality stuff.  Ran into some issues with new lines when committing that I need to finish cleaning up still via a rebase, in particular on the app-manager file.  Will update this with more info and questions shortly.~~

--

As I was learning the system, and using the existing functions to change the quality, my initial approach was to set a load property on the avatar, and to keep it from rendering until the load state was set to true.  I attempted to have the quality settings be applied in the constructor of the Avatar class, but found that the sprite creator made it's own avatar instance, and was causing an infinite loop.  so i moved setting the initial quality settings to just after creating the avatar, and the would enable the load property once the quality was set to start rendering.  Looking at the setQuality function in avatars.js, I saw the current approach was to compress the existing model, attach the new model, and then hide the old one.  But it wasn't keeping track of the sprite models, it just kept adding new ones, so I updated it to track the initial model, crunched model, and sprite, and only create new ones if there wern't any existing already.  

I soon saw that the VRM templates had an activate callback which would set up an avatar for the local player.  The app is manually calling setAvatarUrl, I felt it made sense to be able to utilise the activation callback and have the ability for the VRM template to assign itself as an avatar.  And see as the quality settings were to be applied to all VRMs, not just the player's avatar, I decided to shift the initial quality logic into the template.  I still need to add the cruncher into it though.  I then modified the scene template to identify if it was loading a VRM or not, and if the model being loaded was marked as a player by adding an extra property to a model in the scene manifest.  If the scene gets a VRM that's not the player, it make that VRM into an NPC.  I then updated the setAvatarQuality function in game.js to account for the NPC's as well.  

I also added watch to the app project to have it auto restart if I modified the templates.  It's not a hot reload but it's less annoying than having to manually restart it.

Some questions:
- making the VRM's into avatars discards their initial set position.  I can change that, but wanted to make sure I'm going in the ideal direction before making more changes to to the avatar/NPC logic.
 - the VRM templates were running the toon shader on the models twice, once in default for unskinned, and one in the setSkinning function, and so it's currently doing that for spriting as well.  I assume it should only need to be run once per VRM and reused?  
 - applying the quality changes to all the avatars takes quite a bit of time, it can surely be sped up, but I didn't want to go too far off task adding a loading screen.  Is the expectation that all the down-scaling will be done on the client side on load?  Any plans to store/reuse models online/locally?
 
I'm not 100% on the "whys" of everything yet obviously.  Getting more insight into how it currently works vs how you're intending it to work will help a lot.  Looking forward to discussing in more detail.

resolve #2321

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -257,8 +257,8 @@ class StatePlayer extends PlayerBase {
     });
     this.appManager.addEventListener('appadd', e => {
       const trackedApp = this.appManager.getTrackedApp( e.data.instanceId);
-      const load = trackedApp.get('load');
-      if(!load) return;
+      // const load = trackedApp.get('load');
+      // if(!load) return;
       const app = e.data;
       scene.add(app);
     });
--- a/character-controller.js
+++ b/character-controller.js
@@ -1106,10 +1106,10 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
     this.isNpcPlayer = true;
   }
   async setAvatarAppAsync(app) {
-    await app.setSkinning(true);
+    // await app.setSkinning(true);
     
-    const {skinnedVrm} = app;
-    const avatar = new Avatar(skinnedVrm, {
+    const {skinnedVrms} = app;
+    const avatar = new Avatar(skinnedVrms['active'], {
       fingers: true,
       hair: true,
       visemes: true,
--- a/character-controller.js
+++ b/character-controller.js
@@ -1109,7 +1109,7 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
     // await app.setSkinning(true);
     
     const {skinnedVrms} = app;
-    const avatar = new Avatar(skinnedVrms['active'], {
+    const avatar = new Avatar(skinnedVrms, {
       fingers: true,
       hair: true,
       visemes: true,
@@ -1117,7 +1117,7 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
     });
     console.log("MADE NPC AVATAR", avatar);
     const quality = metaversefile.getQualitySetting();
-    avatar.setQuality(quality);
+    avatar.updateQuality();
 
     unFrustumCull(app);
     enableShadows(app);
--- a/character-controller.js
+++ b/character-controller.js
@@ -1115,10 +1115,13 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
       visemes: true,
       debug: false,
     });
-  
+    console.log("MADE NPC AVATAR", avatar);
+    const quality = metaversefile.getQualitySetting();
+    avatar.setQuality(quality);
+
     unFrustumCull(app);
     enableShadows(app);
-  
+
     this.avatar = avatar;
 
     this.characterPhysics = new CharacterPhysics(this);
--- a/character-controller.js
+++ b/character-controller.js
@@ -256,6 +256,9 @@ class StatePlayer extends PlayerBase {
       appsMap: null,
     });
     this.appManager.addEventListener('appadd', e => {
+      const trackedApp = this.appManager.getTrackedApp( e.data.instanceId);
+      const load = trackedApp.get('load');
+      if(!load) return;
       const app = e.data;
       scene.add(app);
     });
@@ -728,10 +731,11 @@ class LocalPlayer extends UninterpolatedPlayer {
     this.characterHups = new CharacterHups(this);
     this.characterSfx = new CharacterSfx(this);
     this.characterFx = new CharacterFx(this);
+    this.loaded = false;
   }
   async setAvatarUrl(u) {
     const localAvatarEpoch = ++this.avatarEpoch;
-    const avatarApp = await this.appManager.addTrackedApp(u);
+    const avatarApp = await this.appManager.addTrackedApp(u, undefined, undefined, undefined, undefined, false);
     if (this.avatarEpoch !== localAvatarEpoch) {
       this.appManager.removeTrackedApp(avatarApp.instanceId);
       return;
@@ -744,7 +748,7 @@ class LocalPlayer extends UninterpolatedPlayer {
     this.playersArray.doc.transact(function tx() {
       const avatar = self.getAvatarState();
       const oldInstanceId = avatar.get('instanceId');
-      
+
       avatar.set('instanceId', app.instanceId);
 
       if (oldInstanceId) {
--- a/game.js
+++ b/game.js
@@ -1977,8 +1977,14 @@ const gameManager = {
     localPlayer.removeAction('activate');
   },
   setAvatarQuality(quality) {
+    const qualityMap = {
+      "ULTRA":4,
+      "HIGH":3,
+      "MEDIUM":2,
+      "LOW":1
+    }
     const localPlayer = metaversefileApi.useLocalPlayer();
-    localPlayer.avatar.setQuality(quality);
+    localPlayer.avatar.setQuality(qualityMap[quality]);
   },
   playerDiorama: null,
   async bindPreviewCanvas(canvas) {
--- a/game.js
+++ b/game.js
@@ -1985,11 +1985,19 @@ const gameManager = {
     }
     metaversefileApi.setQualitySetting(quality);
     const localPlayer = metaversefileApi.useLocalPlayer();
-    const npcs = metaversefileApi.useNpcManager().npcs;
-    localPlayer.avatar.setQuality(qualityMap[quality]);
-    npcs.forEach((npc) => {
-      npc.avatar.setQuality(qualityMap[quality]);
-    });
+    const trackedApp = world.appManager.getTrackedApp(app.instanceId);
+    const apps = world.appManager.getApps();
+
+    (async()=>{
+        
+      await Promise.resolve(); // wait for quality to change
+      const out = apps.filter(app => {
+        if (app.appType && app.appType == "vrm") return app.updateQuality();
+      })
+      out.unshift(localPlayer.avatar.updateQuality());
+      await Promise.all(out);
+
+    })()
 
   },
   playerDiorama: null,
--- a/game.js
+++ b/game.js
@@ -1983,8 +1983,14 @@ const gameManager = {
       "MEDIUM":2,
       "LOW":1
     }
+    metaversefileApi.setQualitySetting(quality);
     const localPlayer = metaversefileApi.useLocalPlayer();
+    const npcs = metaversefileApi.useNpcManager().npcs;
     localPlayer.avatar.setQuality(qualityMap[quality]);
+    npcs.forEach((npc) => {
+      npc.avatar.setQuality(qualityMap[quality]);
+    });
+
   },
   playerDiorama: null,
   async bindPreviewCanvas(canvas) {
--- a/metaversefile-api.js
+++ b/metaversefile-api.js
@@ -819,6 +819,12 @@ export default () => {
   getAvatarHeight(obj) {
     return getHeight(obj);
   },
+  getQualitySetting(){
+    return localStorage.getItem('avatarStyle') || 'MEDIUM';
+  },
+  setQualitySetting(quality){
+    localStorage.setItem('avatarStyle', quality);
+  },
   useInternals() {
     if (!iframeContainer) {
       iframeContainer = document.getElementById('iframe-container');
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "0.0.1",
   "description": "",
   "main": "index.mjs",
+  "workspaces":[
+    "./packages/*"
+  ],
   "scripts": {
     "start": "forever start index.mjs -p",
     "stop": "killall -SIGINT webaverse || true",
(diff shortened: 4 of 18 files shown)


0reo:
moved access of quality settings for readability/ease of access.  also enabled the unimplemented quality settings, and have a reasonable default being used if one has not yet been set.  the crunched avatar seems visually lighter than I expected, but the materials look correct when looking through the object.  would be good to get confirmation that it's correct.

0reo:
@amnaarshed4224 this is to resolve #2321, it's still a work in progress but you should be able to test changing the quality settings.  the following still needs to be address
- turning a vrm into an npc resets it's position
- diorama bug fixes
- crunched avatar for diorama & post processing
- possibly having the crunched avatar set in the vrm template like i did with the sprite mesh, would be good to hear opinions on that
- possibly adjusting skinning/unskinning logic to make initial loading faster

also note this has a totum change.  the .gitmodules file is currently updated to point to my forked repo.
https://github.com/0reo/totum/commit/0e74f135ea22ad707c7aa98b7384fd10f58a05b0

avaer:
> the crunched avatar seems visually lighter than I expected, but the materials look correct when looking through the object. would be good to get confirmation that it's correct.

That is correct behavior; almost by definition. The shader is much simpler than the MToon which is used at the highest quality setting. Additionally, the cruncher uses one material for the entire avatar which is usually not the case with the MToon shaded avas.

avaer:
> making the VRM's into avatars discards their initial set position. I can change that, but wanted to make sure I'm going in the ideal direction before making more changes to to the avatar/NPC logic.

I think it should be possible to make it work so that when you go into a VRM you "embody" it, while "disembodying" the previous one. So we could for example keep the head transform mapper correctly during the transition, which is what `localPlayer.position` is.

avaer:
> the VRM templates were running the toon shader on the models twice, once in default for unskinned, and one in the setSkinning function, and so it's currently doing that for spriting as well. I assume it should only need to be run once per VRM and reused?

Yes, that is correct, there are a lot of inefficiences in the avatars rendering.

Particularly:

1) There is no real need to have skinned/unskinned avatars at all. This was a stopgap put in place because the bones computation was a killer and not needed for the unskinned (unworn) avatars case, and we wanted to get around it, so we created a mesh that doesn't do it. We would ideally re-use the avatar when we wear it; we just need to perform the proper resets and matrix update skips to keep the updates fast.
2) We are not using the crunched version of the avatar for depth computations, when we really should. The avatar should not contributed 50 more geometry draw calls just to get its depth, when it could be one.

avaer:
> applying the quality changes to all the avatars takes quite a bit of time, it can surely be sped up, but I didn't want to go too far off task adding a loading screen. Is the expectation that all the down-scaling will be done on the client side on load? Any plans to store/reuse models online/locally?

This should ideally be done in a thread for starters, so it doesn't hitch. It is indeed a heavy optimization process to create a full animated spritesheet/texture atlas for an avatar.

> Is the expectation that all the down-scaling will be done on the client side on load?

No, once we get it working in a separate thread, it is straightforward to set this up as part of the backend by importing that code. This could be done in `totum`, and cached, for example.

0reo:
> Maybe removing the dial skinned/unskinned VRM model format should be part of this PR, since it might be getting in the way of clean initialization code.

sounds good to me

avaer:
This PR touches on:

https://github.com/webaverse/app/issues/2339
https://github.com/webaverse/app/issues/2341
https://github.com/webaverse/app/issues/2343

0reo:
I've removed the unskinned meshes from the vrm template and cleaned up some bugs when switching between settings.  There is an option to make other vrm's NPC's, but they are no longer NPC's by default.  I'll push the branch here once I clean up the commit history tomorrow, but i'll link the temporary branch here until that's been done.

I've also started on the diorama updates, would like to confirm that the diorama avatar will ONLY be the crunched avatar, regardless of the user settings, or if it will match the user's settings.

https://github.com/webaverse/app/compare/master...0reo:temp-changes?expand=1

avaer:
> I've removed the unskinned meshes from the vrm template and cleaned up some bugs when switching between settings.

Awesome if true. This has been a problem for a while.

> There is an option to make other vrm's NPC's
I don't understand this... NPCs and VRMs are different types of apps. Can they be converted somehow? I'm probably just confused.

avaer:
> I've also started on the diorama updates, would like to confirm that the diorama avatar will ONLY be the crunched avatar, regardless of the user settings, or if it will match the user's settings.

Since some of the crunched avatars have morph target glitches when speaking, which can look bad in the diorama.

I think the ideal is crunched avatar in the diorama (if it looks good), and also the crunched avatar for all depth rendering. The only place we need a different avatar render is 1) the main color pass in the scene, or 2) spritesheet rendering, which is a completely different type of pass which alpha-discards.

avaer:
There were some recent changes in this area so this PR could use a merge from `master`.

metaversefile-api.js
```diff
@@ -869,6 +869,12 @@ export default () => {
   getAvatarHeight(obj) {
     return getHeight(obj);
   },
+  getQualitySetting(){
+    return parseInt(localStorage.getItem('avatarStyle') || 0);
```
avaer, line 873: Should include second arg to `parseInt`.

src/App.jsx
```diff
@@ -11,11 +11,21 @@ import Footer from './Footer.jsx';
...
+
 const {useLocalPlayer} = metaversefileApi;
 
 import dropManager from '../drop-manager.js';
 
 const _startApp = async (weba, canvas) => {
+
+  const gpuTier = await getGPUTier();
+  console.log("GPU INFO", gpuTier);
+  if (!metaversefileApi.getQualitySetting()){
```
avaer, line 24: I would check for undefined or null, so this isn't confused with `0`.

game.js
```diff
@@ -1959,8 +1959,14 @@ const gameManager = {
     localPlayer.removeAction('activate');
   },
   setAvatarQuality(quality) {
+    metaversefileApi.setQualitySetting(quality);
     const localPlayer = metaversefileApi.useLocalPlayer();
+    const npcs = metaversefileApi.useNpcs()
     localPlayer.avatar.setQuality(quality);
+    npcs.forEach((npc) => {
```
avaer, line 1966: IMO we shouldn't iterate the NPCs, but the VRM apps in the `world.appManager`. NPCs are an abstraction over the VRM apps and would automatically be included.

That way even non-NPC VRMs will be optimized.

0reo:
> 
> I don't understand this... NPCs and VRMs are different types of apps. Can they be converted somehow? I'm probably just confused.

When I initially read `Make the setting apply to all avatars in the scene (all [totum](https://github.com/webaverse/totum/) [VRM](https://github.com/webaverse/totum/blob/main/type_templates/vrm.js))`, and I saw how quality settings were being set using avatars.js, I took that to mean that all app's created from the vrm templates would be attached to a type of Avatar.  I saw the NPC class was a type of Avatar and figured that made the most sense to use.  Since we've now spoken about having vrm's in the scene which aren't necessarily NCPs(for example not needing idle animation), I set it up so that in the scene templates(grid.scn for example) you can add 'is_npc: true' to any vrm's listed, and only if set will the scene template attach the vrm app to a new NPC.

package.json
```diff
@@ -5,12 +5,16 @@
   "main": "index.mjs",
   "scripts": {
     "start": "forever start index.mjs -p",
-    "dev": "node index.mjs",
+    "stop": "killall -SIGINT webaverse || true",
```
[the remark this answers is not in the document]
0reo: on first run of `watch` there's no instance of webaverse running , and so the `restart` command would fail.  This allows us to restart regardless of if there's an instance running already or not, while still showing the output from `killall`.

game.js
0reo: got it, this was based on my assumptions regarding NPC's which I commented earlier, and that the setQuality function was part of the Avatar class.  Will be changing to call the vrm apps's directly.

avaer:
> I set it up so that in the scene templates(grid.scn for example) you can add 'is_npc: true' to any vrm's listed, and only if set will the scene template attach the vrm app to a new NPC.

Ok so in the engine there is currently not something like "an npc" -- though there are many utilities included for coding an NPC, the engine itself doesn't track NPCs or implement them. Instead an NPC app uses the APIs to administer itself, providing the actual coded behavior that the NPC is to perform -- how they move, speak, and respond.

But, we can make npcs a first-class concept, and that might be best. For example, the render settings are a first class-concept -- they are a loadable module implemented as a `totum` type, which lets it be used in any scene. We could do the same for NPCs, making a new e.g. `.npc` type which is JSON describing the NPC, and then implemented as a `totum` loader.

However if we go that route, I don't think it should the the _scene_ that is setting up NPCs. The scene should just have NPC typed objects in it.

Here is the current NPC implementation, though it will not work in `master` (it's not used in any scenes are requires a private repo to run): https://github.com/webaverse/npc

character-controller.js
```diff
@@ -166,6 +166,9 @@ class StatePlayer extends PlayerBase {
       appsMap: null,
     });
     this.appManager.addEventListener('appadd', e => {
+      const trackedApp = this.appManager.getTrackedApp( e.data.instanceId);
+      const load = trackedApp.get('load');
```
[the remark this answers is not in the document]
0reo: this was done because i wanted the ability to add a totum app into the app manager, without it being immediately added to the scene.  i had added a new "load" argument to appTrackedApp which defaults to true, setting it to false allows you to manually add it(load it) to the scene by setting it to true at a later time.  Calling it "addToScene" may be more clear.  But it may be unnecessary.

player-avatar-binding.js
```diff
@@ -76,10 +78,31 @@ export function makeAvatar(app) {
         debug: false,
       });
       avatar[appSymbol] = app;
-      
+
+      const quality = metaversefile.getQualitySetting();
+      avatar.setQuality(quality).then(()=>{
```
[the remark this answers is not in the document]
0reo: agreed, it's already removed from player-avatar-binding in the work i did yesterday.

0reo:
> Instead an NPC app loaded through `totum` uses the APIs to administer itself, providing the actual coded behavior that the NPC is to perform -- how they move, speak, and respond. The app is the brain of the NPC which interfaces over `metaversefile` to perform functions and respond to inputs.
> 
got it.
> However, we can make npcs a first-class concept in `totum`, and that might be best. For example, the render settings are a first class-concept -- they are a loadable module implemented as a `totum` type, which lets it be used in any scene. We could do the same for NPCs, making a new e.g. `.npc` type which is JSON describing the NPC, and then implemented as a `totum` loader. That would allow for fully declarative NPCs.
> 
> However if we go that route, I don't think the _scene_ should be setting up NPCs or aware of them. The scene should just have NPC typed objects in it and the scene loads them blindly like anything in Totum.
> 
> I.e. I don't think we should have `is_npc` at all.

understood and aggreed
> 
> Here is the current NPC implementation, though it will not work in `master` (it's not used in any scenes are requires a private repo to run): https://github.com/webaverse/npc However it could quite naturally become a `totum` type handler -- it's all just apps against the same `metaversefile` API.

0reo:
Diorama and depth rendering are addressed in the most recent push.  Quality settings are set up to be applied to all vrms(rather than avatars/npcs), but ran into a roadblock trying to run the avatar cruncher in a worker, which was needed to apply the quality settings to other vrm's now that they aren't using the NPC class(although that still would have really slogged if there were too many items on screen).  I think I figured it out while writing this though.  I was unable to load the vrm to be cloned into the worker, nor could I send the serialised data and recreate the vrm inside the worker, and I couldn't figure out why.  But I believe it's because of references to `window` in metaversefile, which seems to get loaded into most if not all the totum templates.  Wasn't getting the error because the reference to window wasn't in the worker until after it loaded. Updating metaversefile(and possibly some other files) to reference `self` rather than `window` should get everything playing nicely together.  Will test tomorrow, and that should wrap up the ticket.

0reo:
quality settings now apply to all vrms.  that said, there seems to be an issue with crunching miku specifically, regardless of the number of vrms in the scene.

ahadshams:
if it works fine for other VRMs (I can also provide you a few), will be good to know why its not working with Miku but we can park any specific issue related to Miku as its a stock asset and already is flagged as resource intensive

0reo:
> if it works fine for other VRMs (I can also provide you a few), will be good to know why its not working with Miku but we can park any specific issue related to Miku as its a stock asset and already is flagged as resource intensive

sounds good.  i was able to test with scillia and scillia_drophunter_v15_vian models successfully so far.  one guess is the number of bones the model has, but that's just a guess, haven't dug deeper into it yet

0reo:
this has not been reviewed and is likely stale.  I'm going to close this, update it with the current master, and make a new pr

PR #2355, opened 2022-02-12
`overlap()` return all `physicsId`s.

gonnavis:
Now `overlap()` only return the first encountered object/actor and immediately break the `for` loop.
Though fast but the returned one `physicsId` based on `actors`' order.
So use it to filter out object is not stable.
Such as in `marching-cube.scn` the `npcPlayer`'s id is `5` and the `terrain`'s id is `6`, thus will wrongly stopped.
image

So one solution is let `overlap()` run full `for` loop and return all `physicsId`s, what is this PR done.
Sibling PR: https://github.com/webaverse/app-wasm/pull/29

---
I also tried transfer filterId(s) to wasm, and keep the `for` loop's `break`, but don't know why both failed.
https://github.com/gonnavis/app/tree/overlapFilter
https://github.com/gonnavis/app-wasm/tree/overlapFilter
https://github.com/gonnavis/app/tree/overlapFilters
https://github.com/gonnavis/app-wasm/tree/overlapFilters
And feel it's hard to say which has better prformance, it's based on scene.
So just made this one "return all physicsId" PR.


avaer (approved):
Makes sense.

avaer:
Some arguments refactoring recommended on the WebAssembly side: https://github.com/webaverse/app-wasm/pull/29#pullrequestreview-880844367

physx.js
```diff
@@ -833,12 +832,20 @@ const physxWorker = (() => {
       meshPositionOffset,
       meshQuaternionOffset,
       hitOffset,
-      idOffset,
     );
 
-    return scratchStack.u32[14] ? {
-      objectId: scratchStack.u32[15],
-    } : null;
+    let head = outputBufferOffset / 4;
```
avaer, line 837: Minor, but rather than hardcoded `4`, we can use `Float32Array.BYTES_PER_ELEMENT`.

gonnavis: Changed.

avaer (approved):

PR #2380, opened 2022-02-17
Updating README.md

rob-webaverse:
Updating README.md to improve readability and add context.

Updating buildspec.yml to enable ecs staging deploy.


avaer:
Looks like this won't be published. Someone still needs to redo the readme.

PR #2383, opened 2022-02-18
App name/description from Totum

avaer:
Adds a name/description to `App` objects, coming from Totum/the app's metaversefile.

Depends on and ingests https://github.com/webaverse/totum/pull/64


avaer:
@ohmed This can be used in the UI, particularly the World panel tokens list.

PR #2386, opened 2022-02-18
WASM version of `PathFinder`.

gonnavis:
Follow this js version https://github.com/webaverse/app/pull/2298 .
Now trying to convert it into WASM version https://github.com/webaverse/app/pull/2298#pullrequestreview-884586433.

Sibling PR: https://github.com/webaverse/app-wasm/pull/31


gonnavis:
First step: Added `detectPathVoxel()` wasm function, using it instead of js `detect()` to recursively detect `voxel.position.y` for each voxel. Thus each voxel just need one wasm call now.

gonnavis:
The functional part of `PathFinder` wasm version is ok.
Now can get whole `waypointResult` in just one wasm call.
`let waypointResult = physicsManager.detectPathVoxel(npcPlayer.position, localPlayer.position, .5, 1, .5, position, quaternion, 1000, [5]); `

I'll keep converting waypointResult simplification logic, and may try to improve c++ performance.

gonnavis:
The functional part of `PathFinder` wasm version is ok.
Now can get whole `waypointResult` in just one wasm call.
`let waypointResult = physicsManager.detectPathVoxel(npcPlayer.position, localPlayer.position, .5, 1, .5, position, quaternion, 1000, [5]); `
Sibling PR: https://github.com/webaverse/app-wasm/pull/31#issuecomment-1047207633

I'll keep converting waypointResult simplification logic, and may try to improve c++ performance.
Also need rename `detectPathVoxel()` to `getPath()`.

gonnavis:
Converted all functions from js to wasm.
The usage remains the same https://github.com/webaverse/npc/blob/55486c9c184a2e95c2d0b2fa3cd41a3a3cd311e1/index.js#L42 .

physx.js
```diff
@@ -807,6 +807,51 @@ const physxWorker = (() => {
       z
     );
   };
+
+  w.getPathPhysics = (physics, start, dest, hy, heightTolerance, detectStep, maxIterdetect, maxIterStep, maxVoxelCacheLen, ignorePhysicsIds) => {
```
avaer, line 811: IMO this should just take 2 arguments, the start and end.

A-star does not have any other parameters that are not already present in the scene.

It is especially troubling to see a "cache length" here; I still have no idea what that is, or what I should be inputting it since it has nothing to do with A-star.

I also don't think ignore physics ids should be here, since already have 3 other methods for setting PhysX flags, and the algorithm can check that.

avaer (changes requested):
The API for `getPath` is still quite bad and still not usable.

We should change it to take 3 arguments at most:

```
start: Vector3,
end: Vector3,
flags: {flying: true}
```

If it needs anything else then I think we implemented the algorithm wrong.

However, it is good that we have it in WASM now.

avaer:
There is a cleaner API available on `class PathFinder` (https://github.com/webaverse/app/pull/2386/files#diff-6a50a9c285e09666c1bfc3816458b22f3e90b9e124bb5b08b02cdfcbf85ff057R46), but that is just covering up the problem with the algorithm. There should be no cache and no cache length, for example.

avaer:
I'm gonna merge this since it's progress but we still don't have working pathfinding.

avaer:
Confirmed that this causes a crash in master: https://github.com/webaverse/app/issues/2463

So this PR is being reverted.

PR #2390, opened 2022-02-19
Food sfx

avaer:
Uses the standard Webaverse sound system.


avaer:
This PR also changes our usage of `ffmpeg` to `sox` because `ffmpeg` had static sound glitches with this concatenation.

PR #2406, opened 2022-02-21
Animations combo + envelope

avaer:
This PR adds support for three types of animations added to totums:

`animation`: `'animationName'` a regular animation
`animationCombo`: `['animationName']` a sword-like multi-step animation requiring multiple triggers
`animationEnvelope`: `['attack', 'sustainLoop', 'release']` an envelope-based animation with attack, sustain, and release, like a bow


avaer:
`silsword` and `bow` have been updated to the new API.

Issue #2394, opened 2022-02-19
app UI: World token menu has no option to close the menu


amnaarshed4224:
The world token menu has no option to close the menu. Though we can open the menu by pressing 
"Z" but we are unable to close it using "Z". Also there is no bar like there is one in character tab on which we can click and close the menu.

https://user-images.githubusercontent.com/35926530/154808904-a0398d73-196c-430c-a9e9-e17ccad6eb73.mp4

amnaarshed4224:
Fixed

PR #2407, opened 2022-02-21
UI/scene select refactor and inventory bar

ohmed:
- Refactored scene-select menu [NOTE: room create/remove/select doesn't work as it didn't before].
- Base for inventory UI in bottom right corner.
ps: all features and look of scene-select should be same as it was before.

image


avaer:
Seems to be a gap in the screenshot.

ohmed:
- Fixed error [was happening when scene name was not passed via url].
- Removed gap between scene input & scenes list.

image

amnaarshed4224:
1- The scene panel appears above the settings panel.



2- The scenes panel is appearing above the logged in address. I guess its static and doesnt adjusts based on the login component.



3-Can we add the scroll option in scene switching panel? Right now, if we add more scenes we can not access them.



4-Clicking anywhere on the Room UI panel closes it. BTW the delete room used to work before.

https://user-images.githubusercontent.com/35926530/155146983-f2c99442-9085-4af3-a203-08493c462028.mp4

5-Clicking on the scene switching panel's arrow Opens the panel. But re-clicking it doesn't closes it. It works fine in the master though.

https://user-images.githubusercontent.com/35926530/155147562-e6d524cd-8707-4a48-aeff-4554cd3bd3b0.mp4

amnaarshed4224:
6- Also can we make these circles of same sizes. It would look great.

ohmed:
- Fixed SceneSelect css [fixed zIndex, added more offset from right side (static atm), it'll be changed when login UI will be update].
- Fixed panels toggle issues.
- Improved css of inventory.
- Added scroll for scenes/rooms list.
ps: room remove works, but room create not working.
pps: weapon design is like provided in wireframes and will be updated to final version when will get back to it [still updated size to be same as minimap].

image

avaer:
We can delete that right circle entirely for now.

ohmed:
> We can delete that right circle entirely for now.

But it's in the design. I mean we need it, no ? or we'll remove this part of UI ?

amnaarshed4224:
1-The Scene panel has been moved a bit towards the center, It looked more good at its previous place. Can we adjust it the way it is there in master right now? I mean when we have to login, the scene panel should adjust itself and move towards the center so that it doesn't overlaps with the login address. Just make it responsive.

https://user-images.githubusercontent.com/35926530/155318119-969c59f1-0c4f-4220-af9a-c9baa2da809d.mp4

2-The World bar and the character tab right now appear beaneath the right and left circles of map and other one. Can we make the appear above? It causes another issue that the World items get unaccesable, close to the right circle.

https://user-images.githubusercontent.com/35926530/155318452-fd4554e4-1e85-4262-928a-3c36216a01c4.mp4


3-Can you fix the world panel closing issue?
https://github.com/webaverse/app/issues/2425
https://github.com/webaverse/app/issues/2394

4-The last item in World panel doesn't appears as full. Can we make the scroll bar go a little more down to make it appear full?



5-If we can update the scroll bar's design to something that fits more our them and appears part of the UI, It would be great.

src/components/general/world-objects-list/WorldObjectsList.jsx
```diff
@@ -76,7 +79,16 @@ export const WorldObjectsList = ({ app, opened, setOpened }) => {
 
             case 90: {   // Z
 
-                setOpened( ! opened );
+                if ( opened ) {
```
avaer, line 82: Keys are handled in the IO manager. I don't think React should handle key events.

avaer (changes requested):
I think we need to get rid of all of the key handlers in React, since it distributes handling mysteriously across the codebase.

The chance of bugs with this architecture is essentially 100%.

avaer:
I know the key issue is not related to this PR, it was already a problem. However, it does edit key bindings to fix regressions that were introduced during key handling in React, so we can take the opportunity to fix it properly.

avaer:
On second thought let's just merge this. It's blocking the queue and the PR itself is good.

We can fix the shortcuts in a subsequent PR.

PR #2410, opened 2022-02-21
Correcting the metamask loging and making nft loading work for all supported chains

hamzatrq:
Previously the metamask login was not configured correctly and there was no way of reconnecting the user if he refreshed the page. It is corrected in this PR.

Also the nfts were fetched from opensea which only supported ethereum. In this PR the NFTs are now fetched from webaverse sync server that can fetch NFTs from any supported chains.

Small changes include metamask error handling.


src/tabs/tokens.jsx
```diff
@@ -3,97 +3,102 @@
...
-            //   const imageObjectURL = URL.createObjectURL(res.blob);
-            //   nftPreviews[nft.image_preview_url] = imageObjectURL;
-            //   setNftPreviews(nftPreviews);
-            // });
+      nfts.map(async nft => {
+        try {
+          if (!nft.metadata || !nft.metadata.image) {
+            throw new Error('No image use the default one');
           }
+          const blob = await (await fetch(nft.metadata.image.replace('ipfs://', 'https://ipfs.webaverse.com/'))).blob();
```
avaer, line 49: Is this always going to work no matter the `ipfs` url? And we will be the gateway?

Or is this only for ipfs uploaded to our end?

src/tabs/tokens.jsx
```diff
@@ -3,97 +3,102 @@
...
-            /** Will be switched after Previews-Merge */
-            // preview(nft.image_preview_url, nft.ext, 'png', 100, 100).then(res => {
-            //   const imageObjectURL = URL.createObjectURL(res.blob);
-            //   nftPreviews[nft.image_preview_url] = imageObjectURL;
-            //   setNftPreviews(nftPreviews);
-            // });
+      nfts.map(async nft => {
+        try {
+          if (!nft.metadata || !nft.metadata.image) {
+            throw new Error('No image use the default one');
```
avaer, line 47: IMO trycatch should not be used to `goto`.

avaer (changes requested):
Looks good overall, but some questions.

src/tabs/tokens.jsx
hamzatrq: this line of code checks if the image has following format:
ipfs://Qmf6FowPMZHrXdQXUyxT2YdVntprnSBECegHombaQyggd4

If yes it replaces the ipfs:// with https://ipfs.webaverse.com/

So that browser can resolve it

src/tabs/tokens.jsx
hamzatrq: What do you means by goto?

Here it tries to fetch the image for that NFT from the link which may or may not be on ipfs. If the request fails default image is shown.

src/tabs/tokens.jsx
avaer: So that looks wrong. How is it supposed to know the type? You could use the standard extension/file name handling we have to address this case, but its's not being used here so there is no way it will work.

Alternatively, is there a way to get the type from the HEAD? Will app IPFS servers always echo the correct content-type even if you don't use our path extension hack (not used here)?

src/tabs/tokens.jsx
avaer: This is using `goto`-oriented programming. https://en.wikipedia.org/wiki/Goto

IMO cases like this can just be `if`.

src/tabs/tokens.jsx
hamzatrq: This piece of code is not related to totum. Here the image is fetched for preview only inside the app. No request gets sent to totum. Because it is already image. It fetches the value from "image" attribute of metadata.

the image just gets loaded as src in img element.

```
              <img
                src={nftPreviews[nft.metadata.image] || 'images/object.jpg'}
                className={styles.preview}
              />
```

hamzatrq: Here is where these images appear.
sidebar

src/tabs/tokens.jsx
hamzatrq: Updated the code

avaer:
Abandoned PR, out of sync with upstream.

PR #2413, opened 2022-02-21
Add functions to generate terrain to geometry utils

koreex:


koreex:
Added functions to generate and update infinite terrain to a separate file `geometry-utils.js`

koreex:
Purpose: To add geometry-utils.js that provides interfaces to wasm to generate terrain and update it dynamically
Changes to master: Added a new file that contains `geometryUtils`
To test: Does it pass data to wasm and get back correctly

avaer:
This PR looks abandoned.

It also has no description and is impossible to test.

PR #2418, opened 2022-02-21
Add bow IK

avaer:
This PR adds bow-style use action inverse kinematics in the character physics controller.


avaer:
Existing bows to not need to change; they get this upgrade automatically.

PR #2427, opened 2022-02-23
Player appManager physics update

avaer:
The world appManager tracks all of the apps in the scene. It is also responsible for synchronizing the app's matrix back to the physics object.

However, wearable physics objects would not update while worn, because those apps live in the player's appManager, not the world's.

This PR breaks out the physics update part of the appManager into a separate method and ensures that players call it during the update loop.

This fixes bugs in dioramas, as well as physics highlight shading for currently worn apps.

### DIFF
--- a/app-manager.js
+++ b/app-manager.js
@@ -545,40 +545,41 @@ class AppManager extends EventTarget {
   pushAppUpdates() {
     if (this.appsArray) {
       this.appsArray.doc.transact(() => { 
-        for (const app of this.apps) {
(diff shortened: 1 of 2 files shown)


gonnavis:
This PR, specifically `this.appManager.updatePhysics();` added in `character-controller.js`, broke `heli-hack` repo again.
https://github.com/webaverse/heli-hack/pull/1

I'm trying to fix it again.

gonnavis:
Fixed https://github.com/webaverse/heli-hack/pull/1#issuecomment-1050213109

PR #2431, opened 2022-02-23
Minting [WIP]

hamzatrq:
Allowing users to mint NFTs on polygon. WIP


src/tabs/tokens.jsx
```diff
@@ -3,97 +3,102 @@
...
-            /** Will be switched after Previews-Merge */
-            // preview(nft.image_preview_url, nft.ext, 'png', 100, 100).then(res => {
-            //   const imageObjectURL = URL.createObjectURL(res.blob);
-            //   nftPreviews[nft.image_preview_url] = imageObjectURL;
-            //   setNftPreviews(nftPreviews);
-            // });
+      nfts.map(async nft => {
+        try {
+          if (!nft.metadata || !nft.metadata.image) {
+            throw new Error('No image use the default one');
```
avaer, line 47: Try catch is being used as `goto`. We shouldn't do that.

avaer:
The minting UI should be carried across to the dashboard, so ideally it's portable. It should also work like a Discord upload, with a two-step confirmation screen.

It's hard to tell from this WIP PR where it is headed.

avaer:
Abandoned PR.

PR #2434, opened 2022-02-24
Adds an empty scene to make drag and drop of .scn files easy in app.webaverse.com

amnaarshed4224:
This PR adds an empty scene to master. The purpose of the scene is to make it easy for end user to drag and drop any .scn file into the custom scene, without having to clone the app on local.
The "Custom" .scn contains an invisible grid to make sure avatar doesn't falls down.

https://user-images.githubusercontent.com/35926530/155554747-ad78c4fe-5378-41eb-acae-ea3969c09ad4.mp4

### DIFF
--- a/scenes/custom.scn
+++ b/scenes/custom.scn
@@ -1,0 +1,14 @@
+{
+  "objects": [
+{
+      "position": [
+        0,
+        0,
+        0
+      ],
+      "physics": true,
+      "start_url": "https://github.com/amnaarshed4224/assets/raw/main/floor.glb",
+      "dynamic": false
+    }
+]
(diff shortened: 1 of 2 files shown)


avaer:
I don't think this is a good idea.

There should be import/export scene buttons in the controls, but the UI is still not done so these buttons do not exist.

amnaarshed4224:
Got it. Yes it's just a quick hack to allow app.weba users import scenes until a proper solution is implemented. 
Should I close the PR?

avaer:
Yeah, I think instead it's a feature request to have scenes import/export.

PR #2455, opened 2022-02-25
Fix heli problem && `wear()` add arg `disablePhysics`.

gonnavis:
Fix: https://github.com/webaverse/heli-hack/pull/1#issuecomment-1050815462   No.1.

Physics is now automatically disabled for everything worn.
This will cause heli penetrating the floor after activated.
Even re-enabling physics immediately after `app.wear();` will not prevent.
So I add a `disablePhysics` arg to `wear` to control whether we want disable physics.

Sibling commit: https://github.com/webaverse/heli/pull/1#issuecomment-1050950702

---EDIT---
Sibling PR: https://github.com/webaverse/heli/pull/1


avaer:
I don't think there should be any other arguments to wearing, definitely not arguments enabling/disabling physics. This has a lot of problems in terms of synchronizing apps with each other and over the network.

We can simply disable physics when wearing; I don't think that is a problem. But it shouldn't be an option.

However: note that we still must update all physics transforms while wearing, since we use the physics mesh to highlight the wearable.

avaer (changes requested):
Should remove `disablePhysics` argument.

avaer:
This has been in draft status for a long time. Also, I don't think we have any helicopter object supported since that was never added to any official scenes and therefore will not be tested.

@gonnavis Can we close this?

avaer:
Also CC @Loryhoof, the original author of the helicopter. (it would be awesome to officially support it!)

gonnavis:
> Can we close this?

Oh, yes. Sorry I forgot to close this PR.
Loryhoof backed to work on vehicles, so I have moved to work on ragdoll then animations.

PR #2458, opened 2022-02-25
Avatar refactor

### LINKED ISSUE #2456: Refactor avatar animation/rendering
The avatar animation update loop is also managing state logic, which is complicating adding new animations, transitioning between animations, and using multi-state animations.  Will be separating this logic into classes/state machine which the avatar class will then communicate with.  This should make adding new animations and transitions more straight forward and accessible to other developers

- [x] transition animation logic outside of of avatars.js https://github.com/webaverse/app/issues/2519
- [ ] make reasonable classes for defining animation logic https://github.com/webaverse/app/issues/2533
- [ ] create mechanim like state system for animations https://github.com/webaverse/app/issues/2532

0reo:
PR to resolve #2456 .


avaer (approved):
Looks on track to be a good refactor.

My main concern is that we aren't losing any code paths or functions or adding bugs as the price for this better code, so I would advise being very careful about the copy pastes. One wrong leaky quaternion could be a nightmare.

0reo:
> 

Glad it's looking good so far.  And I agree, I'm trying to be cautious about issues slipping in.  The animationHelper is mainly there to make it easier for me to manage and isolate code in small steps.  Based on what I've done so far, I'm going to focus just on the jump and fly animations for the time being, make sure those work properly in the new setup, and then continue with the other animations one by one.  The other animations should continue to work along side the updated/in-progress ones as is, until we're ready to change them.

0reo:
closing this because it's too large.  will be making several smaller prs to replace

avaer:
Thanks!

PR #2460, opened 2022-02-25
Update physicsObject's position etc after unwear().

gonnavis:
Now after unwear() vehicle, such as hovercraft, it's physicsObject will remain in the initial position, which configed in `.scn` file, not the moved position. If player want re-wear, need go to the initial position then press "E".

This PR update physicsObject's position/quaternion/scale to new ones and update matrix after unwear().


avaer (approved):

avaer:
Thanks!

avaer:
I think this might have caused another regression.

image

PR #2466, opened 2022-02-26
Revert "Update physicsObject's position etc after unwear()."

abeersaqib:
Reverts webaverse/app#2460


ahadshams:
Reverting right now due to regression caused..

PR #2474, opened 2022-02-28
added SERVICE_NAME variable name to buildspec

rob-webaverse:
Making the codepipeline build process a bit more dynamic by adding the AWS project servicename to the build image name.


rob-webaverse:
And making sure that totum always gets updated properly during deployment

avaer (approved):

PR #2475, opened 2022-02-28
Npc compatible appManager

avaer:
Refactoring to support wearables for NPCs. Previously wearables were only for local and remote players.

NPCs now get their own `appManager` which means they can transact with the world like any other player.

Also cleans up the `app` creation APIs, such as making `.addModule` self-initializing in its app metadata. Previously the metadata decorations was a bunch of hacks that happened only in certain places, like the app manager, but not others, like a manual userspace metaversefile load.

The end result of these changes is that we support NPC JSON wearable definition, since it now runs the same code as putting an object in the world and manually wearing it.

Depends on https://github.com/webaverse/totum/pull/69/files

### DIFF
--- a/app-manager.js
+++ b/app-manager.js
@@ -236,15 +236,7 @@ class AppManager extends EventTarget {
         const m = await metaversefile.import(contentId);
         if (!live) return _bailout(null);
         const app = metaversefile.createApp({
-          name: contentId,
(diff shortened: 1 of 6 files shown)


avaer:
Note that some downstream apps need to be updated to the new API.

PR #2486, opened 2022-03-03
Audio manager

avaer:
Break out audio context management to a separate file.

Needed for some new engine rendering features that do not involve the full app.


avaer:
CC @abeersaqib

abeersaqib:
Well recieved. I have synced up changes and they're working fine/

PR #2489, opened 2022-03-03
Avatar async loading

avaer:
Part of avatar loading involves running the VRM springbones importer, which is implemented asynchronously upstream (that doesn't make sense to me, since you already have 100% of the data by that point, but that is the upstream `three-vrm` implementation).

That means we cannot truly set up avatars synchronously, though we had handling in place to ignore spring bones until they snap into existence, usually within frames of the load.

This is usually perfectly fine, but it fails in the case where accurate, reproducible control of hair is required as the avatar loads in (such as motion video). This PR attempts to address that by adding async waiters to all cases where we use avatar -- which it turns out were already `async`, so it's not a big change.

This PR successfully fixes the bug, even improving the initial avatar load experience to be less janky, though it adds more `async` complexity just to enable this upstream corner case, which really makes no sense to me in the first place.

IMO the best path would be to either:
1) Figure out [VRMSpringBoneImporter](https://github.com/pixiv/three-vrm/blob/5ab93c5640ea3b7594e2072b0e0b5615ae3f0ee4/packages/three-vrm/src/springbone/VRMSpringBoneImporter.ts#L22) in the upstream, possibly rewriting it to be synchronous
2) failing that, make avatars loading completely async in `async createAvatar()` instead of being a two-step process (`a = new Avatar(); await a.waitForLoad();`)

### DIFF
--- a/avatar-spriter.js
+++ b/avatar-spriter.js
@@ -1498,7 +1498,7 @@ const _renderSpriteImages = skinnedVrm => {
     visemes: true,
     debug: false,
   });
-  await localRig.waitForLoad();
+  // await localRig.waitForLoad();
   for (let h = 0; h < 2; h++) {
     localRig.setHandEnabled(h, false);
   }
--- a/avatar-spriter.js
+++ b/avatar-spriter.js
@@ -1498,6 +1498,7 @@ const _renderSpriteImages = skinnedVrm => {
     visemes: true,
     debug: false,
   });
+  await localRig.waitForLoad();
   for (let h = 0; h < 2; h++) {
     localRig.setHandEnabled(h, false);
   }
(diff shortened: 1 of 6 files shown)


avaer:
The `async` is due to THREEJS GLTF parser's `getDependency` method: https://github.com/pixiv/three-vrm/blob/5ab93c5640ea3b7594e2072b0e0b5615ae3f0ee4/packages/three-vrm/src/springbone/VRMSpringBoneImporter.ts#L140

avaer:
It looks like `VRMSpringBoneImporter` can be vendorized to fix the problem at the root.

avaer:
I managed to de-async the upstream method by pulling from the already loaded GLTF nodes.

We already vendorize `three-vrm` but it will require an update.

avaer:
https://github.com/webaverse/three-vrm/pull/3

Issue #2337, opened 2022-02-09
Deployment infrastructure


avaer:
Folks have tried unsuccessfully to set up containerization/auto deployment of `app` but never succeeded, and we need to make sure `app` is up and current.

Might have to do this in the next short while.

Issue #2239, opened 2022-01-21
[Architecture] Improving Cache Loading & Seperating Metaversefile.import from React.


abeersaqib:
Our current architecture is taking a lot of time in loading and caching based on the machine and this is effecting our caching strategies with caches being invalidated and all in order to solve this problem out an over view of architectures are attached below

![image](https://user-images.githubusercontent.com/5277220/150600173-3888aba8-f51f-4a52-a9b6-7fc16c9eb119.png)

It is proposed that we will seperate the Metaverse-file import entirely from the React application while still keeping the react refresh in the vite.config we will remove the rollup.config from vite and will actually change it to an entirely new implementation called as Metaverse-Server. the Metaverse-server will be composed of the functionality of rollup in iteself and will provide the option to compile the modules using Just in time compilation mode. Using JIT will let each user have their own cache that will be invalidated in X hours of time. 

The two functions which will be impacted by this are:

![image](https://user-images.githubusercontent.com/5277220/150599073-8e931053-3394-4fc0-8009-c4c35920d73d.png)

In this architectural change 

- No changes are required in Totum 
- No changes are required in React
- Instead a simple branch out will happen when the `index.mjs` recieves the import request this is where the Just in Time compilation will happen.

Issue #1935, opened 2021-12-05
Vite takes a long time to build first time


yaserOSource:
![image](https://user-images.githubusercontent.com/51108458/144750418-76bc396c-0ca5-4406-8ee4-0da3fbc3a16e.png)
Hangs on this for a long time

avaer:
Yes it's quite a nightmare :(. Esbuild is supposed to be fast.

I suspect almost all of this is the massive, complicated libs we include, like THREE.js, and the complex dynamic import dependency chain that causes it to have to restart the build once it finds new includes.

Note that the latter is somewhat fundamental to the dynamic import loading nature of webaverse/totum, so I'm not sure if/how we could optimize that easily. However, I am convinced with a day or two of thought this could be made blazing fast, possibly with some hacks. I think it's important we get this dev experience right.

I believe vite.js has ignore path options we can use. Failing that, we have full control of the server and we can short-circuit any import that causes these issues.

avaer:
We could also temporarily disable things like ceramic which are not hooked in yet.

yaserOSource:
Some very good points about THREE, especially when doing * as THREE, but agree about dynamic imports and the nature of the architecture.

I know we can also try tree-shaking with named imports, that could possibly speed up the initial build.

PR #2496, opened 2022-03-03
New architecture

abeersaqib:
### Working 

- Build process using esbuild, rollup & vite.
- Serving of compiled assets.
- Realtime compile server and transform server for Metaverse-Modules and loading.
- Everything else :)

### Not Working

- everything is working w.r.t break. Architecture breaks nothing.

#### `index.mjs`

- The file implements additional functions to support the production mode. It uses the `-p` flag to know whether it's in production mode or development mode. if the file is in development mode it will pass the request to the vite ssr server and will leverage the dynamic compile. 
- If the server is in production it will serve the assets under the `dist` and dynamic assets are compile through the function `dynamicImport` 
- `dynnamicImport` helper function simply calls the `metaverserfile`'s rollup.js and leverage it's `resolveId` and `load` to serve the module. 

#### `Rollup.JS`

- Whenever a module is dynamically loaded it's inside an environment where it don't the exact paths of the `Metaversefile` & `Three` so a new Babel Transformation is added for all of the metaverse modules that changes the path of the imports dynamically. 

#### `ModuleRewrite` 

- Babel transform passes the path of the moduleRewrite file which process the import path and change them with the server path dynamically. 
- Paths are matched so there won't be duplicate instances of metaversefile or ThreeJS

#### `Vite.config.js`

-  Added Build Configuration
-  Added Development Configuration
-  Create a vite plugin to support the smooth development process.

#### `Workers [FOLDER]`

- Moved workers to seperate folder and made that folder available for dynamic loading

edit by @0reo 
closes https://github.com/webaverse/app/issues/2239
closes https://github.com/webaverse/app/issues/2337
closes https://github.com/webaverse/app/issues/1935


abeersaqib:
https://github.com/webaverse/totum/pull/70

avaer:
Re: workers, what about all the microphone handlers scripts?

https://github.com/webaverse/app/blob/84f1b5e8d3604d55dc1a6d4892305f8b587e5378/avatars/avatars.js#L8

avaer:
Does production differ from dev? How?

index.prod.mjs
```diff
@@ -0,0 +1,230 @@
...
+          return res.end('Failed to load');
+        }
+
+        id = id.replace('/@proxy/', '');
+
+        totum.load(id).then(({code, map}) => {
+          res.writeHead(200, {'Content-Type': 'application/javascript'});
+          res.end(code);
+        }).catch(e => {
+          console.warn(e);
```
avaer, line 67: Catching error but not responding?

index.prod.mjs
```diff
@@ -0,0 +1,230 @@
...
+        });
+      });
+    } else {
+      req.originalUrl = loadUrl;
+      return /^\/(?:@proxy)\//.test(req.originalUrl) ? proxyReq(loadUrl, res) : res.redirect(req.originalUrl);
+    }
+  } catch (e) {
+    console.warn(e);
```
avaer, line 75: Not responding to errors?

index.prod.mjs
```diff
@@ -0,0 +1,230 @@
...
+      totum.resolveId(loadUrl, reqURL.href).then(id => {
+        /** ID might have /@proxy/ in the start that needs to be trimmed */
+        if (!id) {
+          res.status(500);
+          return res.end('Failed to load');
+        }
+
+        id = id.replace('/@proxy/', '');
+
+        totum.load(id).then(({code, map}) => {
```
avaer, line 63: Why isn't this async? There is a try catch block but then there is this async part with a different flow.

server/boot.mjs
```diff
@@ -0,0 +1,26 @@
...
+    base: './',
+  });
+
+  console.log('Copying Scenes');
+  /** Copy scene directory */
+  execSync('cp -r ./scenes ./dist/scenes');
+
+  console.log('Copying Metaverse_Modules');
+  /** Copy Metaverse directory */
+  execSync('cp -r ./metaverse_modules ./dist/metaverse_modules');
```
avaer, line 21: What if FS is not writable?

Also executing shell is not the best idea. Will this even work on windows?

vite.config.js
```diff
@@ -1,28 +1,258 @@
...
+let plugins = [
+  // reactRefresh()
+];
+
+const baseDirectory = 'assets';
+const proxyModuleBase = '-proxy';
+
+let entryPoints = [
+  {
+    path: './packages/totum/index.js',
```
avaer, line 20: These should not be hardcoded.

vite.config.js
```diff
@@ -1,28 +1,258 @@
...
+    manifest: true,
+    rollupOptions: {
+      preserveEntrySignatures: 'strict',
+      output: {
+        // __vite_skip_esbuild__: true,
+        exports: 'named',
+        minifyInternalExports: false,
+        format: 'es',
+        manualChunks: id => {
+          if (id.includes('three/build')) {
```
avaer, line 192: These should not be config exceptions.

vite.config.js
```diff
@@ -1,28 +1,258 @@
...
     fs: {
       strict: true,
     },
   },
-})
+};
+
+const config = {
+  ...defaultConfig,
+  ...process.env.NODE_ENV === 'production' ? viteConfiProduction : {},
```
avaer, line 251: > `viteConfiProduction`

typo

index.prod.mjs
abeersaqib: Resolved the error; Re-worked code

index.prod.mjs
abeersaqib: Responding to errors now

index.prod.mjs
abeersaqib: Resolved promise/async usage.

vite.config.js
abeersaqib: All entry points are now dynamic and read from the dependencies of package.json this means we should keep only those dependencies which can be exposed to the browsers and later on used inside the metaverse-module

vite.config.js
abeersaqib: manual chunking is removed. Only kept for metaverse-modules.

vite.config.js
abeersaqib: Resolved.

abeersaqib:
All issues raised by @avaer are resolved. Have to revert one of the commits related to workers path and test the new dynamic path/entry points with workers and then PR will be ready for review again.

abeersaqib:
closes 
https://github.com/webaverse/app/issues/2239
https://github.com/webaverse/app/issues/2337
https://github.com/webaverse/app/issues/1935

abeersaqib:
Closing in favour of 
https://github.com/webaverse/app/pull/2668

PR #2504, opened 2022-03-04
Sonic boom mesh [WIP]

avaer:
Need to drop in https://github.com/webaverse/sonic-boom to https://github.com/webaverse/app/tree/master/metaverse_modules first before this will work.


avaer:
Merged thanks to @tcm390!

PR #2505, opened 2022-03-04
UI totum names

avaer:
Quick UI upgrade: render object names in UI instead of instance/content ids.

Totum is decent at telling what the correct name is, and this is returned as a `name` export in all compiled modules.


avaer:
Depends on and ingests https://github.com/webaverse/totum/commit/9d7352bba8ff0ff95837ac1ff80ed4178b7b0a5b

PR #2468, opened 2022-02-26
WASM version of `PathFinder`. ( 2nd try )

gonnavis:
Follow https://github.com/webaverse/app/pull/2386 .
Fix https://github.com/webaverse/app/pull/2386#issuecomment-1051612511 .

Rebuilt wasm files with newest `app-wasm`, tested that no errors now.
Sibling PR: https://github.com/webaverse/app-wasm/pull/34 .

I'll keep improving the api.

PR #2506, opened 2022-03-04
`PathFinder` use npc-local space & full 3d voxel map & support flying path.

gonnavis:
Follow: https://github.com/webaverse/app/pull/2298 https://github.com/webaverse/app/pull/2468
Fix: https://github.com/webaverse/npc/pull/2#issuecomment-1059400088

According to @avaer 's guidance, use npc-local space ( start-dest space ) to perform A* algorithm to get straight path result when `start` `dest` not axis aligned and there's no obstacle between them.
image

Old:
image

New:
image

Core changes please see `npc-utils.js` in this comparison https://github.com/gonnavis/app/compare/108492c6db8eb272f6bceb5acc8f5e771c3f675e...6d39539e091563d0d38429764930b3226e802782
Currently is js + wasm overlapBox() version, I'll turn it into pure wasm afterwards.

---EDIT---
Converted to WASM version.
Sibling PR: https://github.com/webaverse/app-wasm/pull/35


avaer:
> use StartDestSpace to perform A* algorithm to get straight path result when start dest not axis aligned and there's no obstacle between them.

Yes, but I think we should do NPC-aligned voxelization for everything, not just this case. It's probably superior in every way and costs only some small amount of matrix multiplication, but the NPC will always be perfectly aligned with their goal so you won't even notice the voxels.

gonnavis:
Ok, I think I already aligned everything to NPC-local ( start-dest space ).
image

gonnavis:
Implemented full 3d voxel map & A-Star algorithm.

Works fine for flying npc:
image
image
image
image

But need some way to compatible with walking npc:
image

gonnavis:
Made this full 3d version support walking npc too.
image

Todo: Add arg to let user control whether flying or walking path finding.

gonnavis:
Added back `heightTolerance` and seperate it's logics with `voxelHeight` again.

Before: Get stuck when hole height is 1.7: 
image

After: Correctly walk around:
image

PR #2508, opened 2022-03-04
remove refs to setSkinning(using totum fork)

### LINKED ISSUE #2507: references to app.setSkinning should be removed
this is to support an proposed PR for totum where the unskinning logic was removed due to it no longer being needed

0reo:
PR to remove unskinned references in app.  Currently pointing to fork of totum until related webaverse/totum#72 is merged.

closes #2507

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -1128,7 +1128,6 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
     this.isNpcPlayer = true;
   }
   async setAvatarAppAsync(app) {
-    await app.setSkinning(true);
     
     const {skinnedVrm} = app;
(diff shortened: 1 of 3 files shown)


avaer:
Seems like it causes a large perf regression, so we can't use this approach.

avaer:
@renovateb0t Can you point to the break which this fixes?

0reo:
this PR will be required for https://github.com/webaverse/totum/pull/72 to work, as setSkinning will no longer exist in the vrm template.  But based on conversation with @avaer, setSkinning should be replaced with a toggleBoneUpdates function which will ensure equivalent performance.  I'll update and reopen this PR once the relevant changes are ready(or I can make a new PR if preferred).

PR #2510, opened 2022-03-04
add watch to dev deps, restart node on packages change

### LINKED ISSUE #2509: APP: restart npm server when you've changed packages code
currently editing totum and other submodules in the packages directory requires a manual restart to apply.  watch the packages folder for local changes and restart node when they occur.

0reo:
package.json changes to reload node when submodules get edited.
- run `npm insatll` to get watch module
- run `npm run watch` to watch submodules directory

`npm run dev` still works as before if you don't want to watch submodules dir.
closes #2509

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -10,7 +10,9 @@ import wsrtc from 'wsrtc/wsrtc-server.mjs';
 Error.stackTraceLimit = 300;
 const cwd = process.cwd();
 
-const isProduction = process.argv[2] === '-p';
+process.title = process.argv.find(e => e.split("=")[0] == '--name');
+process.title &&= process.title.split("=")[1];
+const isProduction = process.argv.find(e => e == '-p') === '-p';
 
 const _isMediaType = p => /\.(?:png|jpe?g|gif|svg|glb|mp3|wav|webm|mp4|mov)$/.test(p);
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,8 @@
         "eslint-plugin-promise": "^4.2.1",
         "eslint-plugin-standard": "^4.0.1",
         "spectorjs": "^0.9.27",
-        "vite": "^2.5.1"
+        "vite": "^2.5.1",
+        "watch": "^1.0.2"
       }
     },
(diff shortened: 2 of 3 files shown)


avaer:
Shouldn't this be handled inside of vite?

It actually seems like a vite config mistake/vite bug, and this is a powerful solution for something that vite is already supposed to be doing...

avaer:
Perhaps just find a way to add watchers in `vite`?

avaer:
Also, I generally disagree with changing the top level `node` name.

This would just confuse sysadmins. If you want to label for `killall`, you could subshell and `node` from there, which will keep process names non-confusing but providing the same killing ability.

0reo:
this was honestly just a quick workaround I figured could be useful, the readme says 
> Any changes inside the packages folder won't recompile automatically and so will require restarting the entire development server by just running again: npm run dev

so my assumption was there was some reason vite didn't/couldn't watch those.  I can spend a bit of time on a solution using vite at some point after I've finished up with my current tasks

avaer:
> this was honestly just a quick workaround I figured could be useful, the readme says

Yeah, this is a case of:

1) I didn't have 5 minutes to fix this watcher while developing 10 other features
2) Every other dev ran in to it, and I didn't have time to fix it, but this was the workaround I gave
3) Nobody bothered to spend 5 mins to fix this watcher so the hack became tribal knowledge
4) It was baked into the readme
5) Now we have much worse hacks (like this) being PR'd on top of the recommended hack

Basically what I am recommending is to spend a few minutes to add a watcher in vite and cut off this problem at its source.

avaer:
This is a good bug to fix, thanks for submitting this PR and unearthing + documenting the problem. But given my understanding of the situation the whole fix is wrong and we can just do this in vite.

Otherwise I can totally see this bug's situation getting even crazier, where the `node` rename causes sysadmin problems which have a workaround which is then added to the README...

avaer:
I removed that note from the readme.

0reo:
spoke with @abeersaqib, this is actively being addressed in the deployment milestone
https://github.com/webaverse/app/milestone/1

PR #2512, opened 2022-03-04
Gfx-settings-via-api

0reo:
Added the ability to get/set graphics settings from the metaverse api rather than calling localstorage, and get specific/all settings as json.

### DIFF
--- a/metaversefile-api.js
+++ b/metaversefile-api.js
@@ -841,6 +841,13 @@ export default () => {
   getAvatarHeight(obj) {
     return getHeight(obj);
   },
+  getGfxSettingJSON(setting){
+    if (setting) return JSON.parse(localStorage.getItem( 'GfxSettings' ))[setting];
+    return JSON.parse(localStorage.getItem( 'GfxSettings' ));
+  },
+  setGfxSetting(quality){
+    localStorage.setItem( 'GfxSettings', JSON.stringify( quality ) );
+  },  
   useInternals() {
     if (!iframeContainer) {
       iframeContainer = document.getElementById('iframe-container');
--- a/metaversefile-api.js
+++ b/metaversefile-api.js
@@ -53,10 +53,12 @@ class App extends THREE.Object3D {
     this.appType = 'script';
     this.lastMatrix = new THREE.Matrix4();
   }
+
   getComponent(key) {
     const component = this.components.find(component => component.key === key);
     return component ? component.value : null;
   }
+
   setComponent(key, value = true) {
     let component = this.components.find(component => component.key === key);
     if (!component) {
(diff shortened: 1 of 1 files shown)


index.mjs
```diff
@@ -10,7 +10,9 @@ import wsrtc from 'wsrtc/wsrtc-server.mjs';
 Error.stackTraceLimit = 300;
 const cwd = process.cwd();
 
-const isProduction = process.argv[2] === '-p';
+process.title = process.argv.find(e => e.split("=")[0] == '--name');
```
avaer, line 13: Looks like this PR is packing in unrelated things?

avaer (changes requested):
Mixed up commits?

avaer:
Based on https://github.com/webaverse/app/pull/2510/files

avaer:
This PR is good but cannot be merged because it has been based on a currently unmergable PR.

0reo:
@avaer resubmitted with the unrelated files removed.

metaversefile-api.js
```diff
@@ -841,6 +841,13 @@ export default () => {
   getAvatarHeight(obj) {
     return getHeight(obj);
   },
+  getGfxSettingJSON(setting){
+    if (setting) return JSON.parse(localStorage.getItem( 'GfxSettings' ))[setting];
```
avaer, line 845: Won't this crash on first load?

0reo: Good catch.  That line will cause a crash if a specific gfxsettings is requested before the local storage is set. But the functions not necessarily called on first load with a specific setting passed as an arg.  Regardless, needs to be corrected

0reo: this has now been corrected

metaversefile-api.js
```diff
@@ -841,11 +856,18 @@ export default () => {
   getAvatarHeight(obj) {
     return getHeight(obj);
   },
+  getGfxSettingJSON(setting) {
+    if (setting) return JSON.parse(localStorage.getItem('GfxSettings'))?.[setting];
+    return JSON.parse(localStorage.getItem('GfxSettings'));
+  },
+  setGfxSetting(quality) {
```
avaer, line 863: 1) Is there ever a need for an app to be setting this? It's an abusable feature (imagine a world that forces quality against your settings) so I don't want to make it easy for apps to override this.

2) Any such getter usually needs an event system (`EventTarget`). Otherwise how are you going to update when it's changed?

3) This should not be owned by `metaversefile`. These settings should be in `settings.js`, which exports an `EventTarget` and abstracts the local storage aspect, rather than littering it across many files in different directories. And the UI should use `settings.js` so it's one point of control, not different apps with different ideas of what setting are.

avaer (changes requested):
Recommend cleaner settings implementation.

0reo:
closing in favor of #2659

PR #2520, opened 2022-03-05
2519 app avatar.js should be broken up to better separate concerns

### LINKED ISSUE #2519: App: Avatar.js should be broken up to better separate concerns
Avatars file is huge, want to clean it up to make implementing an animation state system more manageable.

- move classes to their own files
- move related configs to new class files
- move _applyAnimation and related animation code into a separate file so animation refactoring can be continued while master continues to be updated(this file will be changed/removed in future updates)
- move animation functions into variables for readability, rather than defining them in an if/else block

0reo:
closes #2519

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -53,6 +53,25 @@ import Blinker from './Blinker.js'
 import Nodder from './Nodder.js'
 import Looker from './Looker.js'
 
+
+const localVector = new THREE.Vector3();
+const localVector2 = new THREE.Vector3();
+const localVector3 = new THREE.Vector3();
+// const localVector4 = new THREE.Vector3();
+// const localVector5 = new THREE.Vector3();
+// const localVector6 = new THREE.Vector3();
+const localQuaternion = new THREE.Quaternion();
+const localQuaternion2 = new THREE.Quaternion();
+// const localQuaternion3 = new THREE.Quaternion();
+// const localQuaternion4 = new THREE.Quaternion();
+// const localQuaternion5 = new THREE.Quaternion();
+// const localQuaternion6 = new THREE.Quaternion();
+const localEuler = new THREE.Euler(0, 0, 0, 'YXZ');
+const localEuler2 = new THREE.Euler(0, 0, 0, 'YXZ');
+const localMatrix = new THREE.Matrix4();
+const localMatrix2 = new THREE.Matrix4();
+// const localPlane = new THREE.Plane();   
+
 const textEncoder = new TextEncoder();
 
 // const y180Quaternion = new THREE.Quaternion().setFromAxisAngle(new THREE.Vector3(0, 1, 0), Math.PI);
@@ -129,7 +148,7 @@ const _makeDebugMesh = (avatar) => {
     let srcGeometry = srcCubeGeometries[scale];
     if (!srcGeometry) {
       srcGeometry = cubeGeometry.clone()
-        .applyMatrix4(avatar.localMatrix.makeScale(scale * baseScale, scale * baseScale, 1))
+        .applyMatrix4(localMatrix.makeScale(scale * baseScale, scale * baseScale, 1))
       srcCubeGeometries[scale] = srcGeometry;
     }
     const geometry = srcGeometry//.clone();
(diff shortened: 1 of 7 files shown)


0reo:
This branch currently address the first 3 points of #2519, specifically
- move classes to their own files
- move related configs to new class files
- move _applyAnimation and related animation code into a separate file so animation refactoring can be continued while master continues to be updated(this file will be changed/removed in future updates)

The remain task is
- move animation functions into variables for readability, rather than defining them in an if/else block

I'm going to start that in another branch, but can easily add it to this one if we want.

Also, for clarification, this is an intermediary step towards a cleaner animation system, more code will be changed/removed from avatar.js, and animationHelper.js will be reduced/replaced with more formal animation and state classes in upcoming tickets.

amnaarshed4224:
Issues:
1-Unable to fly. This error pops up.

crashwhile flying

2-"E" activator crashes the app too and throws this error. This way we are unable to equip avatars, items etc.
errwhile trying to change avatar or equip items


https://user-images.githubusercontent.com/35926530/157268130-ae4bef25-f562-4083-8c34-400b7f88d647.mp4

0reo:
corrected the issues with cubicBezier & localQuaternion which was preventing flight and "E" activator from working

avatars/avatars.js
```diff
@@ -3062,13 +1517,12 @@ class Avatar {
...
       }
     };
-    _updateEyeTarget();
 
     const _updateEyeballTarget = () => {
       const leftEye = this.modelBoneOutputs['Eye_L'];
       const rightEye = this.modelBoneOutputs['Eye_R'];
 
-      const lookerEyeballTarget = this.looker.update(now);
+      const lookerEyeballTarget = this.looker.update(now, this);
```
avaer, line 1525: If `this` is needed every frame, why not pass it in the constructor?

avatars/avatars.js
```diff
@@ -3083,14 +1537,14 @@ class Avatar {
           const headPosition = eyePosition;
 
           // Look at direction in world coordinate
-          const lookAtDir = localVector2.copy(headPosition).sub(position).normalize();
-      
+          const lookAtDir = this.localVector2.copy(headPosition).sub(position).normalize();
```
avaer, line 1540: Why make all cache variables local to `this`?

avatars/avatars.js
```diff
@@ -3328,13 +1762,129 @@ class Avatar {
...
+    
+    this.options.visemes && _updateVisemes();
     _updateSubAvatars();
 
     if (metaversefile.isDebugMode() && !this.debugMesh) {
       this.debugMesh = _makeDebugMesh();
       this.debugMesh.wrapToAvatar(this);
       this.model.add(this.debugMesh);
     }
+    //redundant?
```
avaer, line 1887: It's not redundant, but probably dead code.

avaer (changes requested):
Some `this` issues.

avatars/avatars.js
0reo: yes, i wrote that while digging through things and just forgot to remove.  will remove in the update

avatars/avatars.js
0reo: i made the local* variables avatar was using members of the avatar class, rather than defined outside of the class

avatars/avatars.js
0reo: that does make more sense, will adjust

avatars/avatars.js
0reo: this comment has been removed

alisaad673:
Seems to work fine now. No new errors.

avatars/avatars.js
avaer: Yeah, IMO we do not need extras. If the implementation is polluting side effects across instances, that's probably a variable collision bug.

Otherwise this is only a waste of memory.

avaer (approved):

avaer:
This PR regressed the avatars code quality quite a bit by introducing new global variables and the elimination of most idempotency in the code.

Might need to revert this.

PR #2521, opened 2022-03-06
Pointer lock API abstraction

avaer:
Adds an abstraction in the camera manager for managing pointer locks.

This is for supporting other UIs besides the canvas which might want to request the pointer lock; we do not want the canvas UI code to be confused by that, so we abstract the pointer lock API and provide a set of virtual methods to access it.


avaer:
There should be no functionality change.

PR #2522, opened 2022-03-06
Emotes setting

avaer:
Needed for some motion video, but exposed in the diorama panel.

image


avaer:
This should be fully compatible with chat emotion detection, the numpad emotion setter, as well as multiplayer.

PR #2525, opened 2022-03-07
2507 references to app.setskinning should be removed

0reo:
Updated PR to replace setSkinning references in app to toggleBoneUpdates. Currently set as a draft and pointing to fork of totum until related https://github.com/webaverse/totum/pull/72 is merged.  This also removes some async/awaits that were previously needed because of the cloneVrm class in the vrm template, which was removed in webaverse/totum#72.

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -480,7 +480,7 @@ class StatePlayer extends PlayerBase {
     
     const _setNextAvatarApp = app => {
       (async () => {
-        const avatar = await switchAvatar(this.avatar, app);
+        const avatar = switchAvatar(this.avatar, app);
         if (!cancelFn.isLive()) return;
         this.avatar = avatar;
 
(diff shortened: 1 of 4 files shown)


0reo:
master has been rebased into this branch so it's up to date

avaer:
> This also removes some async/awaits that were previously needed because of the cloneVrm class in the vrm template, which was removed in https://github.com/webaverse/totum/pull/72.

Did you follow all code paths in the grep to find anything that could be de-asynced?

avaer (commented):
It looks good, but one question.

0reo:
> > This also removes some async/awaits that were previously needed because of the cloneVrm class in the vrm template, which was removed in [webaverse/totum#72](https://github.com/webaverse/totum/pull/72).
> 
> Did you follow all code paths in the grep to find anything that could be de-asynced?
> 
> I have a feeling if it's not done in this PR it will never be done, and pointless asyncs will be ongoing tech debt. So best to make sure here.

I did, but I'll do one more just as a sanity check

0reo:
found and removed one last missing async

Issue #2533, opened 2022-03-08
App: Create animation factory


0reo:
functions to run animations are currently created in avatar.js(soon avatarHelper.js), which reference other critical functions for loading animation files, blending, lerping, etc.  We should have a class which will 
- contain all animation logic, from loading of animation files to calculations
- let developers create and register animation functions
- access all registered animations
- allow for animation functions to be used with 1 or more models
- allow for easy association with a registered state
- optionally auto-associate animations with a state-tracked object

PR #2537, opened 2022-03-09
2532 app create state machine for animations

### LINKED ISSUE #2532: App: Create state machine for animations
Currently, avatar state information and animation code is stored in the avatar class.  pr #2520 moves the animation code into a separate file, but state information is still stored in the avatar class, and state transitions are handled in a big if block.  there should be a centralized state system which will:

- be able to register any object for state tracking
- trigger state activation/deactivation
- hold related state data(state name, run time, etc)
- provide access to all tracked objects, all the relevant state information for a those objects
- be able to work along side the existing code base until a full transition can be done

Next steps for this will include
- be able to associate animation functions to be triggered when a state is active(want to set up an animation factory class first)
- execute animation functions internally
- animation layers(different animations for different parts of a model)
- complex animation chains(if state X, trigger state A, then B&C, then D, then back to A)

0reo:
closes #2532 

Initial state machine which will be used for animations.  There is already a state system in place for player actions, which this will work along side.  Future updates to this state machine will allow for an action to trigger animations with multiple states which will be managed here.

This PR currently has a minimal actual implementation, Avatars register themselves for tracking, and states are registered for the player avatar with dummy functions, but it's not actually controlling anything right now(although it could).  Submitting this now because I want access to this as I decouple the animation function handling from the render loop.

State machine can currently:

- register any object for state tracking
- trigger state activation/deactivation
- hold related state data(state name, run time, etc)
- provide access to all tracked objects, all the relevant state information for a those objects
- work along side the existing code base until a full transition can be done
- associate animation functions to be triggered when a state is active(not using this until animation functions are decoupled from animationHelper)
- execute animation functions internally(not using this until animation functions are decoupled from animationHelper)

To help with testing, you can add the following to avatars.ja at line 1132, which will provide access to the state machine and avatar via the console.
```
    window.StateMachine = StateMachine;
    window.avatar = this;
```

you can test triggering the dummy functions by running the following in the console
```
StateMachine; //access state machine, view all registered objects and states
avatar.tracker; //see tracked instance of avatar
avatar.tracker.addToGraph(['jump']) //replace jump with any available action, make sure it's an array when adding to graph
avatar.tracker.activate('jump') //triggers selected state function
avatar.tracker.deactivate('jump') //stop selected state function
```

Also, I'd recommend hiding white space changes when reviewing


character-controller.js
```diff
@@ -484,6 +484,13 @@ class StatePlayer extends PlayerBase {
         if (!cancelFn.isLive()) return;
         this.avatar = avatar;
 
+        this.getActionsList().forEach((name) => {
+          this.avatar.tracker.registerState({
+            name,
+            animationFn: ()=>{console.log(`${name} func ran`);}
```
avaer, line 490: For debugging?

avaer:
The state engine needs thinking about how we are blending states (lerp, multiply, or something else), in which order, and for which bone subset. The order and bone blend functions both matter a lot, and are not easily defined declaratively.

Is the plan just to copy the execution logic from Avatar's `update`, but to abstract the state setting?

avaer:
```
window.StateMachine = StateMachine;
window.avatar = this;
```

If you want this you could just float this change yourself. I use the same trick but by committing this it actually breaks my trick, since it's a name collision.

Can we remove?

avaer (commented):
Looks fine, some questions.

avatars/StateMachine.js
```diff
@@ -0,0 +1,160 @@
...
+    // global state can be stored here
+    this.tracking = new Map();
+    this.states = {};
+    this.graph = new Map();
+  }
+
+  // add object to state machine
+  registerObj(name, obj) {
+    this.tracking.has(name) || this.tracking.set(name, {
+      id: crypto.randomUUID(),
```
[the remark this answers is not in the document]
0reo: I'm easy either way, I'll have the SM use `makeId` and we can have it use UUID's in another ticket if we want to

character-controller.js
0reo: yes, for debugging, will remove

0reo:
> The state engine needs thinking about how we are blending states (lerp, multiply, or something else), in which order, and for which bone subset. The order and bone blend functions both matter a lot, and are not easily defined declaratively.
> 
> Is the plan just to copy the execution logic from Avatar's `update`, but to abstract the state setting?

I'm planning on creating an animation factory which will handle how blending is handled and other animation specific code.  To keep things moving swiftly I'm setting things up the new system to follow the same general execution logic we're currently using, and once the new system is handling that I will continue to expand ordering.  The state machine `graph` will be used to chain and order animation sub-states when actions trigger a state transition.  For example, action A1 could trigger state S1, which has animation X, and action A2 could trigger state S2, which has animation X, then Y, then Z, then back to X.  And yes, they aren't easily defined declaratively which is why that part's not yet included.

https://github.com/webaverse/app/issues/2533 is the animation factory ticket, it doesn't reference bone subsets specifically but I'll make it explicit

0reo:
> ```
> window.StateMachine = StateMachine;
> window.avatar = this;
> ```
> 
> If you want this you could just float this change yourself. I use the same trick but by committing this it actually breaks clean global assumptions for everyone.
> 
> Can we remove?

this is not actually in this PR, just a note for QA if it's helpful for them when testing

avatars/StateMachine.js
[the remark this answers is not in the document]
avaer: Just moving to `makeId` is fine for now.

avatars/avatars.js
```diff
@@ -408,6 +409,9 @@ class Avatar {
       };
     }
 
+    const stateName = "playerAvatar"
+    StateMachine.registerObj(options.name ?? stateName, this);
```
avaer, line 413: I know this isn't used yet, but we cannot depend on `options.name` being unique, so this will collide and overwrite.

What was the intent here?

avaer: Also, why is this mutating its argument rather than returning a tracker?

0reo: Registering objects first checks to see if it's already being tracked on line 55 of StateMachine.js.  If it is, it won't reregister it, it will just take the object being passed and provide it with the tracker.

avaer:
The state mechanism probably needs a redesign. This PR is fine but it's not adding any functionality, and the state engine tracking is bugged (name collisions resulting in shared state).

When there is functionality to add we can re-open.

PR #2543, opened 2022-03-10
Map gen

avaer:
Adds a street map generator under secret code.


avaer:
Should be no functionality change, but this PR includes some optimizations.

PR #2554, opened 2022-03-11
Rename KeyHandlers -> IoHandler

avaer:
Technically it handles mouse events as well. This was confusing developers.


avaer:
Should be no functionality change.

PR #2560, opened 2022-03-12
Patch fix multiplayer

lalalune:
Right now, multiplayer is broken. This *mostly* fixes it.

Occasionally, remote players will have a velocity of 0. The position.hmd doesn't change from one frame to the next, probably because the Z store hasn't propagated in time. So the problem seems to be that the position.hmd from remote player and remote player avatar update are not syncing exactly on the same frames durations, so the remote player ends up updating more than once on the same hmd position on different timestamped frames.

The symptom is that the velocity is dependent on the network data updating consistently, and when these timings are mismatched the velocity drops to 0 and the animation blinks from run back to idle pose.

This PR looks at the velocity from last speed (as calculated by currentSpeed in avatars), then looks at the new velocity, and if they aren't congruent then a glitch is detected, the function returns and lets the animation from last frame play. Seems to work well enough for now, but there is definitely a more elegant way to do this.

One other note -- interpolation is still broken, and probably for the exact same reason. The HMD and character frames are not lining up, and it causes interpolation to wig tf out. So I switched to the Uninterpolated base class for the RemotePlayer, which should be fine for a few players demoing, but we'll need to solve the underlying cause.

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -1393,41 +1393,54 @@ class Avatar {
       }
     }
   }
+  lastMoveFactors = {}
+  lastSpeed = 0
+  lastTimestamp = 0
   update(timestamp, timeDiff) {
+    if(this.lastTimestamp > timestamp){
+      console.log("lastTimestamp > timestamp ????????")
+    }
+    this.lastTimestamp = timestamp
+    if(timeDiff === 0) return console.log("update timeDiff === 0")
     const now = timestamp;
     const timeDiffS = timeDiff / 1000;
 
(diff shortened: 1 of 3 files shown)


avaer:
I think you are right about the cause of the bug, but this seems like a hack being deployed instead of actually understanding the data race. It might look somewhat correct but moves the needle even further from good working code.

IMO someone needs to go in and do the temporal mental tick math to fix this properly.

lalalune:
Agreed. Solving the race condition between the state write and the remote object update is really the issue.

PR #2565, opened 2022-03-13
New diorama shaders infra

avaer:
- Add 3 new diorama shaders and their cycles
- Abstract out diorama shaders into per-mesh types in `background-fx/`
- Remove `canvas` as an argument to diorama construction (now adding canvases dynamically)
- Automatically detect whether a diorama should be renderable (does it have canvases)
- Rewrite diorama binding to be synchronous and not require passing a canvas at all


avaer:
This allows the map to include dioramas.

PR #2570, opened 2022-03-13
Multiplayer Interpolation Work

lalalune:
This fixes some smol multiplayer bugs, changes position + quaternion into a transform update and attempts to smooth out the issues with multiplayer interpolation.

However, it doesn't really work yet with interpolation as of right now.

Here is the latest in conversation and advice from Avaer

[](https://cdn.discordapp.com/attachments/694024816684171345/952500535313973268/unknown.png)m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 1:37 AM
Image
If I disable pushAppUpdates, the mesages send consistently, not fucked up
But then multiplayer is broken, transaction not sync
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 1:47 AM
so our delta is getting set in a way that is messed up
[avaer]() — Today at 3:01 AM
Push app updates is what pushes messages right?
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:01 AM
apparently it's not the only one
avaer — Today at 3:02 AM
apps and players are separate?
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:02 AM
they appear to be...
i assume players are apps but there are some extra things
avaer — Today at 3:03 AM
Players wrap apps
Players have an app manager
with their owned apps
Do you have a pr of the other fixes? 
I can’t help with code that is floating extra changes
Because I can’t know what is running
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:05 AM
I mean, I have the PR that basically just filters it
avaer — Today at 3:05 AM
No I mean the snapshot pushes
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:05 AM
And I've traced it down to -- well we are sending two updates out of wsrtc
avaer — Today at 3:05 AM
And other things I suggested
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:06 AM
It's not snapshots
I tried that, didn't work
avaer — Today at 3:06 AM
Snapshots was def a bug
It did work
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:06 AM
Completely turned interpolation off, and noticed that we are blasting two messages out
Every frame pretty much
avaer — Today at 3:06 AM
yes that is normal
Ok I know what is going on 
And the fix
But you need to push code
Or else I can’t help
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:08 AM
Aside from that PR, I haven't made any meaningful code, I'm confounded why I'm getting position update messages with almost no delta
avaer — Today at 3:08 AM
because 1) no tx for messages 
2) updates need to snapshot immediately
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:09 AM
I can reset up the snapshots but I spent many hours yesterday and just didn't get anywhere because it didn't seem like the bug, at best I was smoothing the issue
avaer — Today at 3:10 AM
I think you don’t get how the messages are sent
Or received
Which is surprising
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:10 AM
I think I do
It's all observers. When the crdtState is updated, call the update observer callbacks
avaer — Today at 3:10 AM
Is position and quaternion handled in the same tick?
Yes or no
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:10 AM
I thought it was but apparently not
avaer — Today at 3:11 AM
Wdym you thought
Lol
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:11 AM
Like, I figured you would serialize together 
And don't understand why not
avaer — Today at 3:11 AM
no man you don’t get my concern
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:12 AM
I mean I understand a lot of facts about the system because I've measured them
avaer — Today at 3:12 AM
If you had the capability to debug this you have already stated at this debug loop 100 times, you should understand the code flow of that function like the back of your hand
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:12 AM
I think I do
avaer — Today at 3:12 AM
And you still seem to not know a basic yes or no 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:13 AM
The answer is no
Quaternions are coming right after position
In different messages
But we are updating both for both messages incoming
avaer — Today at 3:13 AM
Ok then do you see how given the current code that will never work
so that needs to be changed
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:13 AM
I can see the race condition happening between updateAvatar and the subscription fn
I think if we filter messages by position and use that for interpolation?
avaer — Today at 3:14 AM
No
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:14 AM
ok
avaer — Today at 3:14 AM
you need to add transactions
and I know you won’t
So you can’t fix this
I don’t mean that offensively
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:15 AM
I don't understand why I would add a transaction
avaer — Today at 3:15 AM
it’s just kinda facts
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:15 AM
Since there appears to be one applied at the message sender and receiver at some point
avaer — Today at 3:15 AM
Position and quaternion must be snapshotted in the same tick
they must happen at the same time
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:16 AM
I see
But lets assume that quat is identity
avaer — Today at 3:16 AM
Then you must never ever send it
And ever observe or update based on it
Sending separate quat identity would create bugs
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:17 AM
Let's say it's just some arbitrary quat
And we are running in that direction
avaer — Today at 3:17 AM
What direction
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:17 AM
Even if pos and rot are synced, would there be an issue?
Because the issue is that given a constant direction, velocity is flipping to 0 in the avatars/animation controller, even though movement is seemingly consistent 
avaer — Today at 3:19 AM
No
Because you must snapshot once per remote frame
You just have all data then
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:19 AM
Let's say I'm not snapshotting
No interp
No latency
It still doesn't work
We can add the transaction, now that I understand, but I'm not confident
avaer — Today at 3:22 AM
What is the bug you experience when you do that?
When you do not send quats ever
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:22 AM
The final symptom bug that is obvious to the user is that the velocity/currentspeed of the remote player jumps from some number down to 0
I disabled quats, and this was still happening
Well I didn't disable quat sending..
avaer — Today at 3:28 AM
well are you lerping positions?
this seems obvious to me
what the bug + fix is
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:30 AM
not lerping positions
I can, but that's not fixing the central issue of timing
avaer — Today at 3:31 AM
do you have mic?
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:31 AM
yeah
avaer
 started a call that lasted 6 minutes.
 — Today at 3:31 AM
avaer — Today at 3:34 AM
https://github.com/webaverse/app/blob/04508d0aeaec5b76a8c2337dfd88118a02c94a6e/character-controller.js#L1064
GitHub
[app/character-controller.js at 04508d0aeaec5b76a8c2337dfd88118a02c9...](https://github.com/webaverse/app/blob/04508d0aeaec5b76a8c2337dfd88118a02c94a6e/character-controller.js)[](https://opengraph.githubassets.com/bbba841d6b355928e10a4465bf212f18338595aeda022b60ceeacf05633af80f/webaverse/app)
Web metaverse client. Contribute to webaverse/app development by creating an account on GitHub.
app/character-controller.js at 04508d0aeaec5b76a8c2337dfd88118a02c9...
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:42 AM
          const transform = trackedApp.get('transform');
          app.position.fromArray(transform.slice(0, 3));
          app.quaternion.fromArray(transform.slice(3, 7));
          app.scale.fromArray(transform.slice(7, 10));
 
something like that?
avaer — Today at 3:43 AM
app.position.fromArray(transform, 0);
app.quaternion.fromArray(transform, 3);
but yes
slicing is bad, it will create garbage to GC for no reason 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 3:44 AM
ahhh thats slick
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 6:46 AM
I am having a stupid seemingly unrelated (or semi related) bug but the transform seems to be working and did fix interpolation
Should be able to post soon, I think it is fixed
avaer — Today at 6:50 AM
🙏
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:23 AM
welp
didn't fix it :_/
avaer — Today at 8:41 AM
I still don't know what the bug is
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:43 AM
bad animation jitter on remote players
avaer — Today at 8:43 AM
yes, unless you are snapshotting all of the animations at the position update time you will get that
otherwise they will have the exact same problem as the position
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:44 AM
    const observePlayerFn = e => {
      const now = performance.now()
      const timeDiff = now - timestamp
      timestamp = now
      const transform = this.playerMap.get('transform')
      this.position.fromArray(transform, 0);
      this.quaternion.fromArray(transform, 3);
      if (this.avatar) {
        this.updateInterpolation(timeDiff);
  
        applyPlayerToAvatar(this, null, this.avatar, []);
  
        this.avatar.update(timestamp, timeDiff);
      }
    };
avaer — Today at 8:44 AM
this is nontrivial
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:45 AM
Can I get the timestamp from the actual playerMap? Like what the server sent?
avaer — Today at 8:45 AM
you could but that won't solve anything
that is not the bug
the bug is that you are doing wrong updates or too many updates
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:45 AM
too many updatesw
avaer — Today at 8:45 AM
if (this.avatar) {
        this.updateInterpolation(timeDiff);
  
        applyPlayerToAvatar(this, null, this.avatar, []);
  
        this.avatar.update(timestamp, timeDiff);
      }
this part basically
you need to separate in your mind two update phases
there is the update phase of snapshotting and stamping data
and then there is the process of lerping it out
that distinction didn't exist in the code before
but it seems now it must
so you need to do all snapshotting here at network time
and all reading (interpo) at frame time
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:46 AM
gotcha
avaer — Today at 8:47 AM
it actually shouldn't be that hard, but I think you'd need to read through the entire update loop to separate it out
the update loop for avatars specifically
which includes player snapshotting, applying the player to the avatar, then animating
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 8:48 AM
Hmmmm
avaer — Today at 8:48 AM
as long as you make sure the frame update part does not do updates of data (like snapshotting or "ticking" any animation), it should work
the goal would be the update function can be called as many times as you want with whatever time you want and it will not be making decisions based on "what happened last frame", just interpolating 
what to do is a decision made at snapshot time, like selecting animation states 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 10:54 AM
  updateAvatar(timestamp, timeDiff) {
      this.positionInterpolant?.update(timeDiff);
      this.position.copy(this.positionInterpolant.get());
      applyPlayerToAvatar(this, null, this.avatar, []);
      this.avatar.update(timestamp, timeDiff);
  }

    const observePlayerFn = e => {
      const now = performance.now()
      const timeDiff = now - timestamp
      timestamp = now

      const transform = this.playerMap.get('transform')

      this.position.fromArray(transform, 0);
      this.quaternion.fromArray(transform, 3);

      newPosition.fromArray(transform, 0);
      newQuaternion.fromArray(transform, 3);

      this.positionInterpolant?.snapshot(timeDiff);
      this.quaternionInterpolant?.snapshot(timeDiff);
    };
Tried this
also this
I don't understand what i'm doing wrong
avaer — Today at 10:59 AM
// applyPlayerToAvatar
this does a huge amount of per tick work
so it should probably only ever be called on network snap
// this.avatar.update(timestamp, timeDiff);
this function's job is to lerp animations on a per frame basis
so why are they even being called from the same function?
they should happen at entirely different times
so already from the first example your code doesn't make sense to me
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:00 AM
tried it both ways
updateAvatar is called by players manager
observePlayerFn is the network update
avaer — Today at 11:01 AM
updateAvatar is called by players manager
updateAvatar(timestamp, timeDiff) {
      this.positionInterpolant?.update(timeDiff);
      this.position.copy(this.positionInterpolant.get());
      applyPlayerToAvatar(this, null, this.avatar, []);
      this.avatar.update(timestamp, timeDiff);
  }

so that is wrong 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:01 AM
yeah
avaer — Today at 11:02 AM
so you're showing me wrong code, I don't understand if you agree it's wrong
lol
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:02 AM
well if it wasn't wrong it would work
From whatyou said before, applyPlayerToAvatar not being in update makes sense
avaer — Today at 11:02 AM
what I am saying is if these two appear in the same function:
applyPlayerToAvatar(this, null, this.avatar, []);
      this.avatar.update(timestamp, timeDiff);

then your code is wrong 
because they shouldn't even happen in the same tick
they happen in response to different things
avatar update means set up bones and shit for the current frame
apply player to avatar means switch state, compute deltas and set targets and animations because we have new information from the network 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:04 AM
apply player to avatar also sets hmd pose
hmd pose same over multiple frames is where our vector length 0 comes from
avaer — Today at 11:04 AM
HMD pose should be interpolated, inside of avatar
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:05 AM
to my knowledge it isn't
avaer — Today at 11:05 AM
apply player to avatar should set up the snapshot to target
avaer — Today at 11:05 AM
that's only because it's not good code
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:05 AM
okay so i should add a lerp for hmd pose in avatars?
avaer — Today at 11:05 AM
I think so. and those two methods should not be called at the same time
this.avatar.update(timestamp, timeDiff);
this would theoretically do nothing when called on network update (next frame will redo the work and all you did was waste time), but the fact that it's there above means the code isn't sensible 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:06 AM
// called by players-manager remotePlayer.updateAvatar
  updateAvatar(timestamp, timeDiff) {
    if (this.avatar) {
      this.positionInterpolant?.update(timeDiff);
      this.quaternionInterpolant?.update(timeDiff);
      this.position.copy(this.positionInterpolant.get());
      this.quaternion.copy(this.quaternionInterpolant.get());
      this.avatar.update(timestamp, timeDiff);
    }
  }

// called by network observer
    const observePlayerFn = e => {
      const now = performance.now()
      const timeDiff = now - timestamp
      timestamp = now

      const transform = this.playerMap.get('transform')

      this.position.fromArray(transform, 0);
      this.quaternion.fromArray(transform, 3);

      newPosition.fromArray(transform, 0);
      newQuaternion.fromArray(transform, 3);

      this.positionInterpolant?.snapshot(timeDiff);
      this.quaternionInterpolant?.snapshot(timeDiff);

      if (this.avatar) {
        applyPlayerToAvatar(this, null, this.avatar, []);
      }
    };
I assume I want to call positionInterpolant.update there, but I'm def confused on why the timeDiffs would work 
avaer — Today at 11:08 AM
no I think the interpolants need to be owned by the update step
ignore these bad function names, we are reordering things so it will be hard to use these names
the thing is you need your interpolants, all of them, to snap at the observe function
and you need all of your interpolants to be lerped every frame prior to the avatar update, or at get time during the update 
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:10 AM
snap in observe
      this.positionInterpolant?.snapshot(timeDiff);
update prior to avatar update
      this.positionInterpolant?.update(timeDiff);
      this.position.copy(this.positionInterpolant.get());
      this.avatar.update(timestamp, timeDiff);
does that look right?
avaer — Today at 11:10 AM
yeah but this is missing many other interpolant snaps
all of the actionInterpolants
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:10 AM
i see
avaer — Today at 11:11 AM
that's for e.g. are you running or jumping which is just as important to be in sync
but to tick them they have methods
maybe that means applyPlayerToAvatar should happen per frame -- but I don't know if Avatar has its own interpolation, it probably does, and that would be a problem 
so all of that would need to be moved
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:13 AM
avatar has no interpolation, it would need to interpolation based on hmd position
avatar really just cares about velocity
avaer — Today at 11:14 AM
then that sounds like it should come from applyPlayerToAvatar
which is primarily I think an interpolant reader
m̸̿́ö̵͊o̸̍̈́ń̶̀ — Today at 11:14 AM
so instead of setting velocity from hmd, we should set velocity on avatar based on interpolated position
avaer — Today at 11:14 AM
yes I think so
it would be the same thing but actually correct
for the single player case this collapses through inheritance to be realtime
since the interpolants just do not interpolate
the rest of the math is the same


alisaad673:
Issues:
-A major issue is that positions of all the objects in scenes is really messed up. Can you please make it in sync with master for it to be merged?

1-Clicking on Create room isn't working(says todo)
2-Clicking on erithor creates another room named "undefined". Clicking on 'create room' should create another room.
3-The Multiplayer works for walking flying, crouch only right now. What about other features such as voice chat, combat etc
4-Trying to change avatar also gives this err and crashes the other instance.

https://user-images.githubusercontent.com/35926530/158869305-077dbc7d-8e24-4f10-897a-d8d8dd974e19.mp4

equipavatar


5-Trying to equip an object brought in through drag and drop crashes both the instances.

eqip items

https://user-images.githubusercontent.com/35926530/158869562-736d7e32-a0c4-4782-88b0-65c6b1e788d1.mp4

https://user-images.githubusercontent.com/35926530/158869614-6c90b224-e457-4b40-afec-1eaaba1db2d9.mp4

lalalune:
@alisaad673 this was the wrong branch (we had merged all to main on our side) -- just merged in, please try again. The room creation and interpolation *should* be working. If not then we are not on correct branch. Main thing that is not working now is just changing the scene on the server.

PR #2580, opened 2022-03-15
fix avatar spriter to use string based emote ids

0reo:
have avatar-spriter use string based emote id


avaer (approved):
Thanks, I missed this API update.

PR #2602, opened 2022-03-17
disable hair spring bones for sprite rig

0reo:
this prevents the sprite hair from going up during rendering.  the hair doesn't animate, but seems like it hasn't been doing so in a while.  so animation will be addressed in another PR


avaer (changes requested):
I don't think we should disable it and this isn't how it ever was.

Issue #1891, opened 2021-11-30
App: ragdoll physics


avaer:
For death animations. We could make some boxes around the avatar limbs, add some constraints, and it would work.

1) Set up necessary physx bindings
2) create a constrained ragdoll physx group, render with e.g. debug capsules
3) bind to avatar bones

PR #2610, opened 2022-03-18
Ragdoll support

gonnavis:
Fix: https://github.com/webaverse/app/issues/1891
Follow https://github.com/webaverse/app/pull/2309
Sibling PR: https://github.com/webaverse/app-wasm/pull/36

Found that the joint of PhysX is very easy to break, especially the long chain, no good solution was found.
So try to reduce the number of bones, then it is not easy to break, get stable result.

https://user-images.githubusercontent.com/10785634/158933539-9af951fc-cd9a-4551-b962-a22803b03b18.mp4

https://user-images.githubusercontent.com/10785634/158940642-d9ba8841-c284-4fd2-9cd0-4ac2357611ba.mp4

Todo:
- Use box geometry instead of capsule to prevent bone rotation on the ground.
- Add back some bones to get more realistic result, and tweak the position/pivot for each one.
- Set the constraint independently for each joint to get more realistic result.


avaer:
When I was testing this I think I had the twists constraints incorrect. If you set initial twist constraints that are impossible, the model will break as it tries to solve the impossible bone equation.

I don't think there should be a problem with large numbers of bones if we get the joints correct.

But starting with a smaller number of bones is a good idea.

gonnavis:
Still tried ordinary joint first.
Exposed a generic wasm joint api, for simpler/faster investigation, and may can be used for such as vehicle rigs afterwards.
And using it composed a simple ragdoll.

> When I was testing this I think I had the twists constraints incorrect. If you set initial twist constraints that are impossible, the model will break as it tries to solve the impossible bone equation.

Turned out that if set joint positions correctly, they are really not tend to break.

image
image

gonnavis:
Feedbacked ragdoll infos to player's real skeleton/bones.
But currently the formula for each bone is written by hand, need to write a generic formula.

https://user-images.githubusercontent.com/10785634/159979050-277fada4-8f0a-4bad-a725-8cdd0da6d746.mp4

gonnavis:
https://user-images.githubusercontent.com/10785634/160253088-8bf4d754-d7d8-461f-84d9-6d66a1046069.mp4

https://user-images.githubusercontent.com/10785634/160253108-30753d2f-d7a6-4009-85f9-44023da10e55.mp4

gonnavis:
Correctly set global position/rotation.

And found that new generic joint formula and global position/rotation formula works fine with -Z facing, so just restored player -Z axis facing.

Todo:
- Add full bones, such as foot, hand.
- Improve performance.

gonnavis:
Merged master branch and solved conflicts.
Separated `ragdollMesh` from `debugMesh`, because of they need different bones and geometry.
Now only press `N` to toggle ragdoll mode, no need press `H` first.

I know there's a lot of todos, but the current state may worth as first version.
Sibling PR: https://github.com/webaverse/app-wasm/pull/36

https://user-images.githubusercontent.com/10785634/161037579-cb3c1374-b17e-4175-9c80-16e8bcbb65dd.mp4

gonnavis:
Close this PR because of the other PR https://github.com/webaverse/app/pull/2731 includes all the codes of this, and added NPC support, and made many improvements.

PR #2613, opened 2022-03-18
Fix spritesheet avatar

avaer:
- Fixes initial hair rendering
- Supports hair animation when moving
- Fixes jump

Note: jump does not currently have tracked hair, since the avatar does not perform a jump parabola during capture. But that is doable.


0reo:
works but takes about twice as long to load for me as previously.  likely because we're adding more fames hardcoding to 60fps.  if this increase in load time is tolerable, then this is good to merge.

0reo:
optimizations will be addressed via #2619 in the future, merging as is

PR #2620, opened 2022-03-19
Performance tracker

avaer:
Adds per-app GPU + CPU timing stats so we can tell which apps are slow.

(it is almost always the HQ avatars that are the longest to process and render, as expected)

Should be pass-aware as well, so each step of the pipeline is separated.


avaer:
I also rewrote the entire stats infrastructure in React and fixed a few bugs in the process.

avaer:
image

It works for both CPU and GPU. It uses GPU queries for accurate GPU timing, and wraps `useFrame` to calculate the used CPU time.

Note: Avatar CPU is not tracked but can easily be with a small refactor.

Right now vrm apps do not `useFrame` to update, but instead the update is triggered by the character controller, which is why the avatar CPU usage does not show up in the list. We need to move the player + avatar update handling to be under a `useFrame` call for the VRM app, which will get use avatar bones/physics CPU stats included in that app.

avaer:
Breaking out VRM CPU update to a separate issue.

PR #2629, opened 2022-03-20
decrease front dust instanced particle count

tcm390:
decrease front dust instanced particle count from 100 to 20. This is the effect now


avaer (approved):
Code looks ok.

PR #2636, opened 2022-03-21
Scene export (Ctrl-S)

avaer:
This adds the feature of being able to ctrl-s to save the current scene to a file.

Everything gets bundled into `scene.json` and the file is downloaded.


avaer:
I'll hook this up to work with the new `O` key so scenes can be easily opened as well.

PR #2639, opened 2022-03-21
2350 naruto run animation has scale transforms

### LINKED ISSUE #2350: Naruto run animation has scale transforms
The naruto run animation is violating the standard by having scale transforms on the animation.

```
$ node animations-baker.js
```

```
unknown track name naruto run.fbx VectorKeyframeTrack {
  name: 'mixamorigRightHandRing1.scale',
  times: Float32Array(2) [ 0, 0.4000000059604645 ],
  values: Float32Array(6) [
    0.9999999403953552,
    0.9999999403953552,
    1,
    0.9999999403953552,
    0.9999999403953552,
    1
  ],
  createInterpolant: [Function: InterpolantFactoryMethodLinear]
}
```

0reo:
This PR ensures that any animations with animation tracks applying scaling throw an error.  Animation tracks with scales all set to 1 are just removed.  Closes #2350.  Disable white space when reviewing.

as per https://github.com/webaverse/app/issues/2350#issuecomment-1074449065

> Failing that I'm ok with just adding to our code a loop that checks that if there is a scale, it must be all = 1, and then it can be thrown away (if that requirement is violated we should crash because there is currently no way we can render such an animation).

### DIFF
--- a/animations-baker.js
+++ b/animations-baker.js
@@ -359,9 +359,10 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
         }
       }
       // remove scale transform tracks as they won't be used;
-      tracksToRemove.forEach(i => {
-        tracks.splice(i, 1);
-      });
+      let i = tracksToRemove.length;
+      while (i--) {
+        tracks.splice(tracksToRemove[i], 1);
+      }
 
       const walkBufferSize = 256;
       const leftFootYDeltas = new Float32Array(walkBufferSize);
--- a/animations-baker.js
+++ b/animations-baker.js
@@ -329,7 +329,7 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
       const rootBone = object; // not really a bone
       const leftFootBone = bones.find(b => b.name === 'mixamorigLeftFoot');
       const rightFootBone = bones.find(b => b.name === 'mixamorigRightFoot');
-      const allOnes = arr => arr[0] === 1 && arr.every(v => v === arr[0]);
+      const allOnes = arr => arr.every(v => v === 1);
 
       const bonePositionInterpolants = {};
       const boneQuaternionInterpolants = {};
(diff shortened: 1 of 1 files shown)


animations-baker.js
```diff
@@ -344,10 +345,23 @@ const baker = async (uriPath = '', fbxFileNames, vpdFileNames, outFile) => {
           // const bone = bones.find(b => b.name === boneName);
           const boneInterpolant = new THREE.QuaternionLinearInterpolant(track.times, track.values, track.getValueSize());
           boneQuaternionInterpolants[boneName] = boneInterpolant;
+        } else if (/\.scale$/.test(track.name)) {
+          if (allOnes(track.values)) {
+            const index = tracks.indexOf(track);
+            if (index !== -1) {
```
avaer, line 351: Why would this ever not be the case?

animations-baker.js
```diff
@@ -344,10 +345,23 @@ const baker = async (uriPath = '', fbxFileNames, vpdFileNames, outFile) => {
...
+            }
+          } else {
+            throw new Error(`Error with the following track.  All scale transforms must be set to 1. Aborting.\n Animation: ${animation.name}, Track: ${track.name}, values: \n ${track.values}`);
+          }
         } else {
           console.warn('unknown track name', animation.name, track);
         }
       }
+      // remove scale transform tracks as they won't be used;
+      tracksToRemove.forEach(i => {
```
avaer, line 362: I don't think this is legal/correct for all combinations. Splicing changes the index that you must splice for other items.

animations-baker.js
```diff
@@ -330,9 +329,11 @@ const baker = async (uriPath = '', fbxFileNames, vpdFileNames, outFile) => {
       const rootBone = object; // not really a bone
       const leftFootBone = bones.find(b => b.name === 'mixamorigLeftFoot');
       const rightFootBone = bones.find(b => b.name === 'mixamorigRightFoot');
+      const allOnes = arr => arr[0] === 1 && arr.every(v => v === arr[0]);
```
avaer, line 332: What a weird check. Why is the first item special? If you want to know there is at least one item, checking `.length` is better.

avaer (changes requested):
I think the code is wrong.

animations-baker.js
0reo: yes you're right, will adjust

animations-baker.js
0reo: removed redundant check

animations-baker.js
0reo: removed redundant check

animations-baker.js
0reo: corrected to go backwards, to prevent changing index's for values which need to be removed

PR #2642, opened 2022-03-22
Subscene support [WIP]

avaer:
Depends on https://github.com/webaverse/totum/pull/84


avaer:
This PR adds a lot of cleanups to support the subscene street technology (blending + transplanting worlds), as well as the interactive map + fast travel menu.

It's mostly bugfixes, but some supporting features are added:

- apps can be paused
- apps can be "detached" and not tracked by the world system
- `totum` can pass components down to children with the `objectComponents` meta-component

PR #2646, opened 2022-03-22
remove unneeded localPlayer updateds

### LINKED ISSUE #2621: Performance tracker: track avatar CPU
Avatar CPU is not tracked but can easily be with a small refactor.

Right now vrm apps do not useFrame to update, but instead the update is triggered by the character controller, which is why the avatar CPU usage does not show up in the list. We need to move the player + avatar update handling to be under a useFrame call for the VRM app, which will get use avatar bones/physics CPU stats included in that app.

0reo:
closes #2621 

this PR removes the update code which will not be needed once the vrm templates handle that via useFrame.  Currently requires https://github.com/webaverse/totum/pull/97 before it can be merged

### DIFF
--- a/webaverse.js
+++ b/webaverse.js
@@ -8,13 +8,11 @@ import Avatar from './avatars/avatars.js';
 import * as sounds from './sounds.js';
 import physx from './physx.js';
 import ioManager from './io-manager.js';
-import physicsManager from './physics-manager.js';
 import {world} from './world.js';
 import * as blockchain from './blockchain.js';
 import cameraManager from './camera-manager.js';
 import game from './game.js';
 import hpManager from './hp-manager.js';
-import {playersManager} from './players-manager.js';
 import minimapManager from './minimap.js';
 import postProcessing from './post-processing.js';
 import loadoutManager from './loadout-manager.js';
@@ -305,8 +303,6 @@ export default class Webaverse extends EventTarget {
 
           transformControls.update();
           game.update(timestamp, timeDiffCapped);
-
-          playersManager.update(timestamp, timeDiffCapped);
 
           world.appManager.tick(timestamp, timeDiffCapped, frame);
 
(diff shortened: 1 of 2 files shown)


0reo:
ready for review, https://github.com/webaverse/totum/pull/86 which this depends on is also waiting for review

webaverse.js
```diff
@@ -292,19 +290,10 @@ export default class Webaverse extends EventTarget {
...
-          if (this.contentLoaded && physicsManager.getPhysicsEnabled()) {
-            physicsManager.simulatePhysics(timeDiffCapped);
-            localPlayer.updatePhysics(timestamp, timeDiffCapped);
-          }
 
           transformControls.update();
           game.update(timestamp, timeDiffCapped);
-          
-          localPlayer.updateAvatar(timestamp, timeDiffCapped);
-          playersManager.update(timestamp, timeDiffCapped);
```
avaer, line 306: This line is removed and not added in the Torum PR. Where did this go?

avaer (changes requested):
One question.

avaer (changes requested):
No comments on the previous review, and the code has not changed.

https://github.com/webaverse/app/pull/2646#r834694958

avaer:
The Totum PR has been merged but this one has not, so they are desynchronized and I suspect puling `totum` into master at this point would be broken.

avaer:
Totum PR has been reverted: https://github.com/webaverse/totum/pull/86

So it will need to be submitted and merged synchronously with this one next time.

0reo:
new totum PR with useFrame was added to this.  have not recieved any feedback regarding the review notes, but both this and totum should be ready to merge.

0reo:
closing this PR as there will be a general change in approach.  will be addressed in a new PR

PR #2647, opened 2022-03-22
UI: Updated Login component UI and structure.

TheoTheDev:
Updated Login UI and structure.

New UI look:
image
image
image

UI still shouldn't be considered as final, but all features should work.


ohmed:
Updated design:
image
image
image

alisaad673:
Issues:
1- In the current design, the Login part isn't prominent and for the first time it can be misleading with the character tab. We should make separate UI components for both the character tab and Login so that it doesn't goes unnoticed. Also the "Tab" after "login" gives the impression that "Tab" would open the login panel.

image

2-There is no UI for the assets. When we login through either meatmask or discord. Clicking on the address opens the character panel instead of assets inventory panel.

https://user-images.githubusercontent.com/35926530/159712326-13f1ee0a-4b06-4937-8e67-d75bc63c899a.mp4

ohmed:
First -> All UI from Avaer explanations, but open for suggestions.
Second -> metamask or discord inventory will be integrated into Character panel, not sure if there's sense to maintain it atm.

PR #2648, opened 2022-03-23
UI: Fixed zoom issue in MapGen component.

ohmed:
Scroll was not working [zoom in/out] in UI Map section.
Now it's fixed.


avaer (approved):
Thanks!

PR #2649, opened 2022-03-23
remove sonic boom from app once the animations are no longer visible

tcm390:
remove sonic boom from app once the animations are no longer visible

### DIFF
--- a/metaverse_modules/sonic-boom/index.js
+++ b/metaverse_modules/sonic-boom/index.js
@@ -241,15 +241,21 @@ export default () => {
         const group = new THREE.Group();
         group.add(frontwave);
         group.add(frontwave2);
-        app.add(group);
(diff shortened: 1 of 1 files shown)


metaverse_modules/sonic-boom/index.js
```diff
@@ -241,15 +241,21 @@ export default () => {
         const group = new THREE.Group();
         group.add(frontwave);
         group.add(frontwave2);
-        app.add(group);
+        //app.add(group);
 
         
-        
+        let sonicBoomInApp=0;
```
avaer, line 247: This should be a boolean. As well as all of the other cases.

avaer (changes requested):
Looks good otherwise.

avaer (approved):

PR #2650, opened 2022-03-23
Sound script new version

Arthur-Acha:
modification to compile-sounds.sh


avaer:
Actually everything seems like it was already correct. @Arthur-Acha was using Notepad which corrupts scripts with Windows `\r` which `bash` cannot handle.

I'm gonna merge this however since it has a good edit to `compile-sounds.sh`.

PR #2652, opened 2022-03-23
Overworld infra

avaer:
Depends on https://github.com/webaverse/totum/pull/88

### DIFF
--- a/metaverse_modules/barrier/index.js
+++ b/metaverse_modules/barrier/index.js
@@ -110,6 +110,8 @@ export default () => {
   // const {CapsuleGeometry} = useGeometries();
   const {WebaverseShaderMaterial} = useMaterials();
 
+  app.name = 'barrier';
+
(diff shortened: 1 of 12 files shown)


avaer:
- Support new scene-previewer, which allows making seamless street transitions. This can be run on the backend, to both optimize and visualize the street.
- Support detached scene mode, allowing multiple scenes to blend together
- Rendersettings is a first class citizen in the render loop; we can change the render stack with a switching method. This allows rendering multiple scenes, with different render settings
- Support pausing apps so they can be frozen/thawed at no cost. Needed for some kinds of rendering.
- Makes lights non-special. They can simply now be added to the scene by apps (totum is updated to reflect this).
- Removes orthocam/orthoscene and post perspective scenes (they were not used)

avaer:
-Add peer cooldown to barrier transition effect, so natural barrier crossing does not occlude the camera.

PR #2655, opened 2022-03-24
2550 app detect users gpu settings to define default performance settings

### LINKED ISSUE #2550: App: Detect user's GPU settings to define default performance settings
First time users, users who clear their cookies, etc should have reasonable default performance setting applied based on gpu tier.
Use this library
https://github.com/pmndrs/detect-gpu

0reo:
this PR closes #2550.  I've set up defaults for all GFX settings, not just the character details.  All GFX settings that go from `LOW` to `ULTRA` will be set to match the output from detect-gpu.  Settings that can be `ON` or `OFF` will be `ON` if detected gpu tier is above 1, which is `HIGH` or `ULTRA`, and `OFF` if lower than that.

If we want any of these GFX defaults to stay fixed, or we want to change the tiers which they are enabled, the adjustment can be done very quickly.

Due to eslinting, I would again suggest turning off white space when reviewing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@react-three/fiber": "^7.0.6",
         "@shaderfrog/glsl-parser": "^0.1.20",
         "classnames": "^2.3.1",
+        "detect-gpu": "^4.0.14",
         "dids": "^2.4.0",
         "encoding-japanese": "^1.0.30",
         "express": "^4.17.1",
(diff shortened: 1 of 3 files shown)


avaer:
We can't enable this (make low a default) until the settings change is performant/done asynchronously.

avaer (approved):
Code looks ok.

avaer:
Blocked on performant/async settings application.

PR #2659, opened 2022-03-24
2657 create settingsjs file

0reo:
This PR adds and event driven settings manager to handle getting/setting user settings to local storage.

- set function is private, can only add settings using event
- access to the settings manager is done via metaversefile-api, as VRM totum apps will need to be able to load graphics settings
- current handles GfxSettings, ControlsSettings, AiSettings, AudioSettings
- this is needed to continue on the quality settings tasks


settings-manager.js
```diff
@@ -0,0 +1,42 @@
+class SettingsManager extends EventTarget {
+  constructor() {
+    super();
+
+    this.addEventListener('gfxSettingsChanged', e => {
```
avaer, line 5: The casing between event names is inconsistent. Also, we generally do not camelcase in event names, which is the standard for the web platform.

settings-manager.js
```diff
@@ -0,0 +1,42 @@
...
+      const settings = e.data;
+      this.#setSettings('GfxSettings', settings);
```
avaer, line 7: Does it make sense to separate these events at all?

Is there ever a case, besides the settings panel itself, where we are generally interested in a page of settings as opposed to a setting itself? What would be the purpose of such a notification/categorizing it this way for users of this API, other than confusion?

IMO it makes more sense to emit that a specific setting value changed.

settings-manager.js
0reo: noted

settings-manager.js
0reo: i'll change it to a single event.

I also agree that a single value makes more sense.  It's currently accepting the full set of perameters because of how the saveSettings functions are written in the settings tabs, But I'll adjust accordingly

avaer:
Doesn't look like this will make it in.

PR #2668, opened 2022-03-26
New architecture rebase

abeersaqib:
### Working 

- Build process using esbuild, rollup & vite.
- Serving of compiled assets.
- Realtime compile server and transform server for Metaverse-Modules and loading.
- Everything else :)

### Not Working

- Nothing

#### `index.mjs`

- The file implements additional functions to support the production mode. It uses the `-p` flag to know whether it's in production mode or development mode. if the file is in development mode it will pass the request to the vite ssr server and will leverage the dynamic compile. 
- If the server is in production it will serve the assets under the `dist` and dynamic assets are compile through the function `dynamicImport` 
- `dynnamicImport` helper function simply calls the `metaverserfile`'s rollup.js and leverage it's `resolveId` and `load` to serve the module. 

#### `Rollup.JS`

- Whenever a module is dynamically loaded it's inside an environment where it don't the exact paths of the `Metaversefile` & `Three` so a new Babel Transformation is added for all of the metaverse modules that changes the path of the imports dynamically. 

#### `ModuleRewrite` 

- Babel transform passes the path of the moduleRewrite file which process the import path and change them with the server path dynamically. 
- Paths are matched so there won't be duplicate instances of metaversefile or ThreeJS

#### `Vite.config.js`

-  Added Build Configuration
-  Added Development Configuration
-  Create a vite plugin to support the smooth development process.

#### `To Copy Configuration`

- Added a function ESMTree analyser that analyses all the build files and if something is not found by the esmTree is marked as `To Copy` 
- `To Copy` always executes before the bundle write as it is possible that recursive build or entry points may re-write the files during rollup output.

closes https://github.com/webaverse/app/issues/2239
closes https://github.com/webaverse/app/issues/2337
closes https://github.com/webaverse/app/issues/1935


abeersaqib:
ThreeJS examples build in progress.

https://github.com/webaverse/three.js/pull/6

avaer:
It's not clear what the purpose of this PR, or https://github.com/webaverse/three.js/pull/6 is. The THREE.js PR has no documentation.

PR #2669, opened 2022-03-26
use totum app to update avatar quality/remove unusable avatar instance

### LINKED ISSUE #2530: app: Un-wearing the avatar with "medium" settings dispatches an unusable avatar instance
When switched to the "Medium" settings of character details, and then we try to change the avatar, there is another unusable instance of previous avatar left there, Also this instance goes along with the avatar to any scene that we switch to.

https://user-images.githubusercontent.com/35926530/157256309-16c14da4-460e-41ee-bb15-53b7a036a200.mp4

0reo:
also closes #2530 

at this time, totum only handles the base and crunched avatars, so the avatars class still creates the sprite meshes.  sprite meshes will be moved in another PR, at which point we can further reduce/remove the update function from avatars.js

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -1373,26 +1373,20 @@ class Avatar {
         break;
       }
       case 2: {
-        this.crunchedModel = this.crunchedModel ?? avatarCruncher.crunchAvatarModel( this.model );
         this.crunchedModel.frustumCulled = false;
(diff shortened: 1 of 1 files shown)


avaer:
It's not clear which version of `totum` this is pulling in since there is no linked PR.

avaer:
Looks like there will be no more progress on this PR, so closing.

PR #2674, opened 2022-03-27
Offscreen engine

avaer:
Add support for `OffscreenEngine`, a generic way to boot parallel instances of the engine which run in a separate render process.

This uses the trick of TLD+1 separation of apps by `iframe`, where the native GPU contexts in this case are entirely different as the render processes are isolated for security. This allows true GPU parallelism, similar to OffscreenCanvas, but with support in all browsers and no special code need to write a non-DOM binding of the engine.

The API is generic and can parallelize any function. Transferables are supported and automatically detected. You simply import what you want from the engine:

```
        const offscreenEngine = new OffscreenEngine();

        const fn = offscreenEngine.createFunction([
          `import * as THREE from 'three';`,
          function(a, b) {
            return new THREE.Vector3().fromArray(a)
              .add(new THREE.Vector3().fromArray(b))
              .toArray();
          },
        ]);
        const result = await fn([1, 2, 3], [4, 5, 6]);
        console.log('final result', result);
        offscreenEngine.destroy();
```

Totum has been updated to support this.

### DIFF
--- a/engine-worker.js
+++ b/engine-worker.js
@@ -1,5 +1,40 @@
-// import metaversefile from './metaversefile-api.js';
-import {getTransferables} from './util.js';
+const isTransferable = o => {
+  const ctor = o?.constructor;
+  return ctor === MessagePort ||
+    ctor === ImageBitmap ||
+    ctor === ImageData ||
+    // ctor === AudioData ||
+    // ctor === OffscreenCanvas ||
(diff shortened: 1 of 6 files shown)


avaer:
This is needed for parallel items preview rendering.

It can also be used to improve the performance of any non-world GPU render.

PR #2678, opened 2022-03-28
adds unix-path for some system's local environments

antpb:
hey hey! I ran into some issues for local setup that required adding unix-path. Want to make it more seamless for folks that may not know to add the package so maybe this PR will help them. (Did this package-lock with 17.8.0)

Error is `cannot find module 'unix-path`

Some versions changed in package-lock so may want to check those in a dev server to test before merge. 

Loving webaverse! Will try to help contribute to the magic!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
         "stats.js": "^0.17.0",
         "three": "./packages/three",
         "troika-three-text": "^0.43.1-alpha.0",
+        "unix-path": "^0.2.0",
         "webm-writer": "^1.0.0",
         "wsrtc": "./packages/wsrtc",
         "xhr2": "^0.2.1",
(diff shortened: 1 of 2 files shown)


avaer:
Thanks!

This PR indeed fixes the issue, but I think the dependency is upstream in [`totum`](https://github.com/webaverse/app), which is a subrepo of this. So it's probably better to add the missing module there.

I tried fixing with this commit: https://github.com/webaverse/app/commit/b6cea1ad712057eb74fa7ac6eba48e6ef91f7398 (commit in `totum`: https://github.com/webaverse/totum/compare/5a100bdca3896dcc2375ebb2759b1d33ee7f5fac...b5d40255a1b6f39075599d35bb8f03719b0ee0a7)

LMK if that works for you.

antpb:
yeah! Nice! I'll go ahead and close this then. thank you!

Issue #2680, opened 2022-03-28
app: Selecting apps in world tokens gives error


alisaad673:
Getting this error when i try to select and move the world tokens. It was working fine before.

![image](https://user-images.githubusercontent.com/35926530/160445398-f91db301-eb6d-4e41-8a90-26c345902d85.png)

alisaad673:
Now instead of giving error the world tokens don't respond to clicks on them. Therefore, unable to change positions, quaternions, and scale of objects.

https://user-images.githubusercontent.com/35926530/161278794-abbb3657-a651-4036-89d6-115419ad56e6.mp4

alisaad673:
solved

PR #2684, opened 2022-03-28
UI/component refactor part 2

TheoTheDev:
Components refactored and cleaned up:
- DragAndDrop
- LoadingBar
- ObjectPreview
- Stats
- Header.jsx


alisaad673:
I have tested it, no new bugs there.
Though these bugs are there in master. Can you have a look at these and see if they are related and can be fixed here?
https://github.com/webaverse/app/issues/2680
https://github.com/webaverse/app/issues/2690

ohmed:
@alisaad673 I think it's better to solve this issues in separate PRs and not combine with this one.

ohmed:
Resolved merge issues.

src/components/general/drag-and-drop/DragAndDrop.jsx
```diff
@@ -0,0 +1,395 @@
...
+            const app = currentApp;
+            _importApp( app );
+            app.activate();
+            setCurrentApp( null );
+
+        }
+
+    };
+
+    const _mint = ( event ) => {
```
0reo, line 332: this currently does nothing?  if it's going to be updated later, can we add a comment here stating that?

0reo (commented):
generally looks good.  1 question about it, and could use an eslinting

src/components/general/drag-and-drop/DragAndDrop.jsx
ohmed: Not sure about part of dead code,
I also saw some parts, but don't have enough info about this and only the 'author' of the code knows if the code is not important anymore.

ohmed:
> generally looks good. 1 question about it, and could use an eslinting
So the code style rules inside of UI src folder are different from the general projects.
I didn't add separate eslint not to make and confusions.
But if you ask to use same as in the rest of the codebase I can argue cause general code style is quite unreadable comparing to this one.
If it was my decision I would this UI code style entirely for the whole project. I already have eslint config files [hmm, maybe need edits, cause I remembered I have it for TypeScript], but this needs to be general conversation and probably not just here in UI refactor part 2 PR.

src/components/general/drag-and-drop/DragAndDrop.jsx
0reo: ok, we'll leave it for now and find out from the author what the situation is

avaer:
The code improvements are quite nice. However, so far every UI component refactor has caused bugs that slipped through testing, and most refactoring PRs had to be reverted as a result.

Therefore I propose we stop this refactoring work and close these PRs as it's a net loss + cost rather than a benefit.

There are techniques which can be used to refactor without breakage, but historically speaking they have not been used, and every UI refactor PR has been a regression in practice. I think it's reasonable to assume any refactor following in the same vein will have the same problems as the previous PRs, so maybe it's time to develop a rule that we simply are not doing this anymore.

avaer:
Thoughts on the above?

avaer:
No comments, looks like this PR is abandoned for now.

PR #2687, opened 2022-03-29
zeroVector -> rayVectorZero

msub2:
Small change, this is the first error that shows up when trying to enter immersive mode. I'm unsure if this fixes it entirely as I get issues connecting to the wsrtc server when testing locally, but it's been broken since dd021235f117c530af2ca7dbd49ef676c38cba9e

### DIFF
--- a/camera-manager.js
+++ b/camera-manager.js
@@ -316,7 +316,7 @@ class CameraManager extends EventTarget {
     }
 
     const _setCameraToAvatar = () => {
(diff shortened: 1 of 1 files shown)


avaer:
Thanks for the fix, looks good.

Btw, `wsrtc` is used for multiplayer only and should not be related to immersive mode.

We wanted to tackle geo/material coalesce for avatars (which will be a huge performance boost) before fixing VR (which requires a stable high frame rate).

avaer (approved):

PR #2702, opened 2022-03-30
Multiplayer

lalalune:
This fixes multiplayer. There are still some issues with wearables and vehicles, but avatars seem to work in all cases.

We need to make sure that the transform send doesn't break single player -- right now it seems to be sending NaN values so all of the objects are snapping to 0,0,0 -- if we fix this then we should pull this PR in to stay in sync.


avaer:
> We need to make sure that the transform send doesn't break single player -- right now it seems to be sending NaN values so all of the objects are snapping to 0,0,0 -- if we fix this then we should pull this PR in to stay in sync.

Needless to say, I think that is a bug which can be fixed. I don't think there is ever a reason to tolerate NaNs for transforms. Let's find where they are generated and fix that.

PR #2704, opened 2022-03-30
fix sonic boom collision bug

tcm390:
I found there is a sonic boom collision bug, where when you collide with the object, the trail should still be attached behind the avatar.  But the trail position is wrong when collision. And that is because of the direction calculation. So I change the calculation.
this is the bug:

https://user-images.githubusercontent.com/60634884/160922003-af8793a0-77e0-477e-92d0-e031f87e6589.mp4

And this is after I fix:

https://user-images.githubusercontent.com/60634884/160922132-e56ed39b-e568-4700-b3b4-bff00bb46e50.mp4


0reo (approved):
looks good and works locally

PR #2705, opened 2022-03-30
Resolves #2593 UI: Updated Setting panel fade-in/out.

### LINKED ISSUE #2593: Settings panel fade
The settings panel fade is not very good, and not in line with the rest of the app.

We should add a slide or some sort, and make the time to display a lot shorter.

TheoTheDev:
Updated fade in/out animation of Setting panel.
Now it slides same as 'character-select' panel, but half time faster to keep short timing.

### DIFF
--- a/src/components/general/settings/settings.module.css
+++ b/src/components/general/settings/settings.module.css
@@ -12,15 +12,14 @@
     background-color: #000000;
     background-image: linear-gradient(315deg, #000000 0%, #414141 74%);
     z-index: 10;
-    transition: visibility 0s 300ms, opacity 300ms ease;
+    transition: transform 0.5s cubic-bezier(0, 1, 0, 1), opacity 0.5s ease;
+    transform: translate( 0%, 100% );
     opacity: 0;
-    visibility: hidden;
 }
 
 .settings.open {
     opacity: 1;
-    visibility: visible;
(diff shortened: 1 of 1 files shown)


0reo (commented):
logic looks good and works on my end.  but if we're reusing the logic from another panel, should we not have a reusable class for panel fades?

ohmed:
> logic looks good and works on my end. but if we're reusing the logic from another panel, should we not have a reusable class for panel fades?

I also at first had this thought, but the panels have different structure and different behaviour sometimes, different positions, different requirements.
I think reusable class will make things much less flexible and if for example one panel will need some custom update for animation this will lead to 'how the hell' should I change it to work and don't affect other or smth like this.

avaer:
A function component could work for abstract transitions but I agree there are enough different transitions that it might not be worth it.

alisaad673:
Overall works well But can we have the fade out speed same as the fade in speed?

https://user-images.githubusercontent.com/35926530/161328936-5617ff30-5cae-4154-b84e-958a0e52dc26.mp4

avaer:
I think it actually is the same speed, fast. It's just not a transition function that mirrors itself.

One problem is that the opacity and the movement do not follow the same timing functions. Maybe they should?

avaer:
This should have been a trivial change, let's not spend days on it.

PR #2708, opened 2022-03-31
upload new sound

tcm390:
Integrate sfx triggering for objects: 'Sword', 'Pistol', 'Bow'. So I recompile the new sounds.mp3 and sound-files.json in public/sounds folder. Also I add `useSound` in metaversefile-api.js so that we could access sound in totum app.


0reo (commented):
code looks good, needs to be tested to ensure that the time stamps are properly aligned @alisaad673

alisaad673:
Tested the sounds for Silsword, pistol and bow. Pistol sfx had an issue of no sound while there wasn't an object in front of the aimed pistol. But its fixed now.
The sfx for all three sounds is in sync with the animations. The already existing sounds are working fine too.
No new errors and warnings so far.
As the sfx adding PRs in respective repos are dependent on this pr first, so i had to test them by checking out this branch and putting the ghp urls of the three forked branches of pistol, silsword and bow repos.

avaer:
Code looks good, but I’ll test it myself too.

PR #2713, opened 2022-03-31
Quests support [WIP]

avaer:
Adds basic quests support to the engine.

This is handled with the new `quest` type that can be inlined into the scene, much like lights/rendersettings.

The concept of quests is that it's an instruction to the engine that completion of some goal should cause some action. For example, defeating a boss should drop loot.

Using quests to drive a scene provides major advantages:
- It's easy to declare a gameplay/narrative loop without programming
- First class UI support for quests, to keep the game streamlined
- Minimap/pathfinding support built in
- An easy mechanism for dropping automated gameplay rewards (which may or may not have any actual value)
- Quests are tracked in the quest log, allowing the player to bookmark gameplay tasks and keep track of what's happening
- Deep AI integration, so characters know how to join, help, or hinder the player in the quest


avaer:
Depends on https://github.com/webaverse/totum/pull/92

PR #2716, opened 2022-04-01
Street breakout

avaer:
This PR swaps out the street scene street mesh + physics to https://github.com/webaverse/street-base, so it can be shared in overworld traversal.


avaer:
Also removes on avatar from the street scene to improve load time.

PR #2719, opened 2022-04-01
Mobs component

avaer:
Similar to pets support, adds handling for the `mob` component which turns a model into a declarative world mob.


avaer:
Apps which can be used to test this:

https://github.com/webaverse/ghost
https://github.com/webaverse/gremlins

avaer:
Mobs now use the Physx character controller to move, just like npcs.

Pets should be moved to the same system, so that they no longer float in the air, but have physical presence + configurable ai behaviors.

avaer:
The hit tracker was also rewritten to more easily slip into the scene graph using the bind/unbind pattern.

avaer:
There is now a spawner component for spawning multiple mobs. There is no need for a per-mob group class like `gremlins`.

PR #2724, opened 2022-04-02
Npcs component

avaer:
Moves https://github.com/webaverse/npc into this repo under the `npc` component, so npcs get a full standard support in the engine.

This makes NPCs a configured subtype of VRM, rather than a ghost app which puppets a VRM. That should make more sense for the typical art creation process (first you make the avatar and then you add details to it, rather than deciding on some details and then making an avatar for them as an afterthought).

Also updates the local npc definitions to the new component style. The old app will still work for backwards compatibility but they are likely to diverge significantly from here.


avaer:
Depends on https://github.com/webaverse/totum/pull/94

PR #2725, opened 2022-04-02
Simplex noise in procgen lib

avaer:
This PR inlines the simplex lib into the `useProcGen` metaversefile api.

Simplex noise is used all over the procedural generation apps. They all import their own copies of this library, which wastes memory and makes it harder to write new procedural apps since you need to copy boilerplate files.


avaer:
Also includes the `MultiSimplex` wrapper for octaved simplex noise.

PR #2728, opened 2022-04-02
Pets component

avaer:
Breaks out pets functionality into a local metaverse component, for easier editing and adding interactive features coming from mobs implementation.


avaer:
Depends on https://github.com/webaverse/totum/pull/95

PR #2730, opened 2022-04-02
add jump gasp grunt and eat drink sound

tcm390:
1. Load 'actionFiles' and 'voicePack name' in voice-pack-voicer.js.

2. Add a function call 'playGrunt' in character-sfx.js, it would load the voice name from 'localPlayer.voicePack.actionFiles', and play the voice pack by 'localPlayer.voicePack.audioBuffer'. So in totum app, we could play the grunt like this `localPlayer.characterSfx.playGrunt('angry', index)`;

3. In this PR, 2am gave me the new sounds of eat and drink. This is the result:

https://user-images.githubusercontent.com/60634884/161388497-7bb3b661-b78b-4597-b723-1abf9241fb92.mp4

 
4. In character-sfx.js, I play the grunt of jump and gasp. This is the result:

https://user-images.githubusercontent.com/60634884/161388499-e2f870e6-63bd-4c63-b417-fa66c4ff54a9.mp4


avaer:
- Use exponential falloff selection to reduce repeated grunts
- Connect the drink, eat, and grunt sounds to the character's mouth (https://github.com/webaverse/app/blob/dcc65451f4c96b69e3dc478f2df7391a9c6b63b6/voice-output/voice-pack-voicer.js#L121)
- Crunch is kind of annoying and repetitive, should try another
- Try for more variety in the sounds if we can get it, especially for common things like jump

avaer:
- We should choose the "default" voice of each character and tune a voice pack specifically for them. Otherwise it can be jarring if it seems like a different voice actor is playing the grunts.

For this we should just take an existing voice pack and morph it with plugins to sound like the character's default voice endpoint. The current settings are found in `hangout.scn`.

For example for Scillia we should morph to sound like `sweetie belle`.

avaer:
Looks like this isn't being worked on anymore.

PR #2731, opened 2022-04-02
Ragdoll support for NPC(s).

gonnavis:
Fix: https://github.com/webaverse/app/issues/1891
Follow: https://github.com/webaverse/app/pull/2610
Sibling PR: https://github.com/webaverse/app-wasm/pull/36

Videos:

https://github.com/webaverse/app/pull/2610#issuecomment-1084416182

https://user-images.githubusercontent.com/10785634/161391728-004c5a87-92fa-48c4-9ced-9f7dc067b404.mp4

https://user-images.githubusercontent.com/10785634/161393407-7f82ab21-c754-4842-89ce-6093daf816b2.mp4

The github compare https://github.com/gonnavis/app/compare/RagdollPr...RagdollNpcPr seems wrong.
Compared to `RagdollPr` branch, only changed these codes:

image
image
image


avatars/avatars.js
```diff
@@ -388,6 +395,605 @@ const _makeDebugMesh = (avatar) => {
...
+    // https://www.euclideanspace.com/maths/algebra/realNormedAlgebra/quaternions/functions/index.htm
+    // // qa' * qb' = (qb*qa)' 
+    // (A * B)' = B' * A'
+    // (A * B)' = C'
+    // B' * A' = C'
+    // B' * A' * A = B'
+    // C' * A = B'
+    // B' = C' * A
+    // B = (C' * A)'   !!! 
+    {
```
avaer, line 879: This complicated repetitive math should be contained in helper functions.

character-controller.js
```diff
@@ -1237,6 +1237,8 @@ class NpcPlayer extends StaticUninterpolatedPlayer {
       visemes: true,
       debug: false,
     });
+    avatar.app = app;
+    this.appManager.apps.push(app); // note: for `physicsManager.simulatePhysics` can correctly update ragdoll's bones from physx result.
```
avaer, line 1241: I don't think this is valid. The character controller should not be adding apps to the app manager.

avaer: If this is needed and was not added then we should understand why that is the case. It's not the character controller's job to fix it.

avaer: The app should either be transplanted from another appManager, or it should be explicitly added to the ownership of this npc. But to make setting the avatar change app ownership like this is probably not correct.

world.js
```diff
@@ -303,7 +303,14 @@ const _bindHitTracker = app => {
   };
   world.appManager.addEventListener('frame', frame);
   const die = () => {
-    world.appManager.removeTrackedApp(app.instanceId);
+    if (app.npcPlayer?.avatar?.fsm?.states?.ragdoll) {
+      app.npcPlayer.avatar.fsms.send('ragdoll');
+      setTimeout(() => {
```
avaer, line 308: Why timeout?

avaer (changes requested):
Recommend some fixes.

We should also think hard about using `xstate`. We shouldn't just use it for ragdolls, but for everything (if that is the approach), otherwise the code would be really inconsistnet.

So if we are using `xtate` we need to make sure it is fast for game loops (e.g we should know for a fact that it doesn't allocate any memory during the loop, because that will cause unfixable GC performance problems later). Do you have 100% confidence there is absolutely no allocations from reading the library's code?

If there is not 100% confidence in this fact we should not use this library.

avaer:
Another question: are the ragdoll bones theoretically movable by grabbing the avatar VRM bones to re-position? VR posing is a future feature and it would be nice if we can delegate the restitution logic for posing to the ragdoll bones system.

character-controller.js
gonnavis: I'll try to use another proper way.

world.js
gonnavis: If no timeout, the npc will stop updating physics and removed from the scene immediately.
So has no time for ragdoll to play death animation.

It's best to watch the deactive state of all ragdoll bones instead of setTimeout, I'll try.
But using hardcoded time may worth as the first version.

gonnavis:
> We should also think hard about using `xstate`. We shouldn't just use it for ragdolls, but for everything (if that is the approach), otherwise the code would be really inconsistnet.

Yes I want to use `xstate` as state-manager for everything, especially character animation/combat system.
I'll working on this afterwards.

> So if we are using `xtate` we need to make sure it is fast for game loops (e.g we should know for a fact that it doesn't allocate any memory during the loop, because that will cause unfixable GC performance problems later). Do you have 100% confidence there is absolutely no allocations from reading the library's code, and allocations will never be added in the future?
> 
> If there is not 100% confidence in this fact we should not use this library.

I used `xstate` in my [personal game project](https://github.com/gonnavis/annihilate). I feel it's very powerful and handy, and the [performance seems good](https://raw.githack.com/gonnavis/annihilate/dev/index.html). Each character in it built with a [complex `xstate` state-machine](https://github.com/gonnavis/annihilate/blob/230f6522842789bfc5f7c7126cf4ae7000501f0f/src/Maria.js#L67),

But I really not sure whether it allocate any memory during the loop.
I'll try to confirm by reading the source code and get help from the community.

gonnavis:
> Another question: are the ragdoll bones theoretically movable by grabbing the avatar VRM bones to re-position? VR posing is a future feature and it would be nice if we can delegate the restitution logic for posing to the ragdoll bones system.

Yes, I think so. I'll try to make a demo.

character-controller.js
gonnavis: Hello @avaer, I aligned `NpcPlayer` with `LocalPlayer`, added `setAvatarUrl()` function, which will push `app` to `appManeger` automatically. Is it proper?
Sibling commit: https://github.com/webaverse/npc/pull/4/commits/30fcbae4a7820dd87dad8cd1b5e7bf53260212de

I'll keep trying to avoid `avatar.app = app;` in `character-controller.js`.

gonnavis: Found that `localPlayer` also set `avatar.app = app` immediately after `new Avatar` : https://github.com/webaverse/app/blob/01407552c6e28067c0af346ca5b0e6d70646191e/player-avatar-binding.js#L61

So I set `npcPlayer`'s `avatar.app = app` immediately after `new Avatar` is proper already? https://github.com/gonnavis/app/blob/77a79ce31e77d950120ff0f0d621e9e2a4d2481c/character-controller.js#L1244

avatars/avatars.js
gonnavis: Added and moved `getDiffQuaternion()` into util.js. https://github.com/webaverse/app/pull/2731/commits/c9e3acb4f28c348cddba72a2ccad59b39d2f5fd8

gonnavis:
Hello @avaer, I feel I need more time to prove the performance of XState, I'll try in other PRs.
So for this PR, I used state variables instead of XState. https://github.com/webaverse/app/pull/2731/commits/ceeedf4977d9722b330e00b4dcc62d721fcc2d6d

gonnavis:
> Another question: are the ragdoll bones theoretically movable by grabbing the avatar VRM bones to re-position? VR posing is a future feature and it would be nice if we can delegate the restitution logic for posing to the ragdoll bones system.

Do you mean something like this?
I added the grabbing logic in the commented codes for testing. https://github.com/webaverse/app/pull/2731/commits/591a6f73948a7d3593072ffcfc39301ae8571bff

https://user-images.githubusercontent.com/10785634/161782121-01dee374-a71c-4f2f-a1dc-e059c3ce8c0f.mp4

world.js
avaer: Makes sense!

character-controller.js
avaer: No this doesn't work. We already have the VRM app by the time we call `createNpc`. We should not load it again. Also, that app is probably already owned by some app manager. You can't add it to another app manager too.

https://github.com/webaverse/app/blob/f6eb31d3467076946b625381b7b4138a70a5953d/metaverse_components/npc.js#L56

I think what you are missing is update() function calling for this app or NPC? If you need that delegation you can do it in metaverse_components/npc.js, which owns the NPC.

gonnavis: @avaer Now I add them in `npc.js`, is it ok?
https://github.com/webaverse/app/pull/2731/commits/fcb6b4ccd18066de36935985551a611f0f7bf9d9#diff-4cab899c225f756cb4eee00c66e53f8a2cd4399d86b0ae3d87d8d3d93919c68fR64

avaer:
> Do you mean something like this?

Kind of! That demo is really cool!

I meant more being able to pose the avatar in a specific way (like moving the hands up, but keeping the spine in place to create a pose).

It looks like we might be able to do it, but it's probably best to ship this first and make that a separate PR.

avaer:
This PR need a rebase.

gonnavis:
@avaer Solved conflicts.

avatars/avatars.js
```diff
@@ -1820,33 +2342,61 @@ class Avatar {
     
     
 
-    _updateHmdPosition();
-    _applyAnimation(this, now, moveFactors);
+    if (this.ragdoll) {
+      if (this.ragdoll !== this.prevRagdoll) {
+        // console.log('switch to ragdoll');
+        if (!this.ragdollMesh) {
+          this.recordCurrentPose();
```
avaer, line 2349: Maybe this isn't the best way to initialize the ragdoll.

It seems we need the save/restore only for the purpose of creating the ragdoll mesh? If that's true, why can't we make the t-pose an argument to the ragdoll creation? It doesn't seem like we need to mess with the base model pose at all, so we don't need to save/restore the bones.

avaer: Instead of "wrap to avatar" it would be "wrap to pose".

If the code was cleaner, then avatar pose could be a `class` and this would be straightforward passing an object.

avaer (commented):
I think this still needs more cleanup.

The avatars implementation has gotten really bloated, and this adds more significant complexity. Can we find a way to take the ragdoll out into its own file? It would also force some more cleanup in this avatars code.

In the meantime I merged the WASM component to this PR (https://github.com/webaverse/app-wasm/pull/36).

avatars/avatars.js
gonnavis: >It seems we need the save/restore only for the purpose of creating the ragdoll mesh? If that's true, why can't we make the t-pose an argument to the ragdoll creation? It doesn't seem like we need to mess with the base model pose at all, so we don't need to save/restore the bones.

Oh, yes, confirmed again that it's no need to reset to TPose to add joint.

The combination of quaternion of `physicsManager.addJoint` and` setJointMotion/setJointSwingLimit` is the real joint setting.

The initial transform has no effect on the joint setup, but does affect the initial velocity as it tries to get back within the proper joint setup limits.
But our avatars bones should always in the limits, so no need to reset to TPose.
Especially I `resetToRecordedPose()` afterwards, so these are really redundant operations, should no difference if delete them.

gonnavis: Deleted `recordCurrentPose()`, `resetToRecordedPose()` and `resetToTPose()`. https://github.com/webaverse/app/pull/2731/commits/eaa791fd36428ddc818a8c6de90635a8562c0939

> Instead of "wrap to avatar" it would be "wrap to pose".
> 
> If the code was cleaner, then avatar pose could be a `class Pose` or something and this would be straightforward passing an object.

Hello @avaer, I'm not very sure what `class Pose` and "wrap to pose" do.
Is it still needed after the deletion of `recordCurrentPose()`, `resetToRecordedPose()` and `resetToTPose()`?

gonnavis:
Separated/Created `ragdoll.js`. https://github.com/webaverse/app/pull/2731/commits/2ec63f19407c2c0ada5d58618e1462e41aec843d

avatars/ragdoll.js
```diff
@@ -0,0 +1,507 @@
...
+    object.isCreatedRagdoll = true;
+    for (const k in flatMeshes) {
+      const meshBone = flatMeshes[k];
+      // const body = physx.physxWorker.addBoxGeometryPhysics(physx.physics, meshBone.position, meshBone.quaternion, meshBone.sizeHalf, meshBone.physicsId, true, localPlayer.characterController.physicsId); // can't hardcode localPlayer here, used disableGeometryQueries instead.
+      const body = physx.physxWorker.addBoxGeometryPhysics(physx.physics, meshBone.position, meshBone.quaternion, meshBone.sizeHalf, meshBone.physicsId, true);
+      physicsManager.disableGeometryQueries(meshBone); // prevent own CCT collide with ragdoll bones.
+      avatar.app.physicsObjects.push(meshBone);
+      // console.log('mass 1: ', physicsManager.getBodyMass(body));
+      // physicsManager.updateMassAndInertia(body, 0.000001);
+      physicsManager.updateMassAndInertia(body, 0); // note: set mass 0 ( ie kinematic? ) will not break joints and get good result, but got much slow animation.
```
avaer, line 245: I think I understand this code better now. Can we _not_ return the body from addBoxGeometryPhysics? It seems inconsistent with the rest of the code.

We are already injecting the physics ID which I think can be used to look up the body on the WASM side, so there is no need to add yet another id being returned here.

Then, `updateMassAndInertia` can just take the physics ID.

physx.js
```diff
@@ -556,6 +556,35 @@ const physxWorker = (() => {
...
+    position2.toArray(scratchStack.f32, 3)
+    quaternion1.toArray(scratchStack.f32, 6)
+    quaternion2.toArray(scratchStack.f32, 10)
+
+    const position1Offset = scratchStack.f32.byteOffset
+    const position2Offset = scratchStack.f32.byteOffset + 3 * Float32Array.BYTES_PER_ELEMENT
+    const quaternion1Offset = scratchStack.f32.byteOffset + 6 * Float32Array.BYTES_PER_ELEMENT
+    const quaternion2Offset = scratchStack.f32.byteOffset + 10 * Float32Array.BYTES_PER_ELEMENT
+
+    const joint = moduleInstance._addJointPhysics(physics, physicsId1, physicsId2, position1Offset, position2Offset, quaternion1Offset, quaternion2Offset, fixBody1)
```
avaer, line 570: Can we get rid of `fixBody`? It seems redundant, because we can simply perform the lock call right afterwards with the lock function.

physx.js
```diff
@@ -1671,15 +1701,17 @@ const physxWorker = (() => {
     quaternion.toArray(q)
     size.toArray(s)
 
-    moduleInstance._addBoxGeometryPhysics(
+    const body = moduleInstance._addBoxGeometryPhysics(
```
avaer, line 1704: IMO it's best to remove this return value.

See https://github.com/webaverse/app/pull/2731/files#r854723719

util.js
```diff
@@ -1057,6 +1057,16 @@ export const getTransferables = o => {
   _recurse(o);
   return result;
 };
+export function getDiffQuaternion(quaternionA, quaternionB) {
```
avaer, line 1060: For this function it's better to pass in the `target` quaternion from the caller rather than return the util local.

It makes it harder to make variable aliasing mistakes, which can be very hard to debug later.

avaer:
A few more wasm-related comments.

avatars/ragdoll.js
gonnavis: Changed. https://github.com/webaverse/app/pull/2731/commits/02f681682908ef3913316bbe8b1ac0202b170170
Sibling PR: https://github.com/webaverse/app-wasm/pull/44

util.js
gonnavis: Changed https://github.com/webaverse/app/pull/2731/commits/fd2aa56664b0fd41cc6f74bc32f568b57d045fe7

physx.js
gonnavis: Removed https://github.com/webaverse/app/pull/2731#discussion_r855127329

physx.js
gonnavis: Deleted https://github.com/webaverse/app/pull/2731/commits/b72105e5452b9c47a53ee84df114d9fc1ddc7210

game.js
```diff
@@ -927,6 +927,7 @@ const _gameUpdate = (timestamp, timeDiff) => {
           const result = metaversefileApi.getPairByPhysicsId(collisionId);
           if (result) {
             const [app, physicsObject] = result;
+            if (app.npcPlayer?.avatar?.ragdoll) return; // npcPlayer already die and in ragdoll animation.
```
avaer, line 930: Early return in a large function is bad. This logic should be refactored to not need it.

world.js
```diff
@@ -294,7 +295,16 @@ const _bindHitTracker = app => {
   hitTracker.bind(app);
 
   const die = () => {
```
avaer, line 297: I think this is backwards. What controls the dying should be the NPC. The app should not be aware of NPCs. The NPC controls the app.

world.js
```diff
@@ -294,7 +295,16 @@ const _bindHitTracker = app => {
...
-    world.appManager.removeTrackedApp(app.instanceId);
+    if (app.npcPlayer?.avatar) {
+      physicsManager.disableGeometry(app.npcPlayer.characterController);
+      physicsManager.disableGeometryQueries(app.npcPlayer.characterController);
+      app.npcPlayer.avatar.ragdoll = true;
```
avaer, line 301: Definitely should not set this avatar internal state here. This code belongs to the avatar.

lalalune:
@gonnavis where we at with this? still needed, or been replaced?

gonnavis:
> @gonnavis where we at with this? still needed, or been replaced?

Still needed, but need a lot of code structure refactor, and should be changed to `z-targeting` branch, I'll keep working on this.

gonnavis:
Closed in favor of https://github.com/upstreet-labs/app/pull/262

PR #2706, opened 2022-03-30
Resolving #2571 UI: Added preventing page zoom change.

TheoTheDev:
Added preventing zoom page UI by touch-pad gestures.

Issue #2571, opened 2022-03-14
app: Camera zoom in/out messing up with UI zoom in/out


alisaad673:
Right now, we are unable to to zoom in/out camera alone. As UI gets zoomed in/out along with the camera. We need to fi this.

https://user-images.githubusercontent.com/35926530/158172633-3f0efb97-5cad-4efc-b0a9-a90110747a5a.mp4

PR #2741, opened 2022-04-04
Resolving #2571 UI: Added preventing page zoom change.

TheoTheDev:
Sequel of https://github.com/webaverse/app/pull/2706

Meta & CSS params added to prevent zoom by pitch on touchpad.
JS code to prevent zoom by pitch on Mac touchpad [only this can prevent zooming on mac from my experiments].


0reo:
no obvious code issues, can be moved to manual QA

0reo (commented):
no obvious code issues, can be moved to manual QA

avaer (changes requested):
Same code issues as before.

0reo:
@avaer using the wheel event seems to be the preferred way to address this for apple devices
https://bugs.webkit.org/show_bug.cgi?id=145214

and webkit has explicitly disabled `user-scalable=no`
https://webkit.org/blog/7367/new-interaction-behaviors-in-ios-10/

avaer:
> using the wheel event seems to be the preferred way to address this for apple devices

Does it require locking the event to the main thread for all browsers?

TheoTheDev:
> > using the wheel event seems to be the preferred way to address this for apple devices
> 
> Does it require locking the event to the main thread for all browsers?

Can you explain ? this is not a complex operation in any way and won't be blocking anything.

avaer:
> Can you explain ? this is not a complex operation in any way and won't be blocking anything.

`passive: false` changes when and how the event is handled (it will block the frame loop).

TheoTheDev:
> > Can you explain ? this is not a complex operation in any way and won't be blocking anything.
> 
> `passive: false` changes when and how the event is handled (it will block the frame loop).

without 'passtive: false' it won't work. it's the way it works.
no not-passive - no fix.

avaer:
Ok, I guess we can't fix this in that case.

PR #2742, opened 2022-04-04
Resolves #2593 UI: Updated Setting panel fade-in/out.

### LINKED ISSUE #2593: Settings panel fade

TheoTheDev:
Updated fade in/out animation of Setting panel.
Now it slides same as 'character-select' panel, but half time faster to keep short timing.

Math formula for Settings panel opacity/position fade in/out is now the same.


0reo (commented):
this looks good for sending to QA

alisaad673:
Fading in look good but fading out doesn't seems to have any animation here unlike the character panel. Can you look into it? @TheoTheDev 

https://user-images.githubusercontent.com/35926530/161827306-14d5b3c1-af55-4bd9-ad14-4dd94fe0c3d2.mp4

TheoTheDev:
@alisaad673 checking, will ping in a bit about it.

TheoTheDev:
@alisaad673 So actually the animation math/time was correct.
But cause it's beziers and both, opacity and position were changed it was making an impression it just disappears when fading-out.

So as a solution just removed opacity from transition params and now change only the position exactly the same as it is for CharacterPanel.

alisaad673:
Yep, The fade out is working fine now. Seems good to go.

0reo (approved):

avaer:
It should probably be coming from the right. Let's not waste days on these issues.

PR #2762, opened 2022-04-05
Refactor Animaton System.

gonnavis:
Try to fix: https://github.com/webaverse/app/issues/2456

Accomplished first step: 
Modified skeleton to let it can use threejs' animation classes: `AnimationMixer` and `AnimationAction`.
So can use them to easily implement features such as transitions later.


avaer:
It’s fine with me to use the animation mixer but I’m not sure we will be able to keep the same performance and features with it. (I would love to be wrong)

It is important that we are able to premultiply the ik into the animations and separate the bones for top and bottom so we can blend the walking with actions, as well as do 7-way walk blends of the quaternions for foot states, which I don’t think three animation mixer is capable of.

If we can make that work it would be great, but I don’t know how that is possible yet.

gonnavis:
My main goals now are to realize the abstracted `action` concept to make logic simple, and to realize the feature that can transition between actions.
I tried to make our own `Action` and `Mixer` classes, but then changed my mind to try to use threejs' first.
I'll try to investigate more on `premultiply` and `7-way walk blends` and other existing codes.

I think maybe the best solution in the end is to make our own `Action` and `Mixer` classes, I'll keep trying.

gonnavis:
Still tried using XState with `THREE.AnimationAction` first, maybe just as testing direction.

Accomplished combo chain feature:
Click once: Play combo1.
Click second time when combo1 playing: Play combo1 then combo2.
Click third time when combo2 playing: Play combo1, combo2 then combo3.

And prevent movement when play combo, which most action game do.

https://user-images.githubusercontent.com/10785634/162247060-5efee869-f0a4-457f-bbbc-7695de3dbd16.mp4

avaer:
Looks like the animations for running have regressed significantly above.

avaer:
Can we close this branch as it is moving in a regressive direction?

gonnavis:
Ok, got it. I mainly doing it as a testing direction.

Issue #2760, opened 2022-04-05
Shader recompilations


avaer:
When doing world hops on the street, we preview the scene up ahead with a LODded version.

Ideally this lodded version is rendered with the same materials + shader settings as the final scene, so that switching is seamless when you decide to go there.

However I identified several reasons this is not happening. If we can fix these then we will have seamless world transitions, as well as reducing the app load time significantly.

# Party avatars must be baked into preview

If you don't include the current avatar in the rendered preview scene, then your avatar's new scene materials will not be compiled during the render phase of the world preview. That means a hitch when you enter the world.

# Depth pass renders with a different fog setting

The depth pass used for SSAO and other things uses its own scene, which does not push the scene render settings, causing things like the `scene.fog` to potentially be out of sync when it comes time to get the program cache key. This does not affect correctness since the depth pass uses a shader that doesn't care about this setting, but it results in a completely pointless hitch + performance degradation since we use 2x the shaders. By fixing this we might be able to halve the GPU load time of the app.

# Depth pass does scene render without pushing rendersettings

Similar to the previous issue, but it turns out the depth pass uses the main scenes as well, and also fails to wrap with a rendersettings set. This is a bug.

# Light/shadow counts

The pistol object, for example, changes the scene light count for its muzzle and explosion flashes. This means that the shaders are invalidated. I think we can get around this by setting a minimum amount of lights/shadows , and then as long as there aren't more than that we won't need to recompile shaders. We could also dole out lights using API rather than using `new THREE.Light`.

avaer:
# totum Scene type template is not actually `.isScene`

This results in the fog settings not being picked up when we try to render() it, even though we attempt to latch these objects onto the scene. This is probably a bug.

The easy way around this is to not render the totum scene itself, but an actual scene above it, set with the totum scene's render settings.

PR #2764, opened 2022-04-06
Multi-world loading

avaer:
Support and debugging for multiple worlds loaded into the same instance, allowing traversal across them.

Enables the weba:// URL protocol, for link geolocation in the overworld. Instead of this being handled by default metaverse modules, it is now handled first-class in the new `overworld.js` manager, which currently uses hardcoded layout. This leverages the full scene pause, auto-lod, and hotswap mechanisms we already had.

Attention is paid to making sure transitions are efficient and seamless, including eliding shader recompilations due to rendersettings being collected and applied incorrectly in many sub-renders.

Depends on https://github.com/webaverse/totum/pull/100


avaer:
Should fix all of the shader recompilation cases in https://github.com/webaverse/app/issues/2760.

avaer:
I think I managed to get every shader compiling correctly cross-world, with a seamless world transition, by using custom program cache key injection. THREE.js does not support this by default, so we need a vendorized edit but it seems to work.

avaer:
The main culprits in the above recompilations were the depth passes. I think with this shader sharing I might have also improved rendering performance a bit, though that was not the goal.

Issue #2153, opened 2022-01-04
App: add FPS limiter


avaer:
For people who cook eggs on their computer and prefer soft boiled to hard boiled.

avaer:
This should also be really useful for testing for frame rate dependent behavior. We can make it a hotkey to increase/decrease the frame rate.

PR #2765, opened 2022-04-06
Resolving #2153  Added ability to set frame-lock.

TheoTheDev:
Added ability to increase / decrease max FPS.
By default disabled.
By pressing shift & -  -> decrease the limit.
By pressing shift & + -> increase the limit.
Also updated stats FPS with render FPS & loopFPS:
image
+ in console you can see what exact limit was set [after changing the limit].


src/components/app/App.jsx
```diff
@@ -141,6 +141,36 @@ export const App = () => {
 
     }, [] );
 
+    useEffect( () => {
+
+        const handleKeyUp = ( event ) => {
+
+            if ( event.which === 187 && event.shiftKey ) {
```
0reo, line 148: please add a comment with which key this is for readability

src/components/app/App.jsx
```diff
@@ -141,6 +141,36 @@ export const App = () => {
...
+
+                app.renderLimitMs = app.renderLimitMs || 16 * 2; // if not set then set to 30fps
+                app.renderLimitMs = Math.floor( 1000 / ( 1000 / app.renderLimitMs + 5 ) ); // +5fps, can be not 100% accurate cause of update/render frame shifted [loop time too long]
+                console.log( `Render limit increased to ${ app.renderLimitMs }ms (${ Math.floor( 1000 / app.renderLimitMs ) }fps)` );
+
+            } else if ( event.which === 189 && event.shiftKey ) {
```
0reo, line 154: please add a comment with which key this is for readability

0reo (changes requested):
Few readability changes would be good

TheoTheDev:
Added comments as requested + added check is typing inside of input [in this case don't apply].

0reo (commented):
should be good for testing @alisaad673

src/components/app/App.jsx
```diff
@@ -141,6 +141,43 @@ export const App = () => {
 
     }, [] );
 
+    useEffect( () => {
```
avaer, line 144: This is not a good place to hook in the functionality. It sounds more like a job for `ioManager`.

TheoTheDev: Updated and moved key handling to ioManager.

alisaad673:
Decreasing and increasing the limit seems to be working fine but render fps don't go be beyond 60 fps even if the limit is set at 100fps. Is it dependent on refresh rate of screen?

https://user-images.githubusercontent.com/35926530/163037376-e26e4acb-55bc-4556-b91d-17f602216418.mp4

TheoTheDev:
> Decreasing and increasing the limit seems to be working fine but render fps don't go be beyond 60 fps even if the limit is set at 100fps. Is it dependent on refresh rate of screen?
> 
>  bandicam.2022-04-13.00-15-51-603.mp4

Usually max refresh rate is 60fps. But if your device has low performance and you get max 40fps for example you can not go beyond this physical limit of computation power of your device.

avaer:
>Usually max refresh rate is 60fps.

This should be determined by the browser and how fast it fires `requestAnimationFrame`. We should not hardcode any values here.

TheoTheDev:
> > Usually max refresh rate is 60fps.
> 
> This should be determined by the browser and how fast it fires `requestAnimationFrame`. That is not necessarily 60FPS.
> 
> (note that we should not hardcode any values here)

so anything to be changed ?
you want to replace 2 * 16.6 with a constant and that's it ?
anything else ?

webaverse.js
```diff
@@ -71,6 +71,10 @@ export default class Webaverse extends EventTarget {
   constructor() {
     super();
 
+    this.lastRenderTime = 0;
+    this.renderFrame = 0;
```
avaer, line 75: These are port variable names. Is it the number of frames?

io-manager.js
```diff
@@ -570,6 +570,30 @@ ioManager.keydown = e => {
       debug.toggle();
       break;
     }
+    case 187: { // + & shift [plus - not on num-pad]
```
avaer, line 573: Rather than this keyboard hack, this should be in the settings.

avaer:
This isn't likely to ship.

PR #2770, opened 2022-04-07
Calculate normals for marching cubes in wasm

koreex:
* Wasm is updated to calculate normals for marching cubes
* And the return of `marchingCubes()` is modified to include normals


avaer:
This PR seems like binary mystery meat. We have no idea where this binary came from.

PR #2774, opened 2022-04-07
Calculate normals for marching cubes in wasm

koreex:
* Calculate normals for marching cubes in wasm
* Add JS interface to get normals from wasm
* `geometry.wasm` is built from the following commit of `app-wasm` repo
https://github.com/koreex/app-wasm/commit/66be2e638c9fa7261b844f69eddbd57e5d457dcc


avaer (changes requested):
This PR computes incorrect zero normals: https://github.com/koreex/app-wasm/commit/66be2e638c9fa7261b844f69eddbd57e5d457dcc#diff-24d992872e730949874247c3e9237cb86249c0b94966ec50984ba967e300e570R403

PR #2778, opened 2022-04-08
use clone for all object material

tcm390:
I used material.clone() for all of the objects. And compress all the texture images.


avaer:
I'm not sure why the textures are here btw.

PR #2780, opened 2022-04-08
Make wear component imply wear on activate

avaer:
Makes the wear component in charge of wearing an app on activate. Previously this was handled in the individual type handlers, but it really should be handled by the relevant component.

Some apps and type templates which were handling this need to be updated to not perform a double-wear.


avaer:
Related: https://github.com/webaverse/totum/pull/101

PR #2781, opened 2022-04-08
Add sonic boom material freeze

avaer:
Replaces https://github.com/webaverse/app/pull/2778

This PR cleans up the sonic boom metaverse module to not create extra materials, and also freezes its material shaders so that they do not need to be recompiled when the scene render settings change (they do not depend on lights).


avaer:
Note the use of `Material.freeze`, which is newly added functionality.

See https://github.com/webaverse/three/pull/5

PR #2785, opened 2022-04-08
Resolves #2126 app world editor uses rotation scn uses quaternions

### LINKED ISSUE #2126: App: World editor uses rotation, scn uses quaternions
The `Z` world editor menu makes editing scenes (`scn`s) a lot easier, the one thing that is tricky is the world editor allows editing a rotation vector, but `scn` uses quaternions and does not seem to use a `rotation` field.

TheoTheDev:
Finished with WorldPanel update.
Fixed WorldObjects item inspector panel.
Updated UI/UX, added objects preview.
Added Euler/Quaternion switch.

https://user-images.githubusercontent.com/1478866/167295188-adacdb0e-93c2-45c0-a15e-4ee2f793c52b.mp4

ps: Physics 'objects' scale not getting updated when changing the scale via WorldPanel menu. This feature is done in other PR [#2908 ].

pps: Updated fonts and added object preview in the bottom when selecting one item from the list [in case it's media].

ppps: Improved a lot NumberInput UX now:
- Value is formatted properly [prevents non number values].
- Has nicer increase/decrease [left/right arrows] btns.
- Added max precision value [min 0.001].
- Scale can not be set to Zero value [but not as frustrating as it was when it was hard to input 0.__ cause 0 was not inputing, and now formatted runtime/after change, but not blocking].
- Added support of UpArrow / DownArrow keys to increase decrease value while input focused.


src/components/general/world-objects-list/WorldObjectsList.jsx
```diff
@@ -124,6 +180,40 @@ export const WorldObjectsList = ({ setSelectedApp, selectedApp }) => {
...
+    useEffect( () => {
+
+        if ( ! selectedApp ) return;
+
+        selectedApp.position.set( px, py, pz );
+        selectedApp.scale.set( sx, sy, sz );
+
+        if ( rotationMode === 'euler' ) {
+
+            selectedApp.rotation.set( rex, rey, rez, 'XYZ' );
```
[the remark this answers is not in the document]
avaer: That makes sense to me.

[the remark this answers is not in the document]
TheoTheDev: Added separate select for this.

alisaad673:
This seems to be working fine, One issue is with scale, The physics of the object gets distorted with scale as shown in the video below. Not sure if this is related to this PR. Otherwise it looks fine.

https://user-images.githubusercontent.com/35926530/163034998-5fb1fa42-08fc-4fec-8d23-f79d56c600ce.mp4

TheoTheDev:
> This seems to be working fine, One issue is with scale, The physics of the object gets distorted with scale as shown in the video below. Not sure if this is related to this PR. Otherwise it looks fine.
> 
>  Untitled.35.mp4

hmm.. @avaer do I need to update also update params of physical part of mesh ?
I tried trivial 
        // const physicsObjects = selectedApp.getPhysicsObjects();
        // const physicsObject = physicsObjects[0] || null;
        // physicsObject.position.set( px, py, pz );
        // physicsObject.quaternion.copy( selectedApp.quaternion );
        // physicsObject.scale.set( sx, sy, sz );
        // physicsObject.updateMatrixWorld();
but this doesn't work, any suggestion for proper object physics collider update ?

TheoTheDev:
btw, this PR also covers the issue #2680

avaer:
First, scaling physics requires completely remaking the physics object, and probably needs PhysX support to delete the old object and replace it with a new one. Basic research: https://documentation.help/NVIDIA-PhysX-SDK-Guide/Shapes.html

avaer:
Second, there is `physicsManager.setTransform` which will push the current transform of the physics object to the PhysX scene.

But if you look in our code it actually ignores the scale because that is simply not allowed in PhysX (an entirely new geometry must be made): https://github.com/webaverse/app-wasm/blob/173487538e98927b321e82ea08349d7c178b6718/physics.cc#L569

The reason for this is that scale completely changes how you need to think about the spatial occupation of the object in the octree.

avaer:
Related: https://github.com/webaverse/app/issues/2819

avaer:
Also, the UI on this panel (written by me!) is terrible.

It was meant to be a placeholder because I didn't have time to fill in the details, but it can be made into a proper design with an hour of work.

The fact that so many PRs have been based on the placeholder and it's still not fixed is a bit of a case of misplaced priorities.

TheoTheDev:
> Also, the UI on this tokens panel (written by me!) is terrible.
> 
> It was meant to be a placeholder because I didn't have time to fill in the details, but it can be made into a proper design (like the equipment panel) with an hour of work.
> 
> The fact that so many PRs have been based on the placeholder and it's still not fixed is a bit of a case of misplaced priorities.

There was zero mentions about the UI style and that it was placeholder and needs to be changed.
If it need to be updated any info about how exactly it should be changed ?

avaer:
Yeah, it's not related to this PR, it's just worrying that it's not being fixed.

avaer:
> If it need to be updated any info about how exactly it should be changed ?

Yes, it should match the equipment panel using the same preview spritesheet subsystem.

TheoTheDev:
> > If it need to be updated any info about how exactly it should be changed ?
> 
> Yes, it should match the equipment panel using the same preview spritesheet subsystem.

Ok, got it. Previews obviously should be there, atm it's only placeholder images.

About this physics object recreate.
I suspected this.
but this there any example of readding this runtime ?
I done it easily in an app js script loaded by totum

But in react there's no normal access to physics adding/removing, only to current object physics object.
And usePhysics not working and causing error 'needs to be called in render fn' or smth like this.
I'm sure I can at some point get this working,
but I will most likely save a lot of time if you point me direction [looked through all files with search for smth key thins but found nothing atm].

avaer:
Instead of `usePhysics`, if you are not an app (e.g. you are a React UI), you can `import physics-manager.js`.

This is almost the same thing, it just won't track your physics objects in the app since you do not have one.

avaer:
If you want to get the app for a physics object, that is in the world the API is here: https://github.com/webaverse/app/blob/4823fb4507ba8f53935b0a99712fb536b87eddd4/app-manager.js#L344

```
import {world} from '../world.js';
...
const app = world.getAppByPhysicsId(physicsObject.physicsId);
```

avaer:
However I do not recommend re-adding a physics object yourself. There could be many bindings that the app has to the old object so most likely destroying it will break the app.

The correct way to handle physics scale is probably to write a low-level recreator which keeps the same physics ID but swaps out the underlying object with the rest of the APIs not noticing the change.

avaer:
More comments after testing:

- We should add all 6 Euler combinations
- Switching the eulers needs to update the values in the box via conversion (can be done with `THREE.Euler`). Otherwise the numbers displayed are actually not correct when you change the order bropdown. Even if changing them will correct them eventually, that is not good UI. The alternative (IMO worse) approach is to actually rotate the object to account for the new order but I don't think that is better.

avaer:
- We should step the eulers on keystroke up/down by some multiple of PI, or set the Eulers in degress. But the current behavior where you increase by `1` rad seems not useful to any user.

TheoTheDev:
- Updated UI to change euler rotation inputs order when changing Euler order.
- Added previews for objects.
- Made custom steps for inputs [smaller for roation -> 0.1rad]

avaer:
The quaternions are not normalized, creating crazy transforms.
The eulers are still in radians with invalid increment.

avaer (changes requested):
Still lots of changes needed.

alisaad673:
Issues:
1- In the fields we can type in while changing (XYZ) positions, scale etc. The number keys meddle with the typing. For example, when we try to type 0,9,8 etc it plays sound along with typing.We need to restrict playing of sound while typing the numbers.

https://user-images.githubusercontent.com/35926530/166656753-c3ca21e1-3ece-433b-81a2-90b5a0afd4c1.mp4



2- The physics mesh isn't being updated by changing the scale, but that is i guess being solved in another PR?
3-In addition to the issue no. 2, also when we decrease any of the (x,y,z) to 0 while changing scale, the object loses physics entirely as it can be seen in the video below.

https://user-images.githubusercontent.com/35926530/166656796-5210254d-4230-4c90-96e8-813c2e563be4.mp4



Other thing seem to be working fine, The UI looks good, the previews are not here for all items but that is a general issue not related to this, the positions adjustment is fine, Quaternions/rotations seems fine

TheoTheDev:
Fixed sound triggering issue.
Yes, physics fixed in other PR.
And added check for scale [zero is not acceptable value now].
@alisaad673

TheoTheDev:
Improved NumberInput UX [check bottom of PR description].

object-screenshotter.js
```diff
@@ -25,6 +25,7 @@ export const screenshotObjectUrl = async ({
   start_url,
   width = 300,
   height = 300,
+  canvas
```
avaer, line 28: Why pass the canvas? We already set the size on the internal one.

metaversefile-api.js
```diff
@@ -1093,7 +1092,7 @@ export default () => {
     app.name = m.name ?? (m.contentId ? m.contentId.match(/([^\/\.]*)$/)[1] : '');
     app.description = m.description ?? '';
     app.appType = m.type ?? '';
-    app.contentId = m.contentId ?? '';
+    // app.contentId = m.contentId ?? '';
```
avaer, line 1095: Which cases does this break on?

src/components/general/spritesheet/Spritesheet.jsx
```diff
@@ -11,8 +12,10 @@ export const Spritesheet = ({
     enabled,
     size,
     numFrames,
+    animated = true,
```
avaer, line 15: IMO there shouldn't be a non-animated spritesheet. Something that is not animated should use something other than a spritesheet.

src/components/general/spritesheet/Spritesheet.jsx
```diff
@@ -11,8 +12,10 @@ export const Spritesheet = ({
...
+    background
```
avaer, line 16: Can we accomplish this with CSS at the parent instead?

story.js
```diff
@@ -59,6 +60,9 @@ export const listenHack = () => {
   let currentFieldMusic = null;
   let currentFieldMusicIndex = 0;
   window.document.addEventListener('keydown', async e => {
+
+    if ( gameManager.inputFocused() ) return;
```
avaer, line 64: There are better ways to route this. It's not the fault of this PR but there should be only one window `keydown` listener.

TheoTheDev: should this be changed in this PR or different ?
I think different, cause this one is already quite big and this thing as you told not very related to the PR.

src/components/general/spritesheet/Spritesheet.jsx
TheoTheDev: yup, removed.

object-screenshotter.js
TheoTheDev: discussed this in discord and added checks.
if canvas passed width/height should not be.

metaversefile-api.js
TheoTheDev: All items which should have PreviewImage were not working cause of this [content Id was wrong] cause of this.

avaer: Then it sounds like there is a bug in that part of the code, but this isn't the right fix.

story.js
avaer: Ok, but then in that case this line should also be removed from this PR.

metaversefile-api.js
avaer: Perhaps the module is not exposing its `contentId`

TheoTheDev:
Improved 'story' keyUp event handling.
- Removed direct window keyup event listener.
- Move key handling to app and using io-handler for this now.

metaversefile-api.js
TheoTheDev: Maybe.
But we need to stop. We shouldn't fix all existing bugs in this PR.
This definitely should be a separate investigation and PR.

story.js
TheoTheDev: Missed this comment.
Already introduced the proper event handling to this PR.

avaer (changes requested):
This still breaks contentId handling.

avaer (approved):

avaer:
Looks like this caused a regression in spritesheets.

The animation on these icons goes crazy.

image

avaer:
Reverting this...

avaer:
Reverted: https://github.com/webaverse/app/pull/2785

Issue #2049, opened 2021-12-20
app: Sword combat animation while walking/running


amnaarshed4224:
Sword combat animation while trying to walk run is really odd and awkward.  Needs much improvement

https://user-images.githubusercontent.com/35926530/146821486-9ff0f0e5-1d34-4e73-9402-160863bb4e8f.mp4

https://user-images.githubusercontent.com/35926530/146821601-cf364066-b46a-4185-860e-be4f5b9da699.mp4

.

amnaarshed4224:
Sword combat animation while being in crouch mod is odd too.

https://user-images.githubusercontent.com/35926530/150186192-6ca7ca0e-0833-4320-bd62-67f7830d65f9.mp4

PR #2791, opened 2022-04-08
Fix: Sil/Sword combat and aim animation problem while walking/running.

gonnavis:
Fix: https://github.com/webaverse/app/issues/2049

Preliminarily fixed this issue. The main changes are [these lines](https://github.com/gonnavis/app/blob/024ab7af67a4024850858fde423e563778f7599a/avatars/animationHelpers.js#L957-L987).

Basic idea is to not apply sword combo animations to legs when walking.

Now not very awkward, but still has many todos:
- Lerp combo and walk/run based on `moveFactors.idleWalkFactor`.
- Check bones by whether shoulder and arm instead of `isTop`.
- Fix when crouch.
- Test with silsword.

https://user-images.githubusercontent.com/10785634/162491430-2eaba968-e9f6-4734-9a2b-4f7c07135ae8.mp4


avaer:
Looks good! Let's also make sure we are blending correctly for aim mode too (right click mouse).

gonnavis:
Crouch:

https://user-images.githubusercontent.com/10785634/162566473-ccc437cd-238b-427a-ace9-defa9cf6eb85.mp4

Aim:

https://user-images.githubusercontent.com/10785634/162566477-7961c29d-c463-4b8d-adf8-01079701aa2f.mp4

Hello @avaer , is it ok?

With silsword, it's still looks a little awkward, but maybe fix in another PR.
image

gonnavis:
Noticed that silsword's aim is different from sword's aim.
And fixed it when walk/run too.

https://user-images.githubusercontent.com/10785634/162581916-24545a06-047e-462c-b7df-ce4b9cdad080.mp4

avaer:
> Noticed that silsword's aim is different from sword's aim logic/codes.
Yes, the animations are not related at all.

Re: the above, I think the walk/run needs to be blended into the arms when aiming the sword or else it will look wrong/stiff.

gonnavis:
Blended walk/run into arms when aiming.

https://user-images.githubusercontent.com/10785634/162791542-116c2c03-1b7b-4acc-aa06-0f13efb53339.mp4

gonnavis:
textametron fixed these animations. https://github.com/webaverse/app/issues/2796

image

avaer:
Awesome work, looks so much better!

avatars/AnimationMapping.js
```diff
@@ -1,72 +1,74 @@
 class AnimationMapping {
-  constructor(animationTrackName, boneName, isTop, isPosition) {
+  constructor(animationTrackName, boneName, isTop, isPosition, isArm, isCombo) {
     this.animationTrackName = animationTrackName;
     this.boneName = boneName;
     this.isTop = isTop;
+    this.isArm = isArm;
+    this.isCombo = isCombo;
```
avaer, line 7: `IsCombo` is not a very good name for this. Combo of what?

avatars/animationHelpers.js
```diff
@@ -189,7 +189,16 @@ async function loadAnimations() {
   const uint8Array = new Uint8Array(arrayBuffer);
   const animationsJson = zbdecode(uint8Array);
   animations = animationsJson.animations
-    .map(a => AnimationClip.parse(a));
+    .map(a => {
+      const animationClip = AnimationClip.parse(a)
+      if (animationClip.name === 'One Hand Sword Combo.fbx') {
+        trimClip(animationClip, 0.4, animationClip.duration - 0.5);
```
avaer, line 195: This is not the place to fix animations. They should be fixed at bake time.

avatars/animationHelpers.js
```diff
@@ -939,30 +949,44 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
...
             const v2 = src2.evaluate(t2);
+            localQuaternion2.fromArray(v2);
 
-            const idleAnimation = _getIdleAnimation('walk');
-            const t3 = 0;
-            const src3 = idleAnimation.interpolants[k];
-            const v3 = src3.evaluate(t3);
+            if (moveFactors.crouchFactor === 0) {
+              if (isCombo) {
+                // do nothing: localQuaternion2 already pure combo animation
```
avaer, line 956: I really do not recommend depending on locals like this. Locals are reused and constantly change their meaning, so this comment will be irrelevant and wrong very quickly.

If the local is meaningful enough to be talked about in a comment, it should be given an actual name like:

```
const headQuaternion = localQuaternion2;
```

avatars/animationHelpers.js
```diff
@@ -939,30 +949,44 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
...
             _clearXZ(localVector2, isPosition);
 
-            const idleAnimation = _getIdleAnimation('walk');
-            const t3 = 0;
-            const src3 = idleAnimation.interpolants[k];
-            const v3 = src3.evaluate(t3);
-            localVector3.fromArray(v3);
+            if (moveFactors.crouchFactor === 0) {
+              if (isCombo) {
+                // do nothing: localVector2 already pure combo animation
```
avaer, line 977: Same locals comment as above.

avatars/animationHelpers.js
```diff
@@ -1060,6 +1085,11 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
 
         _handleDefault(spec);
 
+        if (!activeAvatar.unuseAnimation) return;
```
avaer, line 1088: It's a bad idea to ever early return in this code. It makes the overall flow of the logic (which is the most important part of this code) impossible to analyze at a glance.

avatars/animationHelpers.js
```diff
@@ -1203,6 +1233,17 @@ export function getFirstPersonCurves(vrmExtension) {
   }
 }
 
+export function trimClip(clip, startTime, endTime) {
```
avaer, line 1236: This should be in the animations baker.

avaer (changes requested):
Code cleanups needed.

avaer:
I tested it and for the most part it works great.

Additional issues:

- There is an animation for aiming the sword, but it is much too fast.
- There is no animation for unaiming the sword.
- The running with the sword/items seems very unrealistic. The sword is enormous, but the spine doesn't move when running, like the item is made out of paper. We should find a way to animate the spine more (maybe increase the intensity of the quaternion?) so that there is some weight to the object held.

These issues can probably be fixed with tweaking the animations.

avaer:
I'd love to merge this PR either way. The only blocker is some code cleanups, particularly moving the animations baking to the offline script rather than at load time.

avatars/animationHelpers.js
gonnavis: Moved `trimClip()` from runtime to baker. https://github.com/webaverse/app/pull/2791/commits/f056c61d8d6daa123fee7d9c482dc0ed83211ada

avatars/animationHelpers.js
gonnavis: Moved `trimClip()` from runtime to baker. https://github.com/webaverse/app/pull/2791/commits/f056c61d8d6daa123fee7d9c482dc0ed83211ada

avatars/AnimationMapping.js
gonnavis: `isCombo` bones are arms with Chest and UpperChest. I feel hard to name so named it `isCombo` because of mainly used for combo animation.

Used `isArm` instead, no need `isCombo` now, they are no very big differences. https://github.com/webaverse/app/pull/2791/commits/2f87848f5ea98c59458af068775dc12796a543d6

avatars/animationHelpers.js
gonnavis: Added `isLastBone`, so can make sure `activeAvatar.unuseAnimation` changed in last bone, then no need `return` any more, is it ok? https://github.com/webaverse/app/pull/2791/commits/f5392da41f4c5682b670a371559fe154cfeea226

avatars/animationHelpers.js
gonnavis: Same as above.

gonnavis:
> Additional issues:
> 
> * There is an animation for aiming the sword, but it is much too fast.
> * There is no animation for unaiming the sword.
> * The running with the sword/items seems very unrealistic. The sword is enormous, but the spine doesn't move when running, like the item is made out of paper. We should find a way to animate the spine more (maybe increase the intensity of the quaternion?) so that there is some weight to the object held.
> 
> These issues can probably be fixed with tweaking the animations.

Ok, I'll keep trying to fix these issues, maybe in subsequent PRs.

gonnavis:
> There is an animation for aiming the sword, but it is much too fast.

Tested that 300ms too slow, so increased to 200ms. https://github.com/webaverse/app/pull/2791/commits/8dfa702f48be3c40d7f77ccd1872bc22151ec91a

gonnavis:
> There is no animation for unaiming the sword.

Generally handle end-action transition, works well for jumping and unaiming.

avatars/animationHelpers.js
```diff
@@ -1007,49 +1029,54 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
...
-          dst
-            .sub(localVector2.fromArray(v3))
-            .add(localVector2.fromArray(v2));
         }
+
+        _handleEndActionTransition(spec);
       };
     } else if (activeAvatar.unuseAnimation && activeAvatar.unuseTime >= 0) {
-      return spec => {
+      return (spec, isLastBone) => {
```
avaer, line 1079: Hm this isn't a good approach. If something should happen after the last bone then the caller should do that, not the quaternion handler.

avaer: This might actually affect performance negatively to pass extra arguments to all bones for all frames.

avaer:
Hm I don't know if the jump landing is better. It seems to take too long for some cases.

Let's not mix this up with the sword PR. We can do actual landing animations which will look much better.

Also not related to this PR: the spine sword problem is more obvious when you try to attack while running.

gonnavis:
>Hm I don't know if the jump landing is better. It seems to take too long for some cases.
Let's not mix this up with the sword PR. We can do actual landing animations which will look much better.

Reverted.

gonnavis:
>Also not related to this PR: the spine sword problem is more obvious when you try to attack while running.

Ok, I think I should add back `isCombo` bone group which includes arms and Chest and UpperChest.
What's the name is better, `isUpperBody`, `isAttack`, `isSword`?

avatars/animationHelpers.js
gonnavis: Ok, @avaer , is this proper? https://github.com/webaverse/app/pull/2791/commits/dd0ea6f9ef7307133eebae33b92dd4a2cd7c07d9

animations-baker.js
```diff
@@ -35,6 +35,16 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
     'Crouched Sneaking Left.fbx',
     'Crouched Sneaking Right.fbx',
   ];
+  const trimClip = (clip, startTime, endTime) => {
```
avaer, line 38: This is ok to ship, but I think we can fix this in the animation itself in the long term.

Maybe add a comment here/open an issue to trim this animation in Blender.

gonnavis: Ok, added comment first. https://github.com/webaverse/app/pull/2791/commits/eb1b8da991c6d5c3894dcf9f2f7714ad632ae524
I also want to trim sword animation into three pieces, if this pr can be merged https://github.com/webaverse/app/pull/2804 .
I'll request artist asset afterwards.

gonnavis:
Resolved conflicts.

gonnavis:
Reverted fixing of useAnimation not stop problem, https://github.com/webaverse/app/pull/2791/commits/4dc46253d14685d506d1cb918da017bdbfaf823b
because it's not very related to this PR and not a serious problem,
may try in another PR afterwards.

avaer:
I still think this PR is far too robotic when walking and attacking. I think _overall_ it is not better than what we had before, it's just a different set of animation problems we are trading for.

At least what we had before was very dynamic even if all of the bones were going crazy.

If we need walk-attack and run-attack custom animations to solve this and make powerful swipes that do not look like a robot (especially in the spine), I think that is a valid plan.

gonnavis:
Got it! I'll contact artist and request resource.

gonnavis:
Only improve silsword with `aimAnimation` first, reverted changes of `useAnimations`.

Before:

https://user-images.githubusercontent.com/10785634/166246670-ca6f8912-0ee3-4e88-a773-5b79648ba70c.mp4

After:

https://user-images.githubusercontent.com/10785634/166246683-a4048b0f-57b4-4f09-a865-91c63c07e515.mp4

PR #2794, opened 2022-04-09
Add grunt effect and envelope

tcm390:
1. Upload 2 new sounds. (eating and drinking)
2. Load 'actionFiles' and 'voicePack name' in voice-pack-voicer.js so that we can access actionFiles and voicepack name from localPlayer.
3. Add a function call 'playGrunt' in character-sfx.js, it would load the voice from 'localPlayer.voicePack.actionFiles', and play the voice pack by 'localPlayer.voicePack.audioBuffer'. So in totum app, we could play the grunt like this `localPlayer.characterSfx.playGrunt('angry', index);`
4. Add jump grunt and gasp grunt in character-sfx.js by using exponential falloff selection
This is the jump result (each character has 6 jump grunt)

https://user-images.githubusercontent.com/60634884/162553640-2e5027c7-2a35-4965-ad16-ef88b667b135.mp4

This is the gasp result (each character has 7  gasp grunt, but shishi just has 2 now)

https://user-images.githubusercontent.com/60634884/162553739-b24a6091-13a2-41db-8690-7b4c1ffb8fd7.mp4

5. Add envelope in avatar.js. By default, if we connect the sounds to the character's mouth, it will move the mouth based on the audio volume. If we set the envelope like this `this.player.avatar.envelope(attack, decay, sustain, release);`, It would move the mouth according to the envelope
6. moving mouth when eating and drinking in character-sfx.js by using envelope. This is the result:

https://user-images.githubusercontent.com/60634884/162554380-7c866843-07fe-41de-80b8-f5d69e019282.mp4


avaer:
> 2 of the gulp samples have a click at the end, I don't know if this is a sample or coding issue

If the samples do not have this issue then it's a sound compilation bug which should be debugged.

avaer:
Who is driving this? Can we close it?

avaer:
Closing since nobody seems to be working on this.

PR #2795, opened 2022-04-09
Scene cruncher

avaer:
Adds marching cubes crunching for the map scene previewer.


avaer:
This is nearly done. There is a perfect voxelization from the top down for the scene, but it could be better with a triplanar capture.

avaer:
I think we can merge this and continue in another PR.

Issue #2798, opened 2022-04-09
Aim will cause player move slow when facing +Z axis.


gonnavis:
Moving toward -Z axis, _not aim_ vs _aim_, the same speed:
![image](https://user-images.githubusercontent.com/10785634/162592777-33ed187c-35eb-46b1-ade9-6e5b74bb35db.png)

Moving toward +Z axis, _not aim_ vs _aim_, aim slow:
![image](https://user-images.githubusercontent.com/10785634/162592841-5aa712f9-17a6-49c2-83e8-d007bf80b0fc.png)

Cause obvious problem when using `idleWalkFactor` to lerp animations, _-Z_ vs _+Z_:
![image](https://user-images.githubusercontent.com/10785634/162592915-0bba6d6f-819b-4914-9baa-c160fabd84bf.png)

I confirmed that it's the consequence of `io-manager.js -> game.menuAim();` which added an ( even empty ) `aimAction` to player.
But I searched `getaction('aim` globally, have not found any codes which will cause speed changes yet.

gonnavis:
Should caused by wrongly `gameManager.isMovingBackward(); // = true` in this situation.

gonnavis:
Really `aimAction` related, not accessed through `getaction('aim')` but `hasAction('aim')`.

avaer:
Thank you for the detailed investigation!

PR #2799, opened 2022-04-10
Fix wrong `isMovingBackward()`.

gonnavis:
Fix: https://github.com/webaverse/app/issues/2798


avaer (approved):
I tested and it's ok.

PR #2800, opened 2022-04-10
Resolves #2527 Fixed Edit mode controls and a bit of refactor and clean up.

### LINKED ISSUE #2527: Edit mode (tilde grabbing)
Edit mode (`tilde`/`~` to enable) was working a few weeks ago but now it's broken; it doesn't grab objects correctly. When you grab right now, objects will teleport to origin and stay there.

It was previously working like Fortnite, though the matrix inversion was always wrong and caused spatial skipping when clicking the object, which is a bad experience that we can fix.

When you click the object it should stay in place but attach itself to your character. Moving camera should move the object to the new transposed location based on the camera difference. We also have keys like `E` (rotate left), `R` (rotate right), `V` (toggle colliders) which should be respected. to control the grabbed object.

The code for this lives in `game.js` but that's not good abstraction. This control mode should be its own file, communicating with `io-manager.js` to get keys and the other managers (like physics) to effect the correct transforms when in edit mode.

TheoTheDev:
Fixed editor controls for objects.
Moved key handling from io-manager to editor-mode rect UI component.
Not sure still about the issue mentioned about inverse matrix & whether I need to more more logic to UI from game.js [as for me is enough].

https://user-images.githubusercontent.com/1478866/162615659-0bca84fa-f7ba-4de9-ae04-faa4df7edc95.mov

### DIFF
--- a/src/components/editor-mode/index.jsx
+++ b/src/components/editor-mode/index.jsx
@@ -1,15 +1,85 @@
 
-import React from 'react';
+import React, { useEffect } from 'react';
+
+import game from '../../../game';
 
 import { MagicMenu } from './magic-menu';
 import { SceneMenu } from './scene-menu';
 import { Inspector } from './inspector';
+import { registerIoEventHandler, unregisterIoEventHandler } from '../general/io-handler';
 
 //
 
 export const EditorMode = ({ selectedScene, setSelectedScene, selectedRoom, setSelectedRoom }) => {
 
     const multiplayerConnected = !! selectedRoom;
+
+    //
+
+    useEffect( () => {
+
+        const handleKeyPress = ( event ) => {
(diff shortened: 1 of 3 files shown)


src/components/editor-mode/index.jsx
```diff
@@ -13,6 +16,73 @@ export const EditorMode = ({ selectedScene, setSelectedScene, selectedRoom, setS
 
     //
 
+    useEffect( () => {
```
avaer, line 19: This should be removed from React and handled in io manager.

avaer:
I also see incorrect update glitching in the above video (off by a frame or something).

avaer:
Closed because the entire approach seems wrong.

PR #2804, opened 2022-04-10
One click play a full attack animation & continue combo.

gonnavis:
Made one click play a full attack animation for sil/sword, instead of holding mouse button.
i.e.: `endUse()` by the end of attack animation, instead of `mouseup` event.

And tried to not affect other weapons' functions, such as bow and pistol.

https://user-images.githubusercontent.com/10785634/162634388-d34e0e80-e71d-4f6d-af2b-df8a0fccc351.mp4

Todo:
- Combo chain.


avaer:
This it a good improvement, but the whole attack system needs to be reworked.

I think attacking should put your character in an "aggro" state for a bit, and they eventually return to normal/idle if you wait or move away. And the animations when doing this should be smooth.

avaer:
Somewhat related, in the past we have enabled legs IK when aiming, so only the top body twists but the legs stay in place. This is much more natural than sliding feet.

The IK handles the case of adjusting the feet when you rotate too far.

This is also how a lot of games, including VRChat, work.

gonnavis:
Thanks! I'll try this solution afterwards.

gonnavis:
Accomplished the same combo chain feature as this PR https://github.com/webaverse/app/pull/2762#issuecomment-1091917401
>Click once: Play attack1.
Click second time when combo1 playing: Play attack1 then attack2.
Click third time when attack2 playing: Play attack1, attack2 then attack3.
If stopped combo at any attack, will start from attack1 next time.

But this time done based on the existing action/component system, instead of refactored to using XState with THREE.AnimationAction.

Todo: 
- Currently used a lot of variables to control the states, need to use action/component instead. And maybe can reduce the variables needed and simplify the logic.
- Currently the four attack animations all restore to `sword_idle` pose in the end, which cause the combo not coherently, there will be pauses in between. I asked textametron to help give me a full combo animation, and I can use code to split it into four 
consecutive attack animations.

https://user-images.githubusercontent.com/10785634/162969216-dcd89e4c-f6c3-43f5-a9d7-3b343ed8af90.mp4

I want to achieve something like this eventually:

https://user-images.githubusercontent.com/10785634/162969302-c7790a12-cecf-4439-9cc8-e3cedbcdc4b5.mp4

gonnavis:
Used textametron's new continuous combo animation, and using codes splitted it to four attack animations.
Now no `sword_idle` and glitch in between attack animations.

https://user-images.githubusercontent.com/10785634/163167164-16fd1904-12ef-4b76-b4b8-9964fb6d53fc.mp4

gonnavis:
According to [this principle](https://github.com/webaverse/app/issues/2532#issuecomment-1093879822), I prevented all logics and state changes in `avatar.js` and `animationHelpers.js`.
Put all states in `useAction` instead.

But I have to let other codes know the end of the animation, so I eventually chose to dispatch `animationEnd` event from avatar.

gonnavis:
The change of this PR caused silsword combo animations only play sound of the first attack.
So updated silsword repo. Sibling PR: https://github.com/webaverse/silsword/pull/5

avatars/animationHelpers.js
```diff
@@ -72,6 +72,7 @@ let fallLoop;
 // let swordSideSlash;
 // let swordTopDownSlash;
 let hurtAnimations;
+let lastF;
```
avaer, line 75: This is a new global variable. How does this work when many avatars are animating?

character-controller.js
```diff
@@ -1084,8 +1087,17 @@ class LocalPlayer extends UninterpolatedPlayer {
...
       this.characterFx.update(timestamp, timeDiffS);
 
+      const useAction = this.getAction('use');
+      if (useAction?.needEndUse) {
+        gameManager.menuEndUse(); // must before updateInterpolation
+      }
+
       this.updateInterpolation(timeDiff);
 
+      if (useAction?.needContinuCombo) {
```
avaer, line 1097: Typo. "continue"

character-controller.js
```diff
@@ -40,6 +40,7 @@ import {
   defaultPlayerBio,
 } from './ai/lore/lore-model.js';
 import {makeId, clone, unFrustumCull, enableShadows} from './util.js';
+import gameManager from './game.js';
```
avaer, line 43: I don't think the character controller should ever contact the `game` manager.

There is something wrong with the way this code is structured.

gonnavis: Added a new state `needResetUseTime` and modified checking fn of `actionInterpolants.use` . https://github.com/webaverse/app/pull/2804/commits/5c8c30ac3422242d4cb9481b6c05e28fbcd50769
Now can run `menuEndUse` and `menuStartUse` both before `updateInterpolation()`,
No need the order: `menuEndUse` -> `updateInterpolation()` -> `menuStartUse`.

So now can put them out of `updateAvatar()`.
Moved from `character-controller.js` to `game.js` directly, no need import `game` any more. https://github.com/webaverse/app/pull/2804/commits/f9f0982f7079bf395234a63632da5fc311cde8cd

avatars/animationHelpers.js
gonnavis: Reverted fixing of unuseAnimation not stop problem, https://github.com/webaverse/app/pull/2804/commits/29b8e11ef0d51b86f7008a50a6c835a694263148
because it's not very related to this PR and not a serious problem,
should try in another PR afterwards.

gonnavis: Made it as local variable instead of global. https://github.com/webaverse/app/pull/2804/commits/0ccd4c2681f2b379ce1782b3ee7b01d20f4550e4

avatars/avatars.js
```diff
@@ -393,8 +393,9 @@ const _makeDebugMesh = (avatar) => {
 
 
 
-class Avatar {
+class Avatar extends EventTarget {
```
avaer, line 396: Seems like `Avatar` is not emitting events so it does not need to be an `EventTarget`?

avaer: I see the helper emits events: https://github.com/webaverse/app/pull/2804/files#diff-94f53fcba17d684d2a2411861f461b371bd86726f85126685b76dbf4fa63717eR1186

IMO this isn't good though, if `Avatar` needs to be emitting events it should be an `Avatar` method doing it.

character-controller.js
```diff
@@ -877,7 +879,17 @@ class UninterpolatedPlayer extends StatePlayer {
     this.actionInterpolants = {
       crouch: new BiActionInterpolant(() => this.hasAction('crouch'), 0, crouchMaxTime),
       activate: new UniActionInterpolant(() => this.hasAction('activate'), 0, activateMaxTime),
-      use: new InfiniteActionInterpolant(() => this.hasAction('use'), 0),
+      use: new InfiniteActionInterpolant(() => {
+        const useAction = this.getAction('use')
+        if (useAction) {
+          if (useAction.needResetUseTime) {
```
avaer, line 885: I don't think it's legal to change the action.

This will not be picked up when the action is replicated over network.

avaer: If there is some state that needs to be tracked with the action then I think that state needs to live with whatever class is handling the action.

avaer:
I tested and the behavior is good, but the code needs some more data cleanup.

avaer:
Somewhat related to this PR: we should find a way to "accelerate" these swings a bit when we combo into the next move, so the swing rate more closely matches the buttons the user is pressing.

At the moment it feels a little bit too laggy (but it's definitely better than the instant behavior we had before!).

character-controller.js
gonnavis: Add actions `needEndUse`, `needContinueCombo` and `needResetUseTime`, instead of modify `useAction`. https://github.com/webaverse/app/pull/2804/commits/c0838b61112912eb0bb2f448c71c626d33917a7f
Is this proper?

avatars/avatars.js
gonnavis: Movec dispatchEvent() from animationHelper.js to avatars.js. https://github.com/webaverse/app/pull/2804/commits/fba731c118e8aa17ef0e813a8b7dcfe6405bd17d

avaer:
I merged master in here because it was very outdated.

avaer (changes requested):
It is impossible to test this without changing the code because it hits `debugger`.

```
if (!localPlayer.hasAction('needResetUseTime')) {
          localPlayer.addAction({
            type: 'needResetUseTime'
          });
          debugger;
        }
```

gonnavis:
Sorry forgot clean it. Let me delete it immediately.

avaer:
What is the `needResetUseTime` action? What is the action the player is performing?

character-controller.js
avaer: No I don't think so. Actions are for actions the player is doing. These are not actions.

avaer:
Looks like this needs a lot more work to make the logic understandable.

gonnavis:
@avaer cleaned. https://github.com/webaverse/app/pull/2804/commits/1a7eb5c3d6233ffb6be80643d77100d7600e9792

animations-baker.js
```diff
@@ -260,6 +270,22 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
 
       animations.push(animation);
     }
+
+    const swordStrikeContinuousAnimation = animations.filter(a => a.name === 'sword_strike_continuous.fbx')[0];
+    let animation;
+    animation = trimClip(swordStrikeContinuousAnimation.clone(), 0, 20 / 30);
```
avaer, line 276: This is really not good to depend on. We made these animations. Why is this not fixed on the art side?

avaer (changes requested):
Should fix the animations rather than clipping them.

gonnavis:
> What is the needResetUseTime action? What is the action the player is performing?

Because of I treat continuous attacks ( combo ) as a single `useAction`, so need to reset `useTime` to zero before start each attack.

I think I can also always treat one attack as a `useAction`, but maybe it'll make logic more complicated and less compatible with transition systems afterwards.

animations-baker.js
gonnavis: Used pre-splitted attack animations by textametron, instead of `trimClip()`. https://github.com/webaverse/app/pull/2804/commits/2b5c1c4c12ea016b5b4e2d4e0b15a7147e17ed3e

character-controller.js
gonnavis: @avaer Yeah, they are just states.
How about make them as player's state variables? https://github.com/webaverse/app/pull/2804/commits/0a325db0917c3c6a0876f0c3aef482fb62e2f6fb

gonnavis:
Closed in favor of https://github.com/webaverse/app/pull/3097

Issue #2788, opened 2022-04-08
Avatars shader coalesce


avaer:
Most avatars used have duplicate materials (20-50) and they are terribly optimized, since the author is assuming that the game engine will fix the inefficiencies. This is indeed the case usually.

We already do this in Webaverse for avatar crunching, but it seems we can get rid of 80-90% of the avatar materials even in high quality mode. For example, a typical avatar of ours only has 3 distinct materials despite having 25 materials. At a first approximation, crunching these would result in an 80% reduction in avatar render time with no render cost (though we would need to crunch this avatar in offscreen engine before wearing it, which seems reasonable to me).

https://github.com/webaverse/app/blob/master/avatar-cruncher.js

avaer:
Note that this technique can be easily applied to pets, GLBs.

avaer:
```
ms = [];
os = [];
vrmApp.traverse(o => {
	if (o.isMesh) {
		m = o.material[0] ?? o.material;
		if (m.visible !== false) {
			ms.push(m);
			os.push(o);
		}
	}
});
/* is = ms.map(m => m.map?.image).filter(i => !!i); is.forEach(i => document.body.appendChild(i)); */

set = new Set();
for (let i = 0; i < os.length; i++) {
	k = renderer.getProgramCacheKey(rootScene, os[i], ms[i]);
	set.add(k);
}
console.log(set);
```

avaer:
Can be added to THREE.js to visualize the cached programs:

```
this.getProgramCacheKey = (scene, object, material) => {
		const currentRenderState = renderStates.get( scene );
		currentRenderState.init();
		
		const lights = currentRenderState.state.lights;
		const shadowsArray = currentRenderState.state.shadowsArray;

    const parameters = programCache.getParameters( material, lights.state, shadowsArray, scene, object );
		const programCacheKey = programCache.getProgramCacheKey( parameters );
		return programCacheKey;
	};
```

PR #2805, opened 2022-04-10
[WIP] Avatar optimizer

avaer:
Crunches all avatars by default, where the crunch does not affect the quality (same materials, redundant draw calls).

Should improve avatars performance for high quality mode.

Fixes https://github.com/webaverse/app/issues/2788


avaer:
image

I still have to fix kyojin mode on the texture remap but the performance improvement with this PR is quite large, and no quality should be lost.

avaer:
image

It seems the bug is in the toon shader, because using a different material fixes the mapping.

avaer:
Indeed it appears that changing the texture on the material has no effect.

Probably the VRM toon shader code is doing something incorrect with accessors, or is not picking up map updates. Do we need to hack the raw uniforms object?

avaer:
image

It's almost perfect now after porting the texture constants (filter settings, etc). One strand of hair is still broken.

avaer:
I think the black band errors are due to UVs that are out of `[0, 1]` range, so they overflow the atlas.

Such UVs do exist in the models.

avaer:
The solution to the above could be clamping of the UVs.

avaer:
Clamping UVs to 2% - 98% fixed the above.

avaer:
This is almost done.

Now, the optimized model needs to replace the main one, and ideally all of this optimization is done in a worker (that applies to the other optimization modes as well).

avaer:
When the optimized avatar is the main one, we should also test with many different avatars to make sure we are not breaking any cases.

avaer:
This fixes https://github.com/webaverse/app/issues/2788

avaer:
The performance increase for this is about 10 FPS per each avatar in the scene (note that this is still full avatar quality).

avaer:
The last thing that needs to be done here is integrating a `VrmQualityManager` which is responsible for shimming and generating all of the avatar proxy meshes for the 3 different avatar crunchers. This can speed up loading, too, if we use a placeholder avatar icon while avatars are loading asynchronously in the `OffscreenEngine`.

avaer:
Resurrecting this...

PR #2812, opened 2022-04-12
2759 procedurally generating geometry with the marching cubes technique

koreex:
- Changes
  * Generate a terrian using multiple workers
  * Use 4 workers
  * Generate buffers in JS side and WASM generates one chunk only
  * Dynamic allocation is done in JS

- Result
  * It shows no frame drops (some partially presented frames)
  * The speed of generating terrain is fast enough for the movement of the character
  
- Notes
  * It's using a default Three.js material now.
  * It still has files for material (`idTech.js` and `toonMaterial.js`). ARYA is working on materials from scratch. But those file have remained as references right now. So no need to review.

https://user-images.githubusercontent.com/58914494/162902015-698be2b7-a2ab-4480-b96f-d16c5a99924d.mp4


koreex:
* In the previous version, we generated one chunk for every update call.
* Now we call worker functions to generate chunks at once
* We stage requests of generating chunks when there is no idle worker
* After a worker finishes their job, it checks the staged request queue
* So the usage of worker is 100%, when there is still chunks needed to generate

PR #2822, opened 2022-04-15
2527 edit mode tilde grabbing

TheoTheDev:
Fixed editor controls for objects.
Moved key handling from io-manager to editor-mode rect UI component.
Not sure still about the issue mentioned about inverse matrix & whether I need to more more logic to UI from game.js [as for me is enough].

Cause of branch rename newer commits were not added to it somehow.
Key bind approach is changed back as it was to be handled in the engine.

Frame glitch mentioned by @avaer is in progress of investigation.

### DIFF
--- a/game.js
+++ b/game.js
@@ -508,15 +508,15 @@ const _gameUpdate = (timestamp, timeDiff) => {
   const _updateGrab = () => {
     // moveMesh.visible = false;
 
+    const renderer = getRenderer();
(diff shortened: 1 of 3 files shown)


avaer:
This is not a good PR description. It refers to several different conversations and people without explaining any code or rationale behind the changes. Anyone who was not part of those conversations will probably have no idea what is going on here.

avaer:
I'm going to close this because it is not clear what this code is anymore.

PR #2831, opened 2022-04-15
Avatar poses

avaer:
Adds new pose animations and interface for the character/login UI.

This is accessed in 2 ways:

## Character panel

image

## Q key (hold)

image


avaer:
This could still be improved with smoother transitions but it should be good enough to play with.

It uses the regular action system so should work in multiplayer for free.

PR #2835, opened 2022-04-18
Grunt voice and mouth movement

tcm390:
1. Upload 2 new sounds. (eating and drinking sound without the click at the end now)
2. Load 'actionFiles' and 'voiceActor' in voice-pack-voicer.js so that we can access actionFiles and voiceActor name from localPlayer.
3. Add a function call 'playGrunt' in character-sfx.js, it would load the voice from 'localPlayer.voicePack.actionFiles', and play the voice pack by 'localPlayer.voicePack.audioBuffer'. So in totum app, we could play the grunt like this `localPlayer.characterSfx.playGrunt('angry', index);`
4. Add jump grunt and gasp grunt in character-sfx.js by using exponential falloff selection
This is the jump result (each character has 6 jump grunt)

https://user-images.githubusercontent.com/60634884/163752351-ba5c3789-97ab-4a09-9e18-516acf7f27de.mp4

https://user-images.githubusercontent.com/60634884/163752360-d02a99c4-6880-4c7c-bd31-94f23e9b26d1.mp4

https://user-images.githubusercontent.com/60634884/163752368-7da1a0de-0891-4839-99bf-691bfdf77e70.mp4

https://user-images.githubusercontent.com/60634884/163752376-fee9347f-1c92-451a-aaf1-a98167b12aee.mp4

https://user-images.githubusercontent.com/60634884/163752388-b9dccdbc-8fa9-46f5-ae07-25b92d4f4ce6.mp4

5. This is the gasp result (each character has 7 gasp grunt)

https://user-images.githubusercontent.com/60634884/163752466-d6d94729-ed44-4c58-9009-9daf863e8698.mp4

https://user-images.githubusercontent.com/60634884/163752471-936dfd6d-e060-4448-9e60-c5fecaab2e97.mp4

https://user-images.githubusercontent.com/60634884/163752485-961e02eb-3d6e-474d-97cd-172de241f4fe.mp4

https://user-images.githubusercontent.com/60634884/163752501-f0a92496-5a02-478f-a145-15a9952d6393.mp4

https://user-images.githubusercontent.com/60634884/163752509-af2c6524-bb3e-479f-abda-fc1bee52238d.mp4

5. Add mouth movement function in avatar.js. So we can control the mouth movement manually by `this.player.avatar.setMouthMoving(attack, decay, sustain, release);`, It would move the mouth according to the envelope
6. control mouth movement for eating and drinking in character-sfx.js by using envelope. This is the result:

https://user-images.githubusercontent.com/60634884/163752721-52fe2e85-0c22-4289-bf4e-7beac5a30f37.mp4


Arthur-Acha:
1 - Gasps/pants being played after running to display that the character is tired. We have a great variety of types of gasps/pants between the 5 voice actors. Shishi and Griffin are the most "standard" type of gasp/pant, the other ones have their own characteristics. Having a variety of characteristics for the grunt is great because we can cover a bigger amount of preferences among the players.

2 - Jumps are using a selection of Attack Grunts to display the character is doing some sort of effort, attack grunts were used because they fit well on this action and we had a good amount of them to "spare". The ones used on jump will not be used on attack action to avoid audio duplicates (in case both actions are executed at the same time)

3 - For eat and drink a single sample was used for each of them. I went on this direction because it adds consistency to the action. The sample alone sounded dull, so I changed it's pitch 2 semitones up and 2 semitones down for the 2 actions, a total of 3 samples for each action.

character-sfx.js
```diff
@@ -55,6 +55,37 @@ const _getActionFrameIndex = (f, frameTimes) => {
   return i;
 };
 
+const jumpGruntIndex={
+  'shishi':[{index:2,nonce:0},{index:3,nonce:0},{index:4,nonce:0},{index:5,nonce:0},{index:6,nonce:0},{index:7,nonce:0}],
```
avaer, line 59: We should not be hard coding any voice actors here.

This must be declared from the voice pack, most likely with a naming scheme.

character-sfx.js
```diff
@@ -73,10 +104,28 @@ class CharacterSfx {
     this.preQ=new THREE.Quaternion();
     this.arr = [0, 0, 0, 0];
 
+    this.runStep = 0;
+    this.lastRunningTime = 0;
 
     
     this.oldNarutoRunSound = null;
   }
+  selectVoice(voicer) {
```
avaer, line 113: This is duplicated code from the voice pack voicer.

It seems the voice pack voicer should be making this decision?

avatars/avatars.js
```diff
@@ -1883,7 +1891,72 @@ class Avatar {
...
+        break;
+      }
+      case null: {
+        _handleMouthMovementNull();
+        break;
+      }
+    }
+
+  }
+  setMouthMoving(attack, decay, sustain, release){
```
avaer, line 1952: This should be abstracted better. The avatars implementation is already too complicated, especially since we already have a volume tracking implementation and this is kind of a second one.

I would say that maybe this belongs in `CharacterSfx` since that is most related but since this does not actually generate audio, maybe it belongs in a new class like `CharacterBehavior`?

avaer (commented):
This looks good overall, but I think we need to move the code into better locations.

character-sfx.js
avaer: If we really need to, I think it's better to duplicate files in the voice pack.

E.g. copy the same file to `Jump1.wav` so that we can do `/jump/i.test(fileName)`

avatars/avatars.js
avaer: Then, `CharacterBehavior` could calculate and set the volume on the avatar at player -> avatar binding update time.

alisaad673:
1- The sound for jump seems to work fine, But with the case of dash "shift+double W" the jump sound doesn't plays.

https://user-images.githubusercontent.com/35926530/164051133-da51dae5-4969-493f-8f5f-eb747baff4b7.mp4

2-Similarly, sound for gasp doesn't plays with dash.

https://user-images.githubusercontent.com/35926530/164051550-73e8ffd3-a3bc-4cc1-bba2-3873d2fe561c.mp4

3-We are not doing gasp sound on walk right?? This is being played on run only?

Else looks and sounds fine. the mouth movement seems in sync with sfx.

tcm390:
For 1 I fix it, could you test it again?
For 2, 3, 4, @Arthur-Acha could you reply these?

alisaad673:
Yep 1 is fixed now.

Arthur-Acha:
> Hi @alisaad673 For 1, I fix it. Could you test it again? For 2, 3, 4, @Arthur-Acha could you reply these?

I think you have a good point and thanks for sharing it @alisaad673 , we should implement both gasps and jumps during and after the naruto run, but there will be a VO that will be played when you start running, so jumps VOs should only be able to trigger after the naruto run VO ends, this way we don't have conflict.
@tcm390 can you implement this?

tcm390:
Hi @alisaad673 I add the jump and gasp to narutoRun. Also integrated the new narutoRun grunt. Could you test it again?
This is the result:

https://user-images.githubusercontent.com/60634884/164359695-8099ee0a-2a0b-4008-a2f3-61081b2b9da4.mp4


https://user-images.githubusercontent.com/60634884/164359704-bc1d5f09-b091-4881-8449-bb8cb377a7c5.mp4

character-sfx.js
```diff
@@ -97,6 +99,10 @@ class CharacterSfx {
       if (this.player.avatar.jumpState && !this.lastJumpState) {
         const audioSpec = soundFiles.jump[Math.floor(Math.random() * soundFiles.jump.length)];
         sounds.playSound(audioSpec);
+        // play jump grunt 
+        if(this.player.voicePack!==null && !this.player.hasAction('fly') && this.player.avatar.velocity.y<0){
```
avaer, line 103: I think a better condition for this is `jumping && !lastJumping`.

avaer: `jumping = this.player.hasAction('jump')`

avaer: Instead of checking `this.player.voicePack!==null` I think it's better to let the `playGrunt` handle it.

Otherwise every other `playGrunt` call needs this check too (and it currently doesn't so I'm assuming it is bugged).

character-sfx.js
```diff
@@ -225,21 +243,24 @@ class CharacterSfx {
...
+              }
+            });
+
+            this.narutoRunTrailSoundStartTime = timeSeconds;
           }
         }
 
+        // if naruto run play more than 2 sec, set runStep > 10
+        if(timeSeconds -  this.narutoRunStartTime > 2){
+          this.runStep = 11;
```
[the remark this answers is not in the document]
avaer: Looks like this is for the gasp case. In which case we should probably set some sort of `willGasp` instead of hacking the step counts.

character-sfx.js
```diff
@@ -308,6 +346,67 @@ class CharacterSfx {
...
+    // control mouth movement with audio volume
+    if (!this.player.avatar.isAudioEnabled()) {
+      this.player.avatar.setAudioEnabled(true);
+    }
+    audioBufferSourceNode.connect(this.player.avatar.getAudioInput());
+
+    if(this.oldGrunt){
+      this.oldGrunt.stop();
+    }
+    this.oldGrunt=audioBufferSourceNode;
```
avaer, line 406: For cleanliness purposes I would wait for the `end` of the buffer and then set to `null` to clear it. Otherwise you are infinitely storing the last grunt long after it is done, which is a memory leak.

voice-output/voice-pack-voicer.js
```diff
@@ -23,9 +23,35 @@ function weightedRandom(weights) {
 }
 
 class VoicePack {
-  constructor(syllableFiles, audioBuffer) {
+  constructor(syllableFiles, actionFiles, voiceActor, audioBuffer) {
```
avaer, line 26: `voiceActor` looks dead.

voice-output/voice-pack-voicer.js
```diff
@@ -44,7 +70,24 @@ class VoicePack {
...
+      (async () => {
+        const res = await fetch(indexUrl);
+        let j = await res.json();
+        j = j.filter(({name}) => /^actions\//.test(name));
+        return j;
+      })(),
+      loadAudioBuffer(audioContext, audioUrl),
+    ]);
+
+    const name=audioUrl.substring(
```
avaer, line 85: IMO this should be a regex.

voice-output/voice-pack-voicer.js
```diff
@@ -23,9 +23,35 @@ function weightedRandom(weights) {
...
+      return {
+        name,
+        offset,
+        duration,
+        nonce: 0,
+      };
+    });
+  }
+  selectVoice(voicer) {
+    // the weight of each voice is proportional to the inverse of the number of times it has been used
```
avaer, line 41: This is copy paste, so it seems like this algorithm should be a generic util.

voice-output/voice-pack-voicer.js
```diff
@@ -23,9 +23,35 @@ function weightedRandom(weights) {
 }
 
 class VoicePack {
-  constructor(syllableFiles, audioBuffer) {
+  constructor(syllableFiles, actionFiles, voiceActor, audioBuffer) {
     this.syllableFiles = syllableFiles;
+    this.actionFiles = actionFiles;
     this.audioBuffer = audioBuffer;
+    this.voiceActor = voiceActor;
+    this.actionVoices = actionFiles.map(({name, offset, duration}) => {
```
avaer, line 31: The entire voicepack is one set of files right? If true then we should only pass one set of files here.

Maybe `syllableFiles` should just be `files` and has a bad name?

character-sfx.js
```diff
@@ -308,6 +346,67 @@ class CharacterSfx {
     };
     _handleFood();
   }
+  playGrunt(type, index){
+
+    let voiceFiles, offset, duration;
+    switch (type) {
+      case 'pain': {
+        voiceFiles = this.player.voicePack.actionVoices.filter(f => /Pain/.test(f.name));
```
avaer, line 354: I would `/pain/i` to make the regex case-insensitive.

character-sfx.js
avaer: Same as here: https://github.com/webaverse/app/pull/2835/files#diff-62b89e43ead1a304ba4d7c9b9053d6da5a55029f73da7664fa3a73e30d2993e2R249

character-behavior.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+            case null: {
+                _handleMouthMovementNull();
+                break;
+            }
+        }
+    }
+    setMouthMoving(attack, decay, sustain, release){
+        this.mouthMovementState=null;
+        this.player.avatar.manuallySetMouth=true;
```
avaer, line 74: I think this should be set in the player-avatar-bindings, not here.

avaer (commented):
Recommending a lot of code cleanups.

character-sfx.js
tcm390: In this situation, If the oldGrunt is playing, should I stop the `oldGrunt` and play the new grunt like this:

https://user-images.githubusercontent.com/60634884/164500226-d1a36774-c6e9-463c-adbc-ffb75f45126f.mp4

Or should I ban the new grunt if the `oldGrunt` hasn't finished like this:

https://user-images.githubusercontent.com/60634884/164500394-02099364-d063-4997-8758-e9567944608e.mp4

voice-output/voice-pack-voicer.js
tcm390: removed

voice-output/voice-pack-voicer.js
tcm390: removed, since we don't need `voiceActor` now

voice-output/voice-pack-voicer.js
tcm390: move `selectVoice` into util.js.

character-sfx.js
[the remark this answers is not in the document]
tcm390: created `this.willGasp`

character-behavior.js
tcm390: move `mouthMovementState` into player-avatar-binding.js

voice-output/voice-pack-voicer.js
tcm390: change `syllableFiles` to `files`

tcm390:
code cleanups: https://github.com/webaverse/app/pull/2835/commits/3f4cfa0c927590eb7f9a5a3b55b86c98d2b24449

tcm390:
code cleanups: https://github.com/webaverse/app/pull/2835/commits/b439d317a410e775771a516beb6897f329919e32

character-sfx.js
avaer: Hm ok, but I still think it should check jump in addition to this. Otherwise if the user is hit and pushed back as a result, setting velocity up but not jump, the grunt will be played (but that would be wrong).

character-sfx.js
avaer: I think the first one.

character-sfx.js
avaer: Maybe we should add extra metadata to the `jump` action, like an extra key `trigger`:

`trigger === 'jump'` means they pressed the jump button
`trigger === 'fall'` means they got to jump mode because they fell off a cliff (and don't play the grunt in that case)

avaer: For that solution we just need to add the extra metadata whenever we add the `jump` action.

avatars/avatars.js
```diff
@@ -1649,16 +1651,16 @@ class Avatar {
           if (aIndex !== -1) {
             morphTargetInfluences[aIndex] = volumeValue;
           }
-          if (eIndex !== -1) {
+          if (eIndex !== -1 && !this.manuallySetMouth) {
```
avaer, line 1654: I think this is not correct? We are already preventing `.volume` update (https://github.com/webaverse/app/pull/2835/files/7ff71e32fb96c6dc5e2fe598f47d4101047202dc#diff-fcb2093ac11d9b04a179d3dedbef5c0aae4d14b7f71db90e483becca4170b7ccR1922) so why do we also prevent the update here?

voice-output/voice-pack-voicer.js
avaer: That's good, but then why is it still here?

tcm390: Sorry, I’m not sure what do you mean. You mean selectVoice still in voice-pack-voicer?

avatars/avatars.js
tcm390: Because if I don't ban the 'e', 'i', 'o', 'u' index, it would change the mouth figure:

https://user-images.githubusercontent.com/60634884/164544866-93743cfe-dc47-4806-ba92-569c1042aa63.mp4



https://user-images.githubusercontent.com/60634884/164544918-937a9c79-36e3-41e5-a3a4-4c13027f66d6.mp4


https://user-images.githubusercontent.com/60634884/164544931-08cbdf1b-1923-4127-bb27-86c36ab94229.mp4

character-sfx.js
tcm390: change to first one: https://github.com/webaverse/app/pull/2835/commits/87aef8e77eb6cc546d70ac342de266f40490c990

character-sfx.js
tcm390: Hi @avaer so now I just add the `'jump' trigger` when player press the jump button in `ioManager`. https://github.com/webaverse/app/pull/2835/commits/8432ba8eff814b980e83a1e6dbbaab34d6559452

but I don't know where should I add the `'fall' trigger`. Should I add it in the z.mjs?
unknown22

avatars/avatars.js
avaer: Then it sounds like instead of banning it, the ADSR code/player binding should be setting the correct vowels instead.

E.g. it can simple enable only the `A` and set the rest to `0`.

character-sfx.js
```diff
@@ -97,6 +102,11 @@ class CharacterSfx {
       if (this.player.avatar.jumpState && !this.lastJumpState) {
         const audioSpec = soundFiles.jump[Math.floor(Math.random() * soundFiles.jump.length)];
         sounds.playSound(audioSpec);
+        
+        // play jump grunt 
+        if(this.player.hasAction('jump') && this.player.getAction('jump').trigger=='jump'){
```
avaer, line 107: `===` instead of `==`.

Also incorrect spacing.

character-sfx.js
```diff
@@ -259,6 +284,20 @@ class CharacterSfx {
...
     };
     _handleNarutoRun();
+    
+
+    const _handleGasp = () =>{
+      if(timeSeconds - this.lastRunningTime > 2 && !this.player.avatar.narutoRunState && timeSeconds - this.narutoRunFinishTime > 0.5){
+        this.runStep = 0;
+      }
+      // if running step is more than 10
+      if(currentSpeed < 0.1 && timeSeconds - this.lastWalkTime > 0.01 && (this.runStep > 10 || this.willGasp) && !this.player.avatar.narutoRunState){
```
avaer, line 294: I don't think we should count run steps at all.

We should simply set `willGasp` after a certain amount of time.

character-sfx.js
tcm390: Sorry https://github.com/webaverse/app/pull/2835/commits/242763dc83559799f39139499f6462586c314e8a

character-sfx.js
avaer: That would be more consistent with the naruto run for example, and it would make the gasp not dependent on animation (which is weird).

tcm390: remove `runStep` and just use `willGasp` now: https://github.com/webaverse/app/pull/2835/commits/750d3d0a8e4cf7e31d127fb389f8f282c6d649d6

tcm390: remove `runStep` and play the grunt by `willGasp`: https://github.com/webaverse/app/pull/2835/commits/5cba3697b2848b23ddce62fd62b7632206f4e9a1

avatars/avatars.js
tcm390: move setting vowels in player-avatar-binding https://github.com/webaverse/app/pull/2835/commits/5f4304a25d46c7a70698592c3473b949942639bb

avaer (approved):
Looks good.

voice-output/voice-pack-voicer.js
avaer: Yes, it's fixed now, it just wasn't when I was making the comment.

PR #2841, opened 2022-04-19
User icon in header

avaer:
Adds a new emoting user icon (rendered from the player's avatar) and stats bar to the header, replacing the main icon and character pull tab.

Adds HP/MP/XP/Limit bar synchronized with the character panel.

The intent is to replace the "login bar" as well, since the user icon better represents the currently logged in player's state. However, we can keep a small login button in the top-right.

image


avaer:
This should be good to go.

Issue #2006, opened 2021-12-14
app: crippling while walking


alisaad673:
After CCD update, The crippling of avatar has increased way too much  specially on uneven surfaces, areas with much hurdles and while climbing down the hills. More

https://user-images.githubusercontent.com/35926530/146073619-d1dbe94c-0dfe-4842-9011-1baa48f0e13d.mp4

https://user-images.githubusercontent.com/35926530/146073660-58391e99-b0a5-412d-b251-f5cd3e3132d0.mp4

https://user-images.githubusercontent.com/35926530/146073721-10e12067-d5c1-4dd7-857a-e1a2f00163c5.mp4

.

avaer:
Hat does crippling refer to? The jumping?

amnaarshed4224:
yes. While coming down the slopes and when hurdles come in the way. The jumping trigger, when happens much frequently, It feels like crippling.

0xmetabad:
This is the line to control the sensitivity in the character controller:

https://github.com/webaverse/app/blob/5b364e48e89916807d09398d60733eed64ac222e/character-physics.js#L164-L166

Some other ideas from Discord:

> gonnavis: I use raycast to add tolerance between land & air

> avaer: or we could increase the delay to falling
but better physics should be preferred
but one way to attempt to fix this in physics is to shift the character slightly down more if they are grounded
making them less likely to hop
but it might also cause a slipnslide effect

Also reporting this issue is seen in the silk fountains tunnels, which detracts from the magical experience:

https://user-images.githubusercontent.com/29695350/148236493-75cae11b-a216-46e9-83c1-a9a12b517a92.mp4

avaer:
This is simply the character going to jump state when they are not touching the ground, since they are walking faster than gravity.

It can be fixed by:
1) gluing the character more to the ground using the character controller to push them down (breaks realism but more cartoon-correct)
2) Increase gravity to be higher, at least when grounded (not realistic but would fix the problem)
3) Decrease character move speed (not desirable for gameplay, doesn't fix running)
4) Let the character bounce like a ball but take longer to start falling, increasing the Wile E. Coyote effect, but it might pass under the radar

There are many other options but we can at least test these to see which one is best.

PR #2848, opened 2022-04-20
Prevent unwanted jumping when go down slope.

gonnavis:
Fix: https://github.com/webaverse/app/issues/2006

If player not grounded, move player down a little when and only when he/she is very near to ground.
To prevent unwanted jumping when go down slope.
And not affect intentional jumping, flying, falling.

Before:

https://user-images.githubusercontent.com/10785634/164200309-b77fc64a-a0aa-4201-8610-9250ac8b0d65.mp4

https://user-images.githubusercontent.com/10785634/164200323-f4a42cfa-e3af-4c0d-ba23-9fdc24f52b2c.mp4

After:

https://user-images.githubusercontent.com/10785634/164200356-4e3a965e-d5c9-48a8-85c5-e5f0f8fb49e1.mp4

https://user-images.githubusercontent.com/10785634/164202028-d6c83ef5-d0a4-4ac6-b7b9-af32445c911d.mp4

### DIFF
--- a/character-physics.js
+++ b/character-physics.js
@@ -91,7 +91,22 @@ class CharacterPhysics {
         this.player.characterController.position
       );
       // const collided = flags !== 0;
-      const grounded = !!(flags & 0x1); 
+      let grounded = !!(flags & 0x1); 
+
+      if (!grounded && !this.player.getAction('jump') && !this.player.getAction('fly')) { // prevent jump when go down slope
+        const flags = physicsManager.moveCharacterController(
+          this.player.characterController,
+          localVector3.set(0, -0.06, 0),
+          minDist,
+          0,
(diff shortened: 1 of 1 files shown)


avaer:
After testing I think the behavior was too aggressive, so I:

- reduced the stick constant, to prevent walking being "faster than gravity"
- bailed out the y offset when it misses the ground, to prevent "cliff edge pulls you down"

avaer:
The criteria I'm using for whether the behavior is good is whether walking causes "tripping".

Our character controller is already superhuman speed, so it makes sense to me if running causes tripping.

avaer:
I also broke out the stickiness constant so it can be adjusted in the future.

PR #2850, opened 2022-04-20
reverting camera dist

GitWillG:
When unaiming, you now revert back to the original camera distance.


camera-manager.js
```diff
@@ -5,6 +5,8 @@ import metaversefile from 'metaversefile';
 import physicsManager from './physics-manager.js';
 import {shakeAnimationSpeed} from './constants.js';
 import Simplex from './simplex-noise.js';
+import { lerp } from 'three/src/math/MathUtils';
+import { localPlayer } from './players.js';
```
Loryhoof, line 9: Hmm, that works but I'm not sure if it's intended to be used like that.
You can get the localPlayer like this:
`const localPlayer = metaversefileApi.useLocalPlayer();`

camera-manager.js
```diff
@@ -206,6 +217,25 @@ class CameraManager extends EventTarget {
     }
     return result;
   }
+  
+  //Lerp towards offset for smoother ADS, works both aiming in and out
+  ads(){
```
Loryhoof, line 222: Should probably change "ads" to something more intuitive

camera-manager.js
```diff
@@ -206,6 +217,25 @@ class CameraManager extends EventTarget {
...
+    if (localPlayer.hasAction('aim') && -cameraOffset.z>1){
```
Loryhoof, line 223: `-cameraOffset.z>1` should probably not be hardcoded, could add it as a variable above so it's easier to edit.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
avaer, line 1: Probably shouldn't commit this.

camera-manager.js
avaer: It's fine to import `localPlayer` like this if you are not an app.

game.js
```diff
@@ -1154,6 +1154,10 @@ class GameManager extends EventTarget {
   menuAim() {
     const localPlayer = metaversefileApi.useLocalPlayer();
     if (!localPlayer.hasAction('aim')) {
+      //current ads value is just a hardcoded offset value that seemed appropriate
+      cameraManager.saveaim();
```
avaer, line 1158: Is this missing from this PR?

camera-manager.js
GitWillG: I just entirely removed the dependency on localplayer from within the aim functionality. I only need to pass in the players position once at the start of takeaim anyways

camera-manager.js
GitWillG: This was an oversight while I was testing different distances, but it now uses the inverted value from "maxAim" an already existing variable above

game.js
GitWillG: I actually removed it entirely in a more recent PR

camera-manager.js
```diff
@@ -355,13 +402,23 @@ class CameraManager extends EventTarget {
               localVector.copy(avatarCameraOffset)
                 .applyQuaternion(camera.quaternion)
             );
+
+            //only aims down sight when not in first person mode
+            if (localPlayer.hasAction('aim')){
+            this.takeAim(localPlayer.position);
```
avaer, line 408: Spacing issues.

camera-manager.js
```diff
@@ -206,6 +209,40 @@ class CameraManager extends EventTarget {
     }
     return result;
   }
+  
+  //Lerp towards offset for smoother ADS, works both aiming in and out
+  takeAim(playerPos){
+    if (-cameraOffset.z> (-maxAim.z)){
```
Loryhoof, line 215: No need for parentheses `( )`

camera-manager.js
```diff
@@ -206,6 +209,40 @@ class CameraManager extends EventTarget {
...
+    //we only want to revert to our original z value if there is a descrepancy between our expected zdistance and our actual distance
+    if (zdist !== 0){
+      cameraOffsetZ = lerpNum(cameraOffsetZ, lastCamoffsetz, 0.1);
+      cameraOffsetTargetZ = cameraOffsetZ;
+    }
+    else{
+      //do nothing
+    }
+    cameraOffset.x = lerp(cameraOffset.x, rayVectorZero.x, 0.1);
+    cameraOffset.updateMatrixWorld;
```
Loryhoof, line 243: `updateMatrixWorld` doesn't do anything here

camera-manager.js
```diff
@@ -15,8 +16,12 @@ const localEuler = new THREE.Euler();
 
 const cameraOffset = new THREE.Vector3();
 let cameraOffsetTargetZ = cameraOffset.z;
+//
+const maxAim = new THREE.Vector3(-0.3,0,-2);
 
+//
 let cameraOffsetZ = cameraOffset.z;
+let lastCamoffsetz = cameraOffset.z;
```
Loryhoof, line 24: make it capital Z

camera-manager.js
```diff
@@ -206,6 +209,40 @@ class CameraManager extends EventTarget {
...
+    if (-cameraOffset.z> -maxAim.z){
+      localVector.copy(playerPos).sub(localVector.copy(maxAim).applyQuaternion(camera.quaternion));
+      cameraOffsetZ = lerpNum(cameraOffsetZ, -localVector.z, 0.1);
+      cameraOffsetTargetZ = cameraOffsetZ;
+    }
+    else{
+      //do nothing
+    }
+    cameraOffset.x = lerp(cameraOffset.x, maxAim.x, 0.1);
+    cameraOffset.updateMatrixWorld;
```
Loryhoof, line 224: `updateMatrixWorld` doesn't do anything here either

game.js
```diff
@@ -1154,6 +1154,7 @@ class GameManager extends EventTarget {
   menuAim() {
     const localPlayer = metaversefileApi.useLocalPlayer();
     if (!localPlayer.hasAction('aim')) {
+      cameraManager.saveaim();
```
Loryhoof, line 1157: small detail but try to keep all variable/function names congruent. In this case camelcase

camera-manager.js
Loryhoof: and the spacing in `z>`

avaer:
This doesn't interact very well if you scroll while holding aim. It will just jitter.

avaer:
Due to this PR being out of date from master it cannot be tested (will crash on sprite avatar settings for example).

avaer:
image

msm0823:
As far as the functionality is concerned, it is working fine. The camera does reverts back to its original position while unaiming. Also, the limited transitions seems fine when we tried to scroll while aiming.

A little issue that seems to be here is the aiming point. When we aim at smaller objects, such as a potion we have to aim a little right side to target it actually.

PR #2863, opened 2022-04-21
UI/auth sys

TheoTheDev:
Updated Webawallet logic with latest code from @hamzatrq
And linked everything with UI.
Auth / Auto auth / Login method select / logout done.

Discord auth not working in this PR and will be done in a separate PR.


avaer:
The user bar rendering has broken styles.

image

avaer:
It looks like the "logging in" message is missing.

TheoTheDev:
Fixed css issue with avatar image.
image

TheoTheDev:
https://user-images.githubusercontent.com/1478866/164734273-ff97a244-6a4e-46a0-be36-959751ceabf9.mov

Logging in btn shows all the time it needs. or I'm missing smth ?

core/webawallet/config/config.js
```diff
@@ -0,0 +1,40 @@
+import WebaverseABI from "../abi/Webaverse.json";
+import WebaverseERC1155ABI from "../abi/WebaverseERC1155.json";
+import WebaverseERC20ABI from "../abi/WebaverseERC20.json";
+import WebaverseAccountsABI from "../abi/Accounts.json";
+
+export const config = {
+  sidechainURL: "https://13.57.177.184:8545",
```
avaer, line 7: Do not hardcode any addresses.

Problematic code comes from upstream.

avaer:
The webawallet code is not good so better not include it here. Even if it were usable, it should not be copy pasted here. Copy pasting repositories is not a valid technique unless the repository is being deleted/archived.

avaer:
There is a conflict after other merges.

avaer:
Why does the logging in not show after confirming metamask? It should.

TheoTheDev:
> Why does the logging in not show after confirming metamask? It should, since clicking anything during that time such as logging in again would break the login process.

yes, didn't notice this, since it was changing focus between screens.



> There is a conflict after other merges.

will fix this.



> The webawallet code is not good so better not include it here. Even if it were usable, it should not be copy pasted here. Copy pasting repositories is not a valid technique unless the repository is being deleted/archived.

the repo will be killed after. it's an intermediate place atm.
this code is fully working and seams pretty ok as for me.

core/webawallet/config/config.js
TheoTheDev: and where should it be taken from ?

avaer: DNS.

core/webawallet/config/config.js
```diff
@@ -0,0 +1,40 @@
...
+  authServerURL: "https://auth.webaverse.com",
+  webaWalletURL: "https://auth.webaverse.com/weba-wallet/",
```
avaer, line 9: If this is the weba-wallet URL, then why is it included in this PR?

avaer (changes requested):
Weba-wallet should be removed from this PR.

avaer:
We'll need to redo this repo without weba-wallet (which seems like dead code now).

avaer:
I tried to resurrect this PR but it looks like it has regressed even further and it is no longer capable of logging in at all (just hangs infinitely on "Logging in".

PR #2873, opened 2022-04-22
integrate heal effect to module

tcm390:
integrate heal effect to module, it would be triggered by the action 'cure'


avaer (approved):

metaverse_modules/heal-effect/index.js
```diff
@@ -0,0 +1,426 @@
...
+          const scalesAttribute = flashMesh.geometry.getAttribute('scales');
+          const idAttribute = flashMesh.geometry.getAttribute('id');
+
+          //######################### pixel attribute ##############################
+          const pixelOpacityAttribute = pixel.geometry.getAttribute('opacity');
+          const pixelPositionAttribute = pixel.geometry.getAttribute('position');
+          const pixelScaleAttribute = pixel.geometry.getAttribute('scales');
+          
+          if(localPlayer.hasAction('cure')){
+              localPlayer.removeAction('cure')
```
avaer, line 272: Probably should only remove the action when the cure is done.

avaer: Also this should apply to any player with the action, not just local.

This heal app should not really be deciding things based on actions. Instead the character fx should detect the action and then control this app based on that.

metaverse_modules/heal-effect/index.js
```diff
@@ -0,0 +1,426 @@
...
+                  healMaterial[i].g -= 0.02;
+                  // healMaterial[i].b += 0.02;
+                }
+                else{
+                  healMaterial[i].r = 0;
+                  healMaterial[i].g = 0;
+                  healMaterial[i].b = 0;
+                  app.remove(group);
+                  app.remove(flashMesh);
+                  particleAlreadyInScene=false;
```
avaer, line 409: This isn't the best logic.

This is in the middle of the for loop for several materials, but it is setting app state (multiple times), and also removing apps multiple times.

If you flip the code and put the `for` inside of `if` it will be better.

metaverse_modules/heal-effect/index.js
```diff
@@ -0,0 +1,426 @@
...
+
+          const opacity = new Float32Array(pixelParticleCount * 1);
+          pixelGeometry.setAttribute('opacity', new THREE.BufferAttribute(opacity, 1));
+
+          const scales = new Float32Array(pixelParticleCount * 1);
+          pixelGeometry.setAttribute('scales', new THREE.BufferAttribute(scales, 1));
+
+      }
+      //##################################################### flash material #####################################################
+      const flashMaterial = new THREE.ShaderMaterial({
```
avaer, line 61: If you use `WebaverseShaderMaterial` (`useMaterials()`), then you don't need the log depth fragment chunks at all.

metaverse_modules/heal-effect/index.js
```diff
@@ -0,0 +1,426 @@
...
+                gl_FragColor= vec4(0.0,vScales,0.3,1.0);
+                gl_FragColor.a *= vOpacity;
+                ${THREE.ShaderChunk.logdepthbuf_fragment}
+            }
+        `,
+        side: THREE.DoubleSide,
+        transparent: true,
+        depthWrite: false,
+        blending: THREE.AdditiveBlending,
+
```
avaer, line 205: There are a 3 properties we should add to our materials to make the shaders compile less, if the shader does not depend on lighting:

https://github.com/webaverse/app/blob/master/metaverse_modules/sonic-boom/index.js#L123

avaer:
Almost there, made some suggestions.

metaverse_modules/heal-effect/index.js
tcm390: remove group and flashMesh once: https://github.com/webaverse/app/pull/2873/commits/a9a81a8ff1878f514d9df2da555ef826b24df7eb

metaverse_modules/heal-effect/index.js
tcm390: change material to WebaverseShaderMaterial
(https://github.com/webaverse/app/pull/2873/commits/951a6e6ee293a4dd5fd5424b3ba2f599fe07a3ec)https://github.com/webaverse/app/pull/2873/commits/951a6e6ee293a4dd5fd5424b3ba2f599fe07a3ec

metaverse_modules/heal-effect/index.js
tcm390: add clipping fog lights
https://github.com/webaverse/app/pull/2873/commits/8e6aaeeb50c30349814493b408c403d6f8071805

metaverse_modules/heal-effect/index.js
tcm390: [move detect action to character-fx](https://github.com/webaverse/app/pull/2873/commits/beedaac98cc0c13ab5d25f5f5df4f0bb3ac2e868)

tcm390: [remove localPlayer](https://github.com/webaverse/app/pull/2873/commits/f4c495dbc96c6133e6158df73702a92dd13b67d7)

alisaad673:
I have tested this with infinifruit, It seems to be working fine.

https://user-images.githubusercontent.com/35926530/165385018-db228dc6-c6cf-4689-bbc7-391776961a19.mp4


https://user-images.githubusercontent.com/35926530/165385066-7d28e144-c356-4f24-950c-28c73f6f2e58.mp4

avaer:
I think this PR is too outdated to test, and full of crashes:

```
geometry-texture-atlas.js:4:8: error: No matching export in "util.js" for import "modUv"
    4 │ import {modUv} from './util.js';
```

I can't start the app.

avaer (changes requested):
Looks like it's outdated and crashing.

avaer:
I merged master but it's still crashing.

image

character-fx.js
```diff
@@ -246,6 +247,25 @@ class CharacterFx {
...
+        (async () => {
+          await metaverseModules.waitForLoad();
+          const {modules} = metaversefile.useDefaultModules();
+          const m = modules['healEffect'];
+          await this.healEffect.addModule(m);
+        })();
+        sceneLowPriority.add(this.healEffect);
+      }
+      if(this.player.hasAction('cure')){
+        this.healEffect.playEffect(this.player);
```
avaer, line 263: I think the heal effect should be responsible for playing the animation, not this code.

To bind to the correct player, we can pass the player Id as a component so the app can look it up in the API.

character-fx.js
```diff
@@ -246,6 +247,25 @@ class CharacterFx {
       }
     };
     _updateSonicBoomMesh();
+
+    const _updateHealEffectMesh = () => {
+      if (!this.healEffect ) {
+        this.healEffect = metaversefile.createApp();
+        (async () => {
+          await metaverseModules.waitForLoad();
```
avaer, line 255: I think we can get rid of this. This code should not be running until the modules are loaded.

PR #2877, opened 2022-04-22
Bugfix/master metamask tmp fix

TheoTheDev:
Fixes Metamask login btn.
But:
Discord still doesn't work.
You need to login to metamask each page update.


src/User.jsx
```diff
@@ -108,64 +110,85 @@ export const User = ({ address, setAddress, setLoginFrom }) => {
...
+                await _setAddress( address );
+                // setAddress( address );
+                setLoginFrom( 'discord' );
+                // setShow( false );
 
-                    WebaWallet.waitForLaunch().then( async () => {
+            } else if ( error ) {
 
-                        const { address, error } = await WebaWallet.loginDiscord( code, id );
+                setLoginError( String( error ).toLocaleUpperCase() );
```
avaer, line 130: Why uppercase?

avaer: I realize it is not related to this pr and the problem was already there.

PR #2879, opened 2022-04-22
made damageMesh visibility optional

Loryhoof:
Due to requests I made the damageMesh (damage numbers) optional.

No need to modify the code in the weapon apps, default is set to true (visible) in here.

Can disable points by adding `damageMeshVisible: false` in the `targetApp.hit()` opts


avaer:
IMO the correct fix for this is to add the cinematic UI-less mode.

PR #2882, opened 2022-04-23
Remove `crouchAction` after `useAction`, except when use pistol.

gonnavis:
Using weapon when crouching will cause awkward animations, and not reasonable in some cases, especially for large melee cold weapons, like sword and silsword.
So I want to remove `crouchAction` after adding `useAction`.

But using pistol when crouching seems better and reasonable, so I excluded it.

We can still design special crouching version animations/logics for each weapon afterwards, especially something like bow or knife, and exclude them like what pistol do.
But for now I think it's good to prevent using weapons when crouching.

### DIFF
--- a/game.js
+++ b/game.js
@@ -352,6 +352,9 @@ const _startUse = () => {
         };
         // console.log('new use action', newUseAction, useComponent, {animation, animationCombo, animationEnvelope});
         localPlayer.addAction(newUseAction);
+        if (newUseAction.ik !== 'pistol') {
+          localPlayer.removeAction('crouch');
+        }
 
         wearApp.use();
       }


avaer:
That makes sense to me.

avaer:
I changed my mind on this so will revert.

The animations in battle mode will involve a lot of crouching and attacking, and it looks really good that way. To disable crouch mode because we can't fix the animations seems lame. We should fix the animations.

gonnavis:
But I think we need this limit eventually?
How to imagine/make animations for great sword when crouching?
( Our current sword and silsword are already feel like great sword )

avaer:
I think we can just limit it to a single short crouch attack, or two-part combo. If the player is crouching and they attack, we can freeze them in place (with lerp), so there aren't any moving cases to handle.

gonnavis:
Got it!

PR #2889, opened 2022-04-25
Dual Contouring Seam Generation

aryasalehi:
This is the sibling PR of this : https://github.com/webaverse/app-wasm/pull/46


physx.js
```diff
@@ -1836,7 +1836,58 @@ const physxWorker = (() => {
...
+      positions: positions,
+      normals: normals,
+      indices: indices,
+    }
+  }
+
+  w.createSeam = (x, y, z) => {
+    const ELEMENT_BYTES = 4
+
+    const readBuffer = (outputBuffer, index) => {
```
avaer, line 1871: If this is a common method or should be abstracted and reused.

physx.js
```diff
@@ -1814,7 +1814,7 @@ const physxWorker = (() => {
     }
   }
 
-  w.dualContouring = () => {
+  w.createChunk = (x, y, z) => {
```
avaer, line 1817: Can we ensure the names reflect this is dual quaternions?

Issue #2890, opened 2022-04-25
Jumping on dev build causes crash due to grunt SFX


solsnare:
```
character-sfx.js:378 Uncaught TypeError: Cannot read properties of null (reading 'actionVoices')
    at CharacterSfx.playGrunt (:3000/character-sfx.js:378:46)
    at _handleJump (:3000/character-sfx.js:86:16)
    at CharacterSfx.update (:3000/character-sfx.js:96:5)
    at LocalPlayer.updateAvatar (:3000/character-controller.js:1206:25)
    at _pre (:3000/webaverse.js:286:23)
    at _frame (:3000/webaverse.js:307:9)
    at animate (:3000/webaverse.js:326:7)
    at onAnimationFrame (:3000/packages/three/build/three.module.js:1254:95)
    at onAnimationFrame (:3000/packages/three/build/three.module.js:498:545)
```

This isn't reproducible on a prod build, only on dev does it crash.

PR #2892, opened 2022-04-26
fix crash due to grunt

tcm390:
I think the issue is happened because the voicePack hasn't loaded https://github.com/webaverse/app/issues/2890
I added a `setTimeout` in voicePack loading function and the app crash:
image

So I added a condition in `playGrunt` to check whether the voicePack is already loaded or not. 
image


avaer:
Thanks! This fix looks correct, but I think it can be written cleaner.

avaer:
(Also this PR updates `package-lock.json` for some reason)

avaer:
Closed by https://github.com/webaverse/app/pull/2893 (same fix)

PR #2894, opened 2022-04-26
Target reticle

avaer:
For combat and edit modes.


avaer:
It's not hooked up yet; merging should have no effect.

PR #2897, opened 2022-04-26
solve sfx issues

Arthur-Acha:
-10dB on run files
-10dB on walk files
-10dB on land files
normalized OOT sounds
reworked SonicBoomTrail
recompiled all the sounds


avaer:
In my testing, this PR completely breaks all sounds, including walking.

PR #2899, opened 2022-04-26
Focus bar

avaer:
Adds a focus mode with middle-click. This is a prerequsite for combat targeting and npc target dialogue.


avaer:
Also adds `sounds.playSoundName`, a bunch of sounds, and hooks up many UI sound triggers.

PR #2901, opened 2022-04-26
Resolves #2891 UI: Improved scene-select component and fixed input typing issue.

### LINKED ISSUE #2891: Can't type in scene address bar
whatever I type, even backspace, creates new text

TheoTheDev:
Improved SceneSelect input [top center input on the screen].
- No you can search through scenes just starting typing.
- Now you can insert/type URL of scene file you want to load, press 'enter' end load it immediately.

https://user-images.githubusercontent.com/1478866/165323549-160be3ab-b434-44d2-81b9-e805a1ec514f.mp4

@alisaad673 PR ready to be tested.

### DIFF
--- a/src/components/editor-mode/scene-menu/SceneMenu.jsx
+++ b/src/components/editor-mode/scene-menu/SceneMenu.jsx
@@ -1,5 +1,5 @@
 
-import React, { useState, useEffect, useContext } from 'react';
+import React, { useState, useEffect, useContext, useRef } from 'react';
 import classnames from 'classnames';
 
 import { world } from '../../../../world'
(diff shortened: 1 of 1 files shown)


avaer (approved):
Looks good if it tests ok.

msm0823:
I have identified the following issues.

1. Scene Url is given instead of scene name on the search bar

2. While searching for a scene, On pressing enter, it triggers chat box

3. While we type in a new scene url, the app shift to that scene and default scenes disappeared while default scenes should be displayed irrespective of what is being entered in the search box

4. While try to "dash" by pressing "shift + double w" it throws this error and crashes the app.

5. When we start searching, it only searches from the available scenes in the panel. If there is no scene in the panel, it won't search.

alisaad673:
I have update the description of issues and added some screenshots and videos.

TheoTheDev:
> In the scenes panel, Previously there was the name of scene only along with .scn extension, but right now it shows the whole url for the scene.

it was supposed to be so, cause if having different path in future and same name it will be confusing, no ?

> While searching for a scene, On pressing "enter", it triggers chat box

fixed

> While we type in a new scene url, the app shifts to that scene and default scenes disappear, while default scenes should be displayed irrespective of what is being entered in the search box. IMO, the arrow on left end of panel should always show the default scenes, It should not get updated as we search for some scene.

>When we start searching, it only searches from the available scenes in the panel. If there is no scene in the panel, it won't search. As mentioned in the point no. 3, the scns panel gets updated just as we search for some scene, the panel shows that scn and all other scenes are now non- recoverable

Fixed and now clearing/reseting 'search' when empty input.

> While we try to "dash" by pressing "shift + double w" it throws this error and crashes the app.

I think this is not related to this PR

avaer:
> While we try to "dash" by pressing "shift + double w" it throws this error and crashes the app.

@msm0823 if this is still seen in master it should be filed.

avaer:
The search is styled poorly, with random spaces.

image

avaer:
Merging since it's better than what we had.

Issue #2690, opened 2022-03-29
app: drag & dropped VRMs go invisible after scn switching


alisaad673:
The avatars that are drag and dropped in to the app go invisible when scene is switched

https://user-images.githubusercontent.com/35926530/160667480-bcfd5fd6-ff03-4b50-b310-cd8f1ca527b2.mp4

msm0823:
https://github.com/webaverse/app/issues/2793

related

Issue #2793, opened 2022-04-08
Drag and dropped assets do not get teleported along with avatar


alisaad673:
Issue#1
-The drag and dropped assets such as swords bow, etc do no get teleported along with avatars as we switch scenes. 
-Only the asset itself isn't there with avatar
-The action/animation related to asset still get performed as it can be seen in the video below
-Unable to drop the silsword wit "R" key.(Dropping bow with R key works here)

https://user-images.githubusercontent.com/35926530/162500635-1c61ccc6-6e8f-4a52-9484-e9a5322e54a0.mp4

Issue#2
In the above case, after we have changed the scene after equipping a drag & dropped asset, Dropping the object and again changing the scene produces this error:
Steps to reproduce:
1-Equip a drag and dropped asset
2-Switch the scene (the object doesn't gets teleported along with vrm)
3-Drop the object by pressing "R"
4-Again switch the scene

![image](https://user-images.githubusercontent.com/35926530/162503878-8f7d0a83-c689-41cd-a44e-b2205e4c779d.png)

avaer:
Yes, looks like an item AppManager ownership problem.

avaer:
Note however that this kind of "teleport" will not exist and that menu will not be there.

alisaad673:
Updated the issue with some more info.

alisaad673:
Fixed

PR #2902, opened 2022-04-26
Load fresh new page after changed scene.

gonnavis:
Fix: https://github.com/webaverse/app/issues/2690 , https://github.com/webaverse/app/issues/2793

Because of we don't need _teleport_ feature. https://github.com/webaverse/app/issues/2793#issuecomment-1093237663
So how about changing `location.href` to load a fresh new page after changed scene, instead of using `push/popState`?


avaer:
No I don't think that fixes anything. We still need to make sure push works since we will use it for overworld traversal.

gonnavis:
Ok, I'll try other ways.

PR #2908, opened 2022-04-27
Resolves #2907 physics mesh not being updated when changing the scale of the object

### LINKED ISSUE #2907: Physics mesh not being updated when changing the scale of the object.
Need to add reAddGeometry method for Physics-manager where physics mesh will be readded but without changing the ID of the physics object.

TheoTheDev:
- Added .reAddGeometry method to physics-manager to readd the geometry and keep same physics mesh id.
- Added calling .reAddGeometry method when changing mesh scale via WorldObjects object inspector UI.

Based on the PR #2785 

https://user-images.githubusercontent.com/1478866/165483225-4a065d34-002e-49e5-b271-3d7fd18d5d46.mov

.


avaer:
Needs a rebuild + conflict resolution.

avaer:
Depends on https://github.com/webaverse/app/pull/2928 due to updating `geometry.wasm`

PR #2913, opened 2022-04-27
Physics refactor

avaer:
Ingests https://github.com/webaverse/app-wasm/pull/45.

Prepares us for procedural physics, by breaking out the concepts of shapes/materials from physics objects. The API is refactored with different arguments as a result but there should be no impact on behavior in the app.

This could use a round of testing to ensure no regressions.


avaer:
I'll hold off merging this until it can be tested.

avaer:
The next step for this once merged it so implement a physx forward-sweep so we can do accurate z-targeting.

avaer:
In my testing this has been working fine, so merging.

PR #2914, opened 2022-04-28
adding physx forces methods

Loryhoof:


avaer:
Needs re-testing after the `app-wasm` rebuild. I think I saw a crash here.

avaer:
This requires a WASM rebuild, so it might be behind https://github.com/webaverse/app/pull/2928 to ship

avaer:
After testing it works on latest so merging.

avaer (approved):

PR #2915, opened 2022-04-28
Fix right ikHand wrongly triggered problem.

gonnavis:
Sibling PR: https://github.com/webaverse/pistol/pull/10

Fixed the problem that right ikHand was wrongly triggered when activating sword in aim mode.

image

image


avaer (approved):

gonnavis:
Hello @avaer , if just merge this but not the sibling PR https://github.com/webaverse/pistol/pull/10 , will cause the pistol to not work properly.

Issue #2876, opened 2022-04-22
Performance: Wear/Unwear vehicle repeatedly.


gonnavis:
If wear/unwear ( key E/R ) vehicle repeatedly, the performance will drop dramatically.

![image](https://user-images.githubusercontent.com/10785634/164714630-74488a1a-1eb5-4fcf-b6aa-687a8deb759d.png)
![image](https://user-images.githubusercontent.com/10785634/164714642-e8b907b1-a635-4d89-8444-bc2962225330.png)

avaer:
Interesting, it is keeping the actions around somehow?

avaer:
Looks like something in `zjs`.

avaer:
To fix this I would investigate the actions adding/removal code more deeply, especially things like the listeners and callbacks, to make sure everything is correctly cleared when the app is unworn.

avaer:
I noticed that wearing/unwearing the sword will cause similar glitches. The actions are definitely getting stuck.

gonnavis:
Reopen because of revert https://github.com/webaverse/app/pull/2916#issuecomment-1123095807

gonnavis:
Closed by https://github.com/webaverse/app/pull/2916#issuecomment-1126847544

PR #2916, opened 2022-04-28
Stuck wear/unwear due to components stringify encode

gonnavis:
Fix: https://github.com/webaverse/app/issues/2876

I found the problem caused by repeatedly `JSON.stringify` an empty `components` ( array ).
Will get a very large sting in the end and will try to `stringify/parse` it again, so stuck:

image
image

I'm not sure why `JSON.stringify/parse` `components` in `app-manager.js`. Deep copy or something else?
But I tested that it works well without them, at least on my local, non-multiplayer environment.
( And the old `components` value broken totally, so should works as before )


avaer:
I'm not sure if these stringifys are redundant; there are there because we need to make sure at every step this data can be transferred over URL. Just make sure that it doesn't break when there are special characters in the metadata of the apps.

avaer:
It's a good catch though and this is definitely a bug. We just need to be sure this does not cause regressions.

avaer (approved):
This is probably correct if it does not regress anything in scene handling.

avaer:
Tested and looks good.

avaer:
Thanks for figuring this out @gonnavis!

avaer:
This broke NPC loading in `z-targeting` (and probably all direct `content:` loading in totum).

```
metaversefile-api.js?t=1652233721943:419 error loading "/@proxy/data:application/npc,%257B%2522name%2522:%2522Ann%2522,%2522avatarUrl%2522:%2522/avatars/ann.vrm%2522,%2522voice%2522:%2522Maud%2520Pie%2522,%2522bio%2522:%252213/F%2520Lisk%2520Witch%2520(black%2520mage).%2520She%2520is%2520quirky%2520and%2520is%2520always%2520creating%2520crazy%2520new%2520potions.%2520She%2520has%2520a%2520dark%2520streak%2520and%2520is%2520known%2520to%2520torment%2520her%2520teachers%2520when%2520they%2520give%2520her%2520bad%2520grades.%2522%257D" SyntaxError: Unexpected token ':'
```

I think I'll have to revert this.

avaer:
Reverted.

avaer:
I have been trying this in `z-targeting`, and it has been stable after a lot of totum refactoring. So I'll bring this back in `master` as well.

avaer:
Reverted the revert here: https://github.com/webaverse/app/commit/ad46d586153e5074429cb12c8f69b52e5f5754b6

PR #2928, opened 2022-04-29
Z-targeting

avaer:
### Dependents:
https://github.com/webaverse/chest/pull/5
https://github.com/webaverse/pistol/pull/11
https://github.com/webaverse/flower/pull/2
https://github.com/webaverse/street-base/pull/1
https://github.com/webaverse/silk-grass/pull/1


avaer:
Depends on https://github.com/webaverse/lore/pull/1

avaer:
For now this repo needs a local checkout of https://github.com/webaverse/lore to function. It should be working otherwise.

avaer:
The ai features will only function with a self inserted openai key. That don’t be the case in production.

avaer:
This PR ended up adding a ton of new features in addition to Z-targeting:

- Target reticles
- New RPG story engine
- New RPG commenting engine
- Middle click to focus
- Double-tap E to lock on
- Z-Camera lock
- Per-avatar voice/endpoint setting, works with character select

avaer:
This PR needs testing to make sure it didn't regress any key features.

alisaad673:
Tested the major features, Didn't found much regressions: These are some issues i found:

1- While we select goose ai in AI settings and then try to select character in character tab, It throws this error on each of the characters. Video attached.
image

https://user-images.githubusercontent.com/35926530/167192273-740099a5-8218-4bba-8c13-ff093ab9bec3.mp4



2-Then also, If we don't put the api key in openAi field but select it as our Ai in settings, We are unable to select characters in character tab. Imo it should only block the lore thing but the ability to select character should be there.

https://user-images.githubusercontent.com/35926530/167196094-c25daaa5-05cf-4f69-a31c-20a45da4f6a2.mp4





Some Suggestions:
-The camera shifts to closer to avatar once there is hurdle or an object closer to the avatar, This doesn't always looks good as most of the times while we unwear assets, the camera shifts automatically to first person. And in some cases, we are unable to move it back until we change the direction of avatar.

https://user-images.githubusercontent.com/35926530/167193370-b71d8123-317f-4b87-9276-d0f763e2576e.mp4

-While we go to character select panel, And parse over the characters, it plays separate music for each character, The issue that i see here is it keeps playing the music even if we dont select that character.
-There is no option to stop the music that plays for character as it stops on its on when it gets finished.


https://user-images.githubusercontent.com/35926530/167195108-b04ab31d-2e82-4c49-9769-c0850afaf804.mp4

avaer:
- [ ] Update themes
- [ ] Character select colors should be controllable
- [ ] Cancel music on menu exit
- [ ] Fix camera bump-lock

PR #2930, opened 2022-05-01
Dual Contouring LOD

aryasalehi:
This is the sibling PR of this : https://github.com/webaverse/app-wasm/pull/50


avaer (approved):

avaer:
This looks good, and I would like to merge this, but two problems:

- Needs WASM rebuild (the upstream needs a rebuild too)
- `app-wasm` is floating incompatible changes with `app` right now, since that is used in https://github.com/webaverse/app/tree/z-targeting, a PR which contains a lot of other things and is not easily broken up. So that means this PR is probably waiting behind that branch

Issue #1978, opened 2021-12-12
Aim animation smooth lerp


avaer:
We should smooth in the inverse kinematics when you aim your weapon, rather than jumping to the new pose.

https://user-images.githubusercontent.com/6926057/145730263-a4d0ad3e-8010-4c30-b940-b3453356aef5.mp4

PR #2931, opened 2022-05-01
VRArmIK Aim animation smooth lerp.

gonnavis:
Fix: https://github.com/webaverse/app/issues/1978

https://user-images.githubusercontent.com/10785634/166149310-275557d0-30e8-463d-bc70-dae669934015.mp4


avaer (approved):
Works great! We can probably do similarly with the sword.

PR #2937, opened 2022-05-04
add emote voice

tcm390:
play emote voice in character-sfx, this is the result:

https://user-images.githubusercontent.com/60634884/166614661-d334c58a-115a-4a15-b1d6-51fdd851591a.mp4


tcm390:
@Arthur-Acha Could you check 2 and 3 ?

Arthur-Acha:
For 2 I am aware of this, I made a request for a revision for angry and embarrassed from Bryce. This was 8 days ago and I had no response so far.
Can you please give examples of 3? @alisaad673 I did not noticed that during my testings

Arthur-Acha:
> Yes. 1 is fixed. 3- @Arthur-Acha Here are some examples. Its not something big of an issue, I just noticed it. Probably it can be ignored? Headnod, Headshake, Sad has animations longer than the sound.
> 
>  2022-05-16.23-19-56.mp4 
>  2022-05-16.23-20-29.mp4

Yeah, I think this isn't even an issue. It would be an issue if it was the other way around, but even IRL we don't reproduce sounds the who length of an action. Looks natural to me

avaer (approved):

avaer:
I think this PR caused regressions with emotes getting stuck. Will confirm and revert if that's the case.

avaer:
Actually it seems not related, the emotes simply get stuck if you do several of them at a time.

PR #2938, opened 2022-05-04
update compile-sounds.sh

Arthur-Acha:
update compile-sounds.sh to include /navi and /music


avaer:
I don't think music should be in here. Is that ever played through the sound system? I don't think it is, in which case this is a double load? I could be wrong but please check.

The `navi` part seems ok.

avaer (approved):

PR #2941, opened 2022-05-04
Let localPlayer look at selected z-target.

gonnavis:
https://user-images.githubusercontent.com/10785634/166649681-da7f2a4a-5f9e-4d37-a0e4-c44a3163f637.mp4


avaer:
It looks good! Merging.

We will also want similar technology where the character hips will face the target (and they sidestep around the circle with the target in the center).

The idea is for the double-tap `E` key Z-lock, the character is in combat mode and will always face their body towards the enemy. But for the middle click z-target, the character will only look at the target (what is here).

avaer:
This is still actually broken so I will revert it for now. 
image

The problem is probably in the eye target code, not here, but this PR still interferes a bit.

gonnavis:
> It looks good! Merging.
> 
> We will also want similar technology where the character hips will face the target (and they sidestep around the circle with the target in the center).
> 
> The idea is, for the double-tap `E` key Z-lock, the character is in combat mode and will always face their body towards the enemy.
> 
> But for the middle click z-target, the character will only look at the target (that is what is in this PR).

@avaer Ok! I'm working on this.

gonnavis:
@avaer In Z-lock mode, should camera follow player, and keep target in the center of screen, like most action games do, such as Sekiro?

https://user-images.githubusercontent.com/10785634/166909615-70c4fc92-b9a0-460d-b545-14f9099e0f13.mp4

Or keep the camera at the initial position like what we currently do?

avaer:
> In Z-lock mode, should camera follow player, and keep target in the center of screen, like most action games do, such as Sekiro?

Yes.

> Or keep the camera at the initial position like what we currently do?

There are two lock modes. The middle mouse one holds the camera in place as you inspect something.

The Double-Tap E toggle method should try to follow the scene as you described.

PR #2952, opened 2022-05-08
MeshCutter: Can set `planeNormal/Distance`, and add `isIndexed`.

gonnavis:
Sibling PRs: https://github.com/webaverse/app-wasm/pull/54 , https://github.com/webaverse/mesh-cut-example/pull/3

MeshCutter arg: Use `planeNormal/Distance` instead of `planeQuaternion`, and can set it by needs now.

And add arg: `isIndexed`, to support both indexed and non-indexed geometries, in order to do recursively cutting.

image

image

image


physx.js
```diff
@@ -891,17 +892,14 @@ const physxWorker = (() => {
     const uvsTypedArray = allocator.alloc(Float32Array, numUvs)
     uvsTypedArray.set(uvs)
 
-    const facesTypedArray = allocator.alloc(Uint32Array, numFaces)
-    facesTypedArray.set(faces)
-
-    const positionTypedArray = allocator.alloc(Float32Array, 3)
-    positionTypedArray.set(position)
+    let facesTypedArray;
```
avaer, line 895: Btw this could be done with just passing `faces === null`.

physx.js
```diff
@@ -891,17 +892,14 @@ const physxWorker = (() => {
     const uvsTypedArray = allocator.alloc(Float32Array, numUvs)
     uvsTypedArray.set(uvs)
 
-    const facesTypedArray = allocator.alloc(Uint32Array, numFaces)
-    facesTypedArray.set(faces)
-
-    const positionTypedArray = allocator.alloc(Float32Array, 3)
-    positionTypedArray.set(position)
+    let facesTypedArray;
+    if (isIndexed) {
```
avaer, line 896: Btw this could be done with just passing faces === null.

avaer (approved):
Looks good, just one cleanup I would have recommended.

physx.js
gonnavis: Modified https://github.com/webaverse/app/pull/2954
At js side, I check if `faces` is falsy, is it ok? Or need to input exact 'null' ?

PR #2959, opened 2022-05-10
Particle System 2

avaer:
Adds a new multi-texture particle system.

Needed for https://github.com/webaverse/flower/pull/1.

Part of gameplay definitions.


avaer:
Tested and it works fine.

PR #2960, opened 2022-05-10
Bugfix avatars foot offset

avaer:
Cleans up some math for avatars that crashed due to having their armature local offset below the floor, as well as letting the avatar define the foot-to-floor distance via the initial position of the export.


lalalune:
@avaer is this still valid? I can look at it

avaer:
No, I don't think it's valid at the moment.

PR #2965, opened 2022-05-10
Sword damage based on sword damageBox instead of localPlayer's position.

gonnavis:
Sibling PRs: https://github.com/webaverse/silsword/pull/8 , https://github.com/webaverse/street/pull/6

Let sword damage based on sword damageBox instead of localPlayer's position.

And can damage multiple items simultaneously.

In order to prevent damage objects such as ground, add a new component `canGetHurt` for resource repos, if set false, it'll not be damaged by sword.
Made a PR for street repo first, and let `.scn` access `street/` instead of `street/index.js` in order to get canGetHurt component info from `.metaversefile` file.

Before:

https://user-images.githubusercontent.com/10785634/167693676-05f8b757-bc8b-4103-a211-28d84c43a69e.mp4

After:

https://user-images.githubusercontent.com/10785634/167693688-2f35b220-d41c-4163-94dc-8d70a626201a.mp4


avaer:
Awesome, but what about if instead of `canGetHurt` it's `vincibility: 'invincible'`? I can imagine many other different modes than this (like 'double damage' or 'instantDeath').

avaer:
Also, IMO the timing of damage needs to be reworked per those videos.

I don't think there should be any looping kind of damage when you hold a button. Hits should only happen when the player actually performs the hit action (button press), and the actual hit happens at the time determined by the animation. At that frame, we trigger a hit. Collided objects are tested for vincibility, and this can take into account if they were recently hit (invincibility frames).

This is related to combos/animation work.

avaer:
Still, this PR can be merged if we clean up the naming.

gonnavis:
@avaer Renamed first. https://github.com/webaverse/app/pull/2965/commits/1e9933624ce9d0efcd607d2814caa74a3e33404a , https://github.com/webaverse/street/pull/6/commits/67030e7f7fc62a10d94c790267f878f54f32bea8

avaer:
This code changed a lot in `z-targeting` so probably needs to be ported to `character-hitter`:

https://github.com/webaverse/app/blob/2206fbb4d790d8661efe97cae4ff424744eea22d/character-hitter.js#L63

But that would be a separate PR.

avaer (commented):
This code changed a lot in `z-targeting` so probably needs to be ported to `character-hitter`:

https://github.com/webaverse/app/blob/2206fbb4d790d8661efe97cae4ff424744eea22d/character-hitter.js#L63

But that would be a separate PR (to `z-tareting`).

gonnavis:
@avaer Ported to `z-targeting` branch https://github.com/webaverse/app/pull/3040
And closed in favor of it.

PR #2967, opened 2022-05-10
Integrate heal effect2

tcm390:
Integrate heal effect


character-fx.js
```diff
@@ -246,6 +247,24 @@ class CharacterFx {
       }
     };
     _updateSonicBoomMesh();
+    const _updateHealEffectMesh = () => {
+      if (!this.healEffect ) {
+        this.healEffect = metaversefile.createApp();
+        (async () => {
+          await metaverseModules.waitForLoad();
```
avaer, line 254: IMO we should not wait this here. By the time this code runs, `wevaverse.js` should already have waited for everything we need.

The list of all of the loads is here: https://github.com/webaverse/app/blob/77f51730e851f7599287b30da60d1142f01a84c8/webaverse.js#L79

avaer: Instead, we can create this app the first time it is needed, when the character first gets a heal action.

Since the module is preloaded, it should be ready to go when we hit this code path so there should not be any loading delay.

avaer (commented):
Some timing fixes are probably needed, but it looks good overall!

avaer (approved):

PR #2949, opened 2022-05-06
Sync player's `eyeTarget` and `eyeballTarget`.

gonnavis:
# LocalPlayer

Sync LocalPlayer's `eyeballTarget` with `eyeTarget`, let them both look at `zTargeting.focusTargetReticle`.

### Before:

image

### After:

lastNeedsEyeTarget = true:
image
image

lastNeedsEyeTarget = false:
image

# NpcPlayer

Sync NpcPlayer's `eyeTarget` with `eyeballTarget`, let them both look at localPlayer by default.

### Before:
image

### After:
image

PR #2971, opened 2022-05-11
Sync player's `headTarget` and `eyeballTarget`. ( 2nd try )

gonnavis:
Follow and Fix: https://github.com/webaverse/app/pull/2949#issuecomment-1123134816

Not reset `needLimitEyeballTargetRange` in the previous PR, caused eyeball not look into mirror after did z-targeting problem.
Should be fixed now. https://github.com/webaverse/app/pull/2971/commits/92a67cd45e464aec1a40b0e127f02ede1a621e61


gonnavis:
@avaer Solved conflicts.

alisaad673:
This seems to be working fine, the avatar behaves in similar way in front of mirror as it does in the master. When not in z-targeting, the eyeballs do point towards the mirror.
While being in z-targeting, the head moves along with eyeballs towards the target. Also the npc looks at the avatar along with head and eyeballs in sync but upto a certain limit.

character-controller.js
```diff
@@ -491,6 +497,21 @@ class PlayerBase extends THREE.Object3D {
...
     }
   }
+  setTarget(target) { // set both head and eyeball target;
+    if (target) {
+      this.headTarget.copy(target);
+      this.headTargetInverted = true;
+      this.headTargetEnabled = true;
+
+      this.eyeballTarget.copy(target);
+      this.needLimitEyeballTargetRange = true;
```
avaer, line 507: I don't get the point of this. The range needs to always be clipped to not have the eyes too far our.

avaer: Can we just get rid of this?

gonnavis: @avaer Ok, deleted `needLimitEyeballTargetRange`.

I added `needLimitEyeballTargetRange` mainly due to two concerns:
1. Prevent re-calculate root-to-head angle, use `lastNeedsHeadTarget` which already calculated in head for eyeball.
2. To keep the same behavior as before, which eyeball will forced to rotate toward mirror, regardless to angle limit.
Before:
image
image

For 1, though it simplified calculations, but really made logics complicated and unstable.
For 2, I think it's fine to not rotate eyeballs in front of mirror if angle is too large. ( Still look into mirror if angle is in limit )
After:
image
image
image

So deleted `needLimitEyeballTargetRange`.
Re-calculated `needEyeballTarget` instead of using `lastNeedsHeadTarget`, untangled logics, and can set different max angle limitation  if need.

BTW, fixed `mouse look` feature not work problem.

lalalune:
@avaer good for @alisaad673 to review this?

alisaad673:
Functionality wise, It seems fine. But it is based on an older version of z-targeting and has a lot of bugs. @gonnavis Can you please rebase it over to the latest?

rebase

gonnavis:
@alisaad673 Updated.

PR #2973, opened 2022-05-11
2968 localhost -> local.webaverse.com on startup

mwmwmw:
Updated docs and README.md to reflect the new instructions. 

This does not resolve the forever loading avatar picture issue. I tracked that down to a separate problem and will open a PR once I'm sure I understand the problem and have a fix.


avaer (approved):

avaer:
> This does not resolve the forever loading avatar picture issue. I tracked that down to a separate problem and will open a PR once I'm sure I understand the problem and have a fix.

Awesome, thanks!

PR #2996, opened 2022-05-13
Added hide mode.

TheoTheDev:
- ActionMenu btn working.
- ctrl+h hot-key for toggling.

Mostly done, but needs some UX review and maybe tweaks.


avaer (changes requested):
Can we keep the styling at the top level instead of deep classes. Specifically User.css

avaer:
Also, let’s make a ShowUI component that abstracts this change and wraps the nodes.

avaer:
Might be a while before this gets updated.

Issue #2974, opened 2022-05-11
OffscreenEngine startup/engineReady race condition


mwmwmw:
The loading speed and event order for `engine-worker.js` and `offscreen-engine.js` are unreliable and can cause a race condition that results in an unloaded avatar. 

## Fail Case
In the failed case, the `engineReady` message is sent too early, before `offscreen-engine.js` is listening for it. 
![image](https://user-images.githubusercontent.com/889851/167959334-d813be54-e6eb-4804-b684-5aa2ef257af4.png)

## Success Case
In the successful case, so long as `engineReady` is received after `OffscreenEngine` is listening for it. The avatar image will appear. 
![image](https://user-images.githubusercontent.com/889851/167962258-4af9d97e-e9ba-4a98-bcf3-fad264d686e5.png)

## Hack Proof of Concept Fix
I created a hack fix where I manually delay `engineReady` to allow `offscreen-engine.js` to set up and wait for the event.
![image](https://user-images.githubusercontent.com/889851/167959583-bc0e03d4-6a79-45f1-9291-79a42a22c150.png)

Every case where I was able to have `offscreen-engine.js` wait for `engineReady` before it was sent from `engine-worker.js` resulted in a profile pic appearing. 

I don't know why this order changes in different environments. It could be network speed, extensions, cache, etc. For example, in incognito mode, the `engineReady` event can be sent at almost any time, and the `offscreen-engine.js` will receive it (see Success Case log). 

That might be worth looking into more.  
 
However, I know that ensuring that `offscreen-engine` receives the `engineReady` event is a fix for all environments. 

## Fix
In `offscreen-engine.js`, set up a message handler before loading the iframe. This fix will ensure that no matter how quickly `engineReady` is sent, `offscreen-engine.js` can capture it. 

---

PR #2998, opened 2022-05-13
2974 offscreen engine race condition

mwmwmw:
## Issue
https://github.com/webaverse/app/issues/2974

## Fix

- Bind the `engineReady` handler before the iframe source is set. Once the port has been received and the iframe has loaded, the port will be started and the `waitForLoad()` promise is resolved.

- Update the `_inappPreviewHost` constant for dev, so that iframes are loaded locally. (but the fix works even if the iframe is one of the remotes, e.g. app.webaverse.com)

image

## Tested
Tested in `master` and a merge spot check for `z-targeting` (no conflicts)

Be sure to test in a worst-case scenario. User-land browser, with extensions on, and cache enabled.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -40,7 +40,7 @@ let _inappPreviewHost = '';
 
 switch ( origin ) {
     case 'https://local.webaverse.com': {
-        _inappPreviewHost = `https://local.webaverse.com:${window.location.port}`;
+        _inappPreviewHost = `https://local.webaverse.online:${window.location.port}`;
         break;
     }
     case 'https://dev.webaverse.com': {
--- a/constants.js
+++ b/constants.js
@@ -40,7 +40,7 @@ let _inappPreviewHost = '';
 
 switch ( origin ) {
     case 'https://local.webaverse.com': {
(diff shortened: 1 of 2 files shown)


constants.js
```diff
@@ -40,7 +40,7 @@ let _inappPreviewHost = '';
 
 switch ( origin ) {
     case 'https://local.webaverse.com': {
-        _inappPreviewHost = 'https://local.webaverse.online';
+        _inappPreviewHost = `https://local.webaverse.com:${window.location.port}`;
```
avaer, line 43: This is not legal. The iframe must be on a different tld+1 to get a new render process.

avaer (changes requested):
Incorrect iframe url was added.

constants.js
mwmwmw: ahh, I remember you mentioning the new render thread. It makes sense because it thinks it's a new website.  

This was initially set to `.online` with no port and if you ran it off 3000, it would never load. I thought it was the wrong URL, but it was just missing the port. Reverted to `.online` but kept the port so that it'll work for any port when running local.

avaer (approved):

PR #2999, opened 2022-05-14
Z-targeting: Z-lock. [WIP]

gonnavis:
Related discussion: https://github.com/webaverse/app/pull/2941#issuecomment-1117965752 , https://github.com/webaverse/app/pull/2941#issuecomment-1118419112

https://user-images.githubusercontent.com/10785634/168422094-b4c7f7c5-5fda-4ac4-b508-d3d9eee179ed.mp4

Todo:
- Should apply combat Z-lock logics on `keysDirection` in `io-manager.js`, instead of directly apply to `moveCharacterController` in `character-physics.js`.


avaer:
It's a good start. IMO the camera should be attempting cinematic angles and smoothing between them rather than trying to be centered.

I think maybe we should break out the cinematic camera logic into a separate manager. `z-targeting` already has a new camera manager implementation, but the cinematic tracking seems it should be separate from that.

There are 4 places I know we need special cinematic camera angles which can be covered by a cinematic camera manager abstraction:
- Z-targeting
- Story scenes
- Introductions (e.g. to world)
- Combat focus

gonnavis:
Put target/enemy at upper right.

https://user-images.githubusercontent.com/10785634/168485490-7e709168-c84c-4896-8d9f-c0250d269fa5.mp4

gonnavis:
- Generalize z-lock logics to all focused objects ( except when chatting ) , not only when combat.
- Focus to `focusTargetReticle.position` instead of object's position.
- Focus to head bone position if the object is avatar.
- Support dynamic/moving `focusTargetReticle`.
- Let player's moving in z-lock respect velocity damping too.
- Adjust camera offset/pan in z-lock, and not affect camera when chatting.
- Fix naruto run orientation in z-lock.

https://user-images.githubusercontent.com/10785634/169782322-7334a936-390f-4aba-940a-ad55c738fb85.mp4

But this version still too robotic, all using fixed camera offset/pan etc.
I'll refer more to Zelda's camera controlling. And I found Sifu's camera controlling is very good, should be a good reference too.

lalalune:
@gonnavis where we at with this? need qa and review? seems like we have a few smol merge conflicts, anything else to pull this in?

gonnavis:
@lalalune 
At first I thought avaer wanted something like Sekiro to lock the enemy in the center, and then the player would do a circular motion around the enemy.  And preliminarily implemented.
https://github.com/webaverse/app/pull/2941#issuecomment-1118419112
However, it was later made clear that avaer does not want this kind of hard locking, but comprehensively considers the position of the player and the target, as well as the collision between the camera and the scene, and needs a viewing angle with a certain inclination.
https://github.com/webaverse/app/pull/2999#issuecomment-1126835310
Mainly like Zelda https://www.youtube.com/watch?v=0aJIVLN8E68
Maybe can refer to the Sifu game's camera control too I think. 
So, need to remove the player circle moving logic, and do a better camera position/rotation calculation and transition. 

BanditBan has tried to continue this feature before.
I'll keep working on it if needed too. ( But my current time is fully occupied by animation tasks ~~ )

PR #3001, opened 2022-05-15
Hide mode

TheoTheDev:
Added UIMode component wrap for each UI block to enable/disable 'UI mode'.

### DIFF
--- a/src/AvatarIcon.jsx
+++ b/src/AvatarIcon.jsx
@@ -107,8 +107,8 @@ const CharacterIcon = () => {
   );
 };
 
-export const AvatarIcon = () => {
-    const { state, setState, uiMode } = useContext( AppContext );
+export const AvatarIcon = ({ className }) => {
+    const { state, setState } = useContext( AppContext );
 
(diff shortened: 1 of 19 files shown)


avaer (approved):
Code seems ok.

TheoTheDev:
@alisaad673 can you plz test
and @avaer merge if all ok.

TheoTheDev:
> This Looks good. Just Two Questions: 1- We can hide the UI by clicking "Hide" but to bring it back? There's no obvious key. Though pressing tab or z brings back the UI but it seems to be a hack. Can we allocate a separate Key to Hide mode? SO that pressing that key can hide/unhide the UI?
> 
> 2-While we have equipped an item, Such as sword or anything else, and then we hide the UI, we still have the item preview in the bottom right corner. Shouldn't it hide too? @avaer Can you suggest on these too?
> 
>  2022-05-16.20-09-27.mp4

yes, updated PR and it's description.
1. Added a tooltip with hot-key.
2. Fixed. now equipped item menu also is hiding.

PR #3002, opened 2022-05-15
UI/component editor

TheoTheDev:
Added showing Components params [key is not changeable yet, and can not add/remove yet].

- Show app components.
- Remove componenent.
- Add component.
- Edit component value.
- Edit component key.
- Set component value type.
- Validate component value [if not valid doesn't save].

https://user-images.githubusercontent.com/1478866/168781710-f804ff6d-725c-48b8-8965-98e1ca9ebb0a.mov

ps: based on pr #2969


avaer:
I think perhaps instance I’d and content id should not be editable. At least having these change would require detecting and resolving naming conflicts (eg auto rename).

avaer (approved):

alisaad673:
1- While we click on "add new", There is no real time response as no new component is added. The new component box appears after we, close the world editor and re open it. 

https://user-images.githubusercontent.com/35926530/168896199-33d17463-2827-41df-8e65-1bfc4b610831.mp4

2-Continuation of the above issue, as it doesn't opens the text box in realtime, the user can click the "add new" multiple times and that throws this error.

https://user-images.githubusercontent.com/35926530/168896130-f06d0f3a-feb6-4538-8ef4-6e87c8f1628e.mp4

image

3-There are a few more issues related to the same above issues like, once multiple instance of same new item are created, and we try to edit the name, it updates the name of all three instances and creates some more new instances.

avaer:
Closing because there is nobody to ship this.

PR #3009, opened 2022-05-17
Action generation API integration

crazylazylife:
The first version of the Action generation API is implemented. Tested with "hangout.scn" file.

- Action generated for the last spoken character.
- Multiple actions are generated to complete a certain task.
- Actions are console logged, to avoid any conflict in the implementation.

Note:
- The API works well for one-on-one conversations. But the accuracy degrades for multi-party conversations.

### DIFF
--- a/ai/lore/lore-ai.js
+++ b/ai/lore/lore-ai.js
@@ -16,6 +16,11 @@ const numGenerateTries = 5;
 const temperature = 1;
 const top_p = 1;
 
+//--------------------------------- ConvAI Mods ------------------------------
+// Initializing the action generation URL from ConvAI
+const actionGenerationURl = `https://test.convai.com/getActions`
+//----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


avaer:
Why add another scene?

ai/lore/lore-ai.js
```diff
@@ -16,6 +16,11 @@ const numGenerateTries = 5;
 const temperature = 1;
 const top_p = 1;
 
+//--------------------------------- ConvAI Mods ------------------------------
+// Initializing the action generation URL from ConvAI
+const actionGenerationURl = `https://test.convai.com/getActions`
```
avaer, line 21: This is not how we integrate apis. All generators must support all modes of operation. We do not special case them.

avaer (changes requested):
The design of this uses hardcoded apis.

PR #3010, opened 2022-05-17
Fix pick up tree animation problem.

gonnavis:
Fix: https://github.com/webaverse/app/issues/3060

Sibling PR: https://github.com/webaverse/plants/pull/1

Fix pick up tree animation problem.
- Updated new model `scilly_drophunter_v31.6_Guilty.vrm` by Guilty, fixed wrong bone structure problem.
- Tried two versions of new blending codes, fix arm rotation problem.
- Adjust tree position in sibling pr https://github.com/webaverse/plants/pull/1 .

Before: Animation:
image

Before: Model: scilly_drophunter_v31_Guilty.vrm
image

After: Model: scilly_drophunter_v31.6_Guilty.vrm
image

After: Animation: Version1:

https://user-images.githubusercontent.com/10785634/168777324-a8f0a2cd-88ef-426b-bd8e-4f763138d8d0.mp4

After: Animation: Version2:

https://user-images.githubusercontent.com/10785634/168777363-ebf0b1b6-1195-4376-b7b5-614e6816ecf7.mp4


gonnavis:
Let `boneAttachment` support array of bones ( from one to any length ), and averagely lerp their position/quaternion/scale.
Thus can do something like place tree at the center of to hands.
Sibling commit: https://github.com/webaverse/plants/pull/1/commits/789bdb10f06d3096a6d7134d26767c5325317aa0

Works very well when walk/run/crouch I think:

https://user-images.githubusercontent.com/10785634/168899339-bec235e6-c3da-4160-83a4-9c94078299dd.mp4

But has a problem that, when throw, the tree, i.e., the averaged quaternion, will flipping.

https://user-images.githubusercontent.com/10785634/168900544-3a754691-5745-44ef-ad64-d8eb638217b4.mp4

Slow down version:

https://user-images.githubusercontent.com/10785634/168900913-2948152c-369c-44dc-8992-22ca7f5c94d1.mp4



Generally speaking, slerp of smoothly transitioning quaternion should not cause flipping.
But for the throw animation, I think encountered a special case, which the two quaternions I want to slerp, are mainly opposite, and both rotate from lower to upper at some moment.
If attach tree to only one hand separately, can see this opposition more clearly:
image
image

I think the solutions maybe:
1. Just attach to only one hand when throw.
2. Still lerp position and scale, only attach quaternion to only one hand. ( But that should make the `.metaversefile` settings to complex).
3. Shorten `throwReleaseTime` a lot.
4. Due to the whole throw animation is very short, maybe just ignore this flaw.

gonnavis:
Guidance by avaer from chat:
>The bones array seems fine
I think about the flipping you need to compute the hand quaternion better
Using eg matrix4.lookAt with a correct up vector
Up vector can be vector from avatar root to hands bone average 
So it points away from avatar
Also the throw should release the item earlier in the animation (before it completes) 
And finally the throw should not require hold, but only single tap to complete the throw

- Added quaternion type: `upVectorHipsToPosition`. If set, will use `matrix4.lookAt` to control the tree's rotation. Sibling commit: https://github.com/webaverse/plants/pull/1/commits/f208c010e378cffb11ee472ee9834fdd51b27489
- Speeded up `throwReleaseTime` from 750 to 220.
- Added one click play a full throw/attack animation logics. ( codes copy/pasted from an old PR https://github.com/webaverse/app/pull/2804 )

avaer (commented):
- the throw is very nice, but it should take into account the player's speed when they release

image

- I can't get any grab animations at all with this branch. That probably regressed recently outside this PR but needs to be fixed.
- the single-click attacks need a lot of optimization before they should be the default (they are too long, cannot be combo'd)

gonnavis:
Used textametron's new continuous combo animation.
Now no `sword_idle` pose in between attack animations.

https://user-images.githubusercontent.com/10785634/163167792-9e18c296-d919-47ae-8b71-6bb246914ee5.mp4

gonnavis:
> the throw is very nice, but it should take into account the player's speed when they release

Added player's velocity to the thrown object.

https://user-images.githubusercontent.com/10785634/169622619-782ad90d-0e41-430a-b158-1bb8c647b43b.mp4

avaer (approved):

character-controller.js
```diff
@@ -1195,6 +1210,12 @@ class LocalPlayer extends UninterpolatedPlayer {
       this.characterHups.update(timestamp);
     }
   }
+  handleAnimationEnd(e) {
```
avaer, line 1213: This eventing + multi-frame solution seems nasty. Is there no way to do it cleanly in one function?

metaverse_components/wear.js
```diff
@@ -188,26 +195,68 @@ export default (app, component) => {
...
+    const boneAttachments = Array.isArray(boneAttachment) ? boneAttachment : [boneAttachment];
+
+    // lerp app's transform to average position/quaternion/scale of boneAttachments.
+    let count = 0;
+    boneAttachments.forEach((boneAttachment, i) => {
+      const boneName = Avatar.modelBoneRenames[boneAttachment];
+      const bone = player.avatar.foundModelBones[boneName];
+      if (bone) {
+        if (count === 0) {
+          bone.matrixWorld
```
avaer, line 207: I think this could be very slow to decompose each bone. Is that really needed?

avaer:
I tried it and it works well, but some code questions + suggestions. I think the multi-frame eventing solution is a bit too complicated.

character-controller.js
gonnavis: Simplified logics: https://github.com/webaverse/app/pull/3010/commits/f596444cb71204c4cac9ad23f5bd63399b46d7eb , https://github.com/webaverse/app/pull/3010/commits/d98645faa5072b3ac63260c59f14d2d9c0b73289 .

And can even greatly simplify the logics by calling `game.handleAnimationEnd()` directly from avatar, instead of through avatar's `animationEnd` event. But not sure if this way is proper. https://github.com/webaverse/app/pull/3010/commits/f7518187c5c4f4e73b259dfcf6c55ac6970d1baa

metaverse_components/wear.js
gonnavis: It's derived from existing code https://github.com/webaverse/app/blob/295fb394e236aa40d8719ee28ec20fdc62990f72/metaverse_components/wear.js#L195 , and I think it is needed, because of will use them to do calculations later on https://github.com/webaverse/app/blob/295fb394e236aa40d8719ee28ec20fdc62990f72/metaverse_components/wear.js#L196-L204 .
And for now, need to lerp them through their lerp/slerp functions due to matrix has not lerp function.

But it's generally not a big performance issue, because of for new plants repo, will just loop two bones, not all the bones of avatar https://github.com/webaverse/plants/pull/1/files .
And for all old wearable repos, only has one single bone to `decompose`.

Deleted two other unneeded `decompose`s though https://github.com/webaverse/app/pull/3010/commits/2c96fbc466304768350e9ed88217b3d8f7fbcfa1 .

gonnavis:
> 1-While we try to equip the tree, in the first try, the tree just falls down, while it gets picked up in the 2nd try. This is the case with all trees. Is it supposed to be this way?

It's original behavior, I think it's supposed to be.
But event not supposed to be, I think will not solve this problem in this pr, due to this pr is mainly focused on animation, 

> 2-We are not covering the pickup animations while flying right?

Applied holding animation to fly.
But didn't apply thow animation, because it'll let codes way too complicated based on current animation system.
I'm refactoring animation system, should can handle this situation afterwards.

Before:
image

After:
image

> the tree is being picked up from its bottom. that makes the long trees look unrealistic being picked up.

This should to be solved in `plants` repo.

avaer (approved):

avaer:
> This should to be solved in plants repo.

Yeah the plants aren't done. This is just being used to test grab.

PR #3023, opened 2022-05-19
UI: Added modal box animation.

TheoTheDev:
Added lottie animation borders for text modal-boxes.
Maybe needs some visual tweeks, but in general works.


src/StoryTime.jsx
```diff
@@ -30,11 +34,72 @@ const MegaChatBox = ({
...
+
+    lottie.loadAnimation({
+        container: boxRightRef.current,
+        renderer: 'svg',
+        loop: false,
+        autoplay: true,
+        path: `./lottie/modal/${ type }/right.json`
+    });
+
+    if ( openTimeoutId !== null ) clearTimeout( openTimeoutId );
```
avaer, line 90: Can this be put into a returned cleanup function?

TheoTheDev: Fixed.

avaer:
image

It's a good start, but:

1) border is too cramped with text
2) the border horizontal and vertical should be the same width
3) the out animation takes too long, there it no need to wait for it to complete before closing the dialog (though we can still play it during fade-out)

avaer (commented):
Would like to merge this, but holding up the dialog box for the animation is a bit of a gameplay regression.

GitWillG:
Ran into this issue before when trying to run 'UI: Added modal box animation' PR in Dev mode?

From a fresh clone of the repo and fresh installation of both npm and Lottie-web dependancies followed by checking out PR.
image

PR #3029, opened 2022-05-19
wind

tcm390:
Add wind in the scn file. Sibling PR: https://github.com/webaverse/totum/pull/114

https://user-images.githubusercontent.com/60634884/169364223-377d6cc6-b3ad-4af5-abd2-a44f86e86a62.mp4

Just have 2 wind type now: directional and spherical (reference: https://docs.unity3d.com/Manual/class-WindZone.html)
And the other parameters are:
"direction"
"radius" (only active for spherical)
"position" (only active for spherical)
"windForce"
"noiseScale"
"windFrequency"

Issue:

1. hair would pass through face if the windForce is too large
image
2. Some characters do not have springBones


avaer:
Looks great!

avaer:
What do we need to ship this? (it's still marked as WIP)

tcm390:
noiseScale 0.1 

https://user-images.githubusercontent.com/60634884/170783059-a80787e1-71bd-4080-aa4f-e101c5c34e6e.mp4

noiseScale 0.5


https://user-images.githubusercontent.com/60634884/170783077-abefd480-157d-4731-8276-7f8c551e123c.mp4

noiseScale 1


https://user-images.githubusercontent.com/60634884/170783096-10dcb691-c70a-472b-b33d-93c58b368f22.mp4

avatars/avatars.js
```diff
@@ -81,6 +84,11 @@ const textEncoder = new TextEncoder();
 const maxIdleVelocity = 0.01;
 const maxEyeTargetTime = 2000;
 
+const simplex = new SimplexNoise();
```
avaer, line 87: We can move this winds code to a separate file to make the code cleaner. Avatars.js update function is already far too long.

avatars/avatars.js
```diff
@@ -1,4 +1,5 @@
 import * as THREE from 'three';
+import { SimplexNoise } from 'three/examples/jsm/math/SimplexNoise.js';
```
avaer, line 2: We have a different noise in procgen.js: https://github.com/webaverse/app/blob/215519e150ffa94168d52e77f9cc8b489583bb89/procgen/procgen.js#L25

I don't care which one we use but other code uses this one so we should be consistent.

avatars/avatars.js
```diff
@@ -1935,6 +1943,91 @@ class Avatar {
 
     // this.springBoneTimeStep.update(timeDiff);
     this.springBoneManager && this.springBoneManager.lateUpdate(timeDiffS);
+   
+    const _applyWind = () => {
+      const winds = metaversefile.getWinds();
+      //console.log(winds)
+      const timeS = performance.now() / 1000;
```
avaer, line 1950: This can be read from a `timestamp` passed in from `update()`.

In general we should not do `performance.now()` in our code so the timestamps for all apps are consistent, rather than slightly off.

avaer (commented):
Couple of cleanups recommended, but the code looks good.

avatars/avatars.js
tcm390: [move wind to simulation](https://github.com/webaverse/app/pull/3029/commits/4d7328e91cffc16fb3b727f991f0d77fb1237ee9)

avatars/avatars.js
tcm390: used noise from simplex-noise.js

avatars/avatars.js
tcm390: used the timestamp from avatar.js

tcm390:
{
      "type": "application/wind",
      "content": {
        "windType": "directional",
        "direction": [1, 0, 0.8],
        "windForce": 0.8,
        "noiseScale": 1,
        "windFrequency": 1
      }
    },
    {
      "type": "application/wind",
      "content": {
        "windType": "spherical",
        "direction": [0.25, 0.15, -0.25],
        "radius": 5,
        "position": [0, 0, 0],
        "windForce": 5,
        "noiseScale": 10,
        "windFrequency": 5
      }
    },

tcm390:
This is sibling PR: https://github.com/webaverse/totum/pull/114

tcm390:
And this is sibling PR: https://github.com/webaverse/totum/pull/114

avaer (approved):

avaer:
We should add an example `scn` file to the repo. It can be a copy of another scene, like the `block` in the original video.

PR #3040, opened 2022-05-20
Sword damage based on sword damageBox instead of localPlayer's position. (z-targeting version)

gonnavis:
Follow: https://github.com/webaverse/app/pull/2965
Fix: https://github.com/webaverse/app/pull/2965#pullrequestreview-980098357
Sibling PRs: https://github.com/webaverse/silsword/pull/8 , https://github.com/webaverse/street/pull/6


alisaad673:
This Seems to be working fine. The objects that comes in contact with the sword gets damaged. So if sword touches an object twice during its action, it is damaged twice. 

Just some minor reservations, 
-As with this we will be needing to apply the `vincibility: invincible` logic to all our environments, otherwise the ground would disappear, But that seems to be a must do for all as I talked to vis.
-Second thing is Right now, it is difficult to aim or damage small object such as potion, as sword action is random right now based on animations. SO we will be needing to figure something out for that too in future.

But As of now this works fine.

https://user-images.githubusercontent.com/35926530/169892389-345a6b63-a17a-4a4c-928d-aca5dc965692.mp4

gonnavis:
> As with this we will be needing to apply the vincibility: invincible logic to all our environments, otherwise the ground would disappear, But that seems to be a must do for all as I talked to vis.

Added `vincibility` logics to all ranged weapons and shooting-range scene.
See the linked 5 PRs: machine-gun, bow, pistol, uzi, shooting-range, in this issue https://github.com/webaverse/app/issues/3062

PR #3042, opened 2022-05-20
Convai PR for emotion integration.

crazylazylife:
This is a WIP PR from Convai for emotion integration.

- Currently we are successfully able to trigger emotes for localPlayer.
- Modified triggerEmote function to work for any player passed as parameter
- Complete emote based interaction in streets.scn


avaer:
This did not work.

crazylazylife:
@avaer , we just took time to confirm by cloning the codes and running them in different systems, that it is indeed producing emotions. Here is a screenshot of the street scene with the Emotions generated being console logged:
webaverse-ss1

We have the PR to `z-targeting` branch of `app` repository and also had a PR in `lore` repository from `convai-update` to the corresponding `z-targeting` branch.

Here is a link to a small video showing the emotion animations being triggered: https://drive.google.com/file/d/1_iGmgX0CqipX-DlFOFJgUHC81XNM6VdV/view?usp=sharing

We haven't updated any of our code since the PR except for fetching your commits and it is running perfectly. Can you please check once more.

alisaad673:
I tried to test it, For some reason whenever I try to talk to the npc, It throws this error. I have set the openai key and the panel is green.

image
https://user-images.githubusercontent.com/35926530/174320736-97a50e65-3b90-4622-b67d-277179967317.mp4

crazylazylife:
> I tried to test it, For some reason whenever I try to talk to the npc, It throws this error. I have set the openai key and the panel is green.. I pulled the latest though.
> 
> image
> 
>  2022-06-17.19-40-29.mp4


Hi @alisaad673 , the issue might arise due to an invalid OpenAI API key that is being set.

Also, afaik the current process of initiating conversation is by double pressing 'E' and Left-Click to start and continue with responses and chat options. Those are the prompts we updated and are being sent to OpenAI.
Let us know if this method of conversation still exists and we need to update the prompt to incorporate this.

avaer:
Looks like it's broken and can't be fixed.

src/StoryTime.jsx
```diff
@@ -14,6 +21,108 @@ const _progressConversation = () => {
   sounds.playSoundName('menuNext');
 };
 
+
```
purn3ndu, line 24: remove commented out lines.

story.js
```diff
@@ -215,8 +225,14 @@ class Conversation extends EventTarget {
     }
 
     // say the option
-    this.addLocalPlayerMessage(option, 'option');
-    
+    this.addLocalPlayerMessage(option.message, 'option');
+    //------------------- Convai ---------------------
```
purn3ndu, line 229: remove these Convai markers.

purn3ndu (commented):
cleanup changes requested.

crazylazylife:
Hi @alisaad673, can you please replace the OpenAI api key that you are using and check.
@purn3ndu I am pushing the cleanup right away.

alisaad673:
**Hey yes I made a new account for open ai and it worked so this is what I got:**

**-Right now The feature works for Automated prompts only. We can select between options in between some of the prompts.
1-It doesn't works for general chat/voicer yet.
2-As you can see there is a little delay in between the emotes and the response. I think if we can get the emotes in sync with the talk, It would be great.
3-After the end of one conversation this error popped up once. I am unable to reproduce it and it is appearing frequently. But it maybe something you know about.**

image


**I think its good as a start, but having it for general chat would be great and more useful imo. As in the prompts form, we dont have any choice of where the conversation will go.**

https://user-images.githubusercontent.com/35926530/174660353-8571fe09-c616-4726-a481-793a477a3dd6.mp4

crazylazylife:
Thanks a lot for the update. I am working on generating emotions for general chat.
I am also trying to reproduce the error that you got wrt to chat progression. In the sample videos that I have shared, I was able to trigger multiple automated conversation blocks one after another, without any issue. But I am looking into it.
As for the delay in emotion and text, I am not sure if we can call the triggerEmote function asynchronously with the text response, because some emotions also have sound effects attached to it. Please help me out here if I am missing something.

alisaad673:
I have tested it. There are a few issues:

1- The ratio that emotes existing in our list is too low. As you can see in the video below, there are few instances where it should have nodded with yes, but it didn't. It either produces sad, angry, surprised, happy, normal emotes.

2- The only emote I was able to produce that did play the emote was "angry" and "sad". Even though surprise and alert are in our list but they didn't play.

3-I guess We should point the responses that don't happen to be in our list of emotes(such as surprised, happy etc) to the ones that are in our list for now.(joy, surprise)

4- How are we catering our self character?  I mean in the prompts our character had emotes too, but in general chat, we only have emotes for npcs.


https://user-images.githubusercontent.com/35926530/175561798-c0c6090a-192b-4952-a1fa-1461328ad6c8.mp4


https://user-images.githubusercontent.com/35926530/175561855-ccadbe52-4183-45e7-9b1e-5b65555fab5a.mp4


https://user-images.githubusercontent.com/35926530/175561969-fdb6a5c3-0377-4cd3-b10b-817622d2d5db.mp4

purn3ndu:
@alisaad673 

> 1- The ratio that emotes existing in our list is too low. As you can see in the video below, there are few instances where it should have nodded with yes, but it didn't. It either produces sad, angry, surprised, happy, normal emotes.

The prompt is working as few-shot currently with 22 preset examples. Once you have a long chat where the most recent emotes are a specific set (sad, angry, surprised, happy, normal in your case), the few shot responses get biased towards those responses and do not work well for the other ones. This is a limitation of few-shot prompting and can be solved with either finetuning or by implementing a window that truncates older than X (say 5) chat utterances such that it will keep the preset examples relatively recent always. In any case, few-shot will always have a low accuracy on this objective and the proper solution would be finetuning. However, this wouldn't produce an error response only missing animation sometimes.


> 2- The only emote I was able to produce that did play the emote was "angry" and "sad". Even though surprise and alert are in our list but they didn't play.

This connects with the same problem mentioned above in #1. If you want to test out other emotions, the only way currently to experience the other emotes is to restart the conversation (once it has led to a long conversation with only a small set of emotes used recently).


> 3-I guess We should point the responses that don't happen to be in our list of emotes(such as surprised, happy etc) to the ones that are in our list for now. (joy, surprise)

Yes, a better mapping can be done for this case. However, this is likely a long tail of possible responses from GPT3 and needs to be noted and updated as experienced. However, we should add the above-noted cases in this PR cc @crazylazylife


> 4- How are we catering our self character? I mean in the prompts our character had emotes too, but in general chat, we only have emotes for npcs.

For the player(self) character, the emote input is via pressing 'Q' and choosing an emote from the emotion wheel which gets enacted immediately after clicking via mouse whereas the chat input is via voice or via text input. Both are separate ways of input and combining them would be a complex UI UX issue from what I understand. For the automated conversation(double click E), it will work fine but for general conversation(using Enter) where the player/user is in full control of the input, the emote will not go to the GPT prompt.

purn3ndu:
For the PR record, here are a few chosen inputs that have emote reactions using general input(using Enter) - https://www.youtube.com/watch?v=NFjf7PYfd5s

lalalune:
Closing in favor of https://github.com/webaverse/app/pull/3187
Since the PR merge history is trashed

lalalune:
Closed in favor of https://github.com/webaverse/app/pull/3206

PR #3046, opened 2022-05-23
Heightmap dualcontouring

aerydna:
Height map terrain demo example integrated with dual contouring

image


avaer:
This is not mergeable and isn't being worked on.

PR #3066, opened 2022-05-25
Z-target swapping

GitWillG:
- allow player to select next enemy
- allow player to detach from target based on angle


avaer:
Note there is a mob manager which knows all mobs in the scene: https://github.com/webaverse/app/blob/z-targeting/mob-manager.js

PR #3069, opened 2022-05-25
Add sword dash attack.

gonnavis:
- Play sword dash attack animation if attack when running, keep existing velocity and add some velocity at start direction, then apply damping.
- Stop WASD movement if attack when walking, will add "move CharacterController by attack animation's XZ movement" feature in another pr, so that the character doesn't stop completely when WASD movement is disabled.
- Prevent attack when jump/fly/narutoRun, will add specific attack animations for them afterwards.
- Add one click play full attack logics.

https://user-images.githubusercontent.com/10785634/170358868-2be9713e-79fe-46d2-969d-6abac4861907.mp4


avaer:
Can we smooth this more with the walk?

PR #3070, opened 2022-05-25
Dual Contouring Cached Noise

aryasalehi:
Sibling PR of : https://github.com/webaverse/app-wasm/pull/58


physx.js
```diff
@@ -2121,15 +2121,38 @@ const physxWorker = (() => {
...
-  w.createChunkWithDualContouring = (x, y, z, lod) => {
+  w.generateChunkDataDualContouring = (x, y, z) => {
+    moduleInstance._generateChunkDataDualContouring(x, y, z)
+  }
+
+  w.setChunkLodDualContouring = (x, y, z, lod) => {
+    moduleInstance._setChunkLodDualContouring(x, y, z, lod)
+  }
+
+  w.clearTemporaryChunkDataDualContouring = () => {
```
avaer, line 2132: It is not a good idea to have a singleton like this.

it you want to multi-phase these calls then return an instance of the dual contouring class and handle cleanup in a destructor.

This implementation is hard to follow and hard to parallelize.

avaer (commented):
This pr has some  api problems but I will merge anyway to test.

PR #3076, opened 2022-05-27
3074 UI unable to scroll click nfts in sidebar

### LINKED ISSUE #3074: UI: Unable to scroll / click NFTs in sidebar
When signing in via metamask I am unable to scroll up n down on my inventory on the right sidebar nor click any of my NFTs to spawn in-world

mwmwmw:
Fixes #3074 

With this fix, you can scroll and click the NFTs. You can drag and drop the NFTs to the world and it will upload the application metadata. 

But, that upload doesn't put anything in the world.

### DIFF
--- a/src/Header.jsx
+++ b/src/Header.jsx
@@ -292,7 +292,7 @@ export default function Header() {
                     /> */}
                 </div>
                 <div className={styles.panels}>
-                <Tokens
+                    <Tokens
                         nfts={nfts}
                         hacks={hacks}
                         address={address}
--- a/src/Header.jsx
+++ b/src/Header.jsx
@@ -290,7 +290,9 @@ export default function Header() {
                         claims={claims}
                         panelsRef={panelsRef}
                     /> */}
-                    <Tokens
+                </div>
+                <div className={styles.panels}>
+                <Tokens
                         nfts={nfts}
                         hacks={hacks}
                         address={address}
--- a/src/Header.module.css
+++ b/src/Header.module.css
@@ -624,10 +624,6 @@ input[type=range]::-webkit-slider-thumb {
 .tabs {
   display: flex;
   position: fixed;
-  top: 0;
-  bottom: 0;
-  height: 100vh;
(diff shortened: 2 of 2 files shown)


src/Header.jsx
```diff
@@ -290,7 +290,9 @@ export default function Header() {
                         claims={claims}
                         panelsRef={panelsRef}
                     /> */}
-                    <Tokens
+                </div>
+                <div className={styles.panels}>
```
avaer, line 294: This isn't properly formatted.

src/Header.module.css
```diff
@@ -542,7 +542,7 @@ input[type=range]::-webkit-slider-thumb {
   background-color: #FFF;
   transform: translateX(400px);
   overflow: auto;
-  z-index: 1;
+  z-index: 3;
```
avaer, line 545: Why do we need another crazy z-index hack?

mwmwmw: I agree, there is way too much z-indexing going on. Also, there is a lot of absolute positioning where a simple flexbox or grid would be a better choice. 

The cleanest way to fix it is to reorder the dom hierarchy. 

It's quite a few minor code changes to do it. I think it would be better to do them in a separate PR as it isn't specifically related to this fix.

src/Header.jsx
mwmwmw: fixed

src/Header.module.css
mwmwmw: https://github.com/webaverse/app/pull/3085

mwmwmw: I closed #3085 because the fixes to the dom are no longer relevant on `z-targeting` and it'll just create a ton of conflicts. 

I'll do the z-index overhaul in `z-targeting`  as part of this issue #3087 

this PR merges into `z-targeting` just fine as is.

avaer (approved):

avaer:
@alisaad673 Yes those design updates should be a separate issue + PR.

PR #3077, opened 2022-05-28
Add backflip jump.

gonnavis:
https://user-images.githubusercontent.com/10785634/170805302-530d35ac-4741-4b27-b7be-ed071b1e125b.mp4


avaer:
It looks good, but IMO it's better faster, maybe even 2x, since this is usually meant as a quick dodge.

avaer:
Alternatively, for this speed of backflip, I think the character would need to jump much higher.

gonnavis:
Because jump air time is defined by gravity and jump velocity.y, and I think can't modify gravity.
So then can just decrease jump velocity.y to decrease jump air time.
( Also can't increase jump velocity.y and height, will increase jump air time, which not expected )

How about this version? ( Decreased jump air time from 999ms to 600ms )

https://user-images.githubusercontent.com/10785634/170833668-84b167da-7377-47e8-a9ec-e6168147cbae.mp4

gonnavis:
Used existing `game.isMovingBackward()` to check if backflip instead of check in separated codes.

Found a problem that, 
though this PR https://github.com/webaverse/app/pull/2799 used `avatar.direction.z` instead of `keysDirection.z` , 
successfully solved the inconsistent velocity issue https://github.com/webaverse/app/issues/2798 .
But compare to instant `keysDirection.z`, `avatar.direction.z` has a very short lag due to it's frame comparing and normalizing https://github.com/webaverse/app/blob/0b710da77b85c798e1f34a7c530bba8503017a65/avatars/avatars.js#L1495-L1507
Will cause consecutive aimming jumps appear unexpected backflip.

So I introduced `keysDirectionLocal` to be used for `game.isMovingBackward()`, in order to solve the problem of wrong velocity while maintaining instant keyboard input reaction.

avaer:
> https://github.com/webaverse/app/pull/3077#issuecomment-1140291165

The version here looks ok, but it could be smoothed in more. (the animation jumps to hands up immediately)

gonnavis:
Added transition at the start of backflip animation.

gonnavis:
> 2- As you can see, the backflip works and looks fine while aiming and walking/running backwards. But while we aim and do naruto run backwards, the direction of avatar is not towards backside, but the back flip jump happens and looks kind of weird. Either we should disable backflip jump for naruto run or we should make sure that direction of avatar while aiming and moving backward is towards the front.

Hello @alisaad673 , fixed this bug, and one more "can't moving when flying" bug.

alisaad673:
> > 1-These errors are appearing in the branch, Are they related?
> 
> I think they are not related to this PR. Do you have these errors in pure newest `z-targeting` branch?
> 
These error are not appearing in your other PR https://github.com/webaverse/app/pull/3099. 
Can you please check this out? Also I didn't see these in z-targeting too. Maybe try rebasing?

The other bug is fixed.

gonnavis:
Hello @alisaad673 , I think the main error is `ReferenceError: Cannot access 'cameraManager' before initialization at io-manager.js` right?
I can't reproduce it on my local, but fixed a potential `import` problem which may be the cause.
Is it fixed now?

alisaad673:
Yes. This is fixed now.

lalalune:
Hello @alisaad673 and @gonnavis is this ready to be pulled in?

gonnavis:
> Hello @alisaad673 and @gonnavis is this ready to be pulled in?

Yes, ready for merge or more reviews.

lalalune:
I tested this, it's good. If you z-target and jump backwards you'll backflip nicely. Going to pull it in. The other jump PR will been to be rebased on top of this.

PR #3078, opened 2022-05-28
Dual Contouring Biomes Weights

aryasalehi:
Sibling PR of : https://github.com/webaverse/app-wasm/pull/59


avaer (changes requested):
IMO this entire approach is wrong, the whole `BufferManager` class is bad since it ignores the constructor. Adding more special cases that ignore the constructor is not the way to go.

This class has already wasted hours of dev time with its bugs.

Thanks for trying, I'll handle this.

Issue #484, opened 2021-01-02
Jump animation loop


avaer:
The jump animation currently runs all the way through to standing, even when the user is still jumping.

We'll want to loop the jump part of the animation until the user lands.

msm0823:
https://github.com/webaverse/app/issues/3093

PR #3080, opened 2022-05-28
Split jump animation to `jump start`, `fall loop` , `land` three stages.

gonnavis:
Fix: https://github.com/webaverse/app/issues/484

https://user-images.githubusercontent.com/10785634/170843102-dae5dc94-7812-48ee-8105-e869b5432246.mp4

Split jump animation to `jump start`, `fall loop` , `land` three stages all by codes now.
If this PR is ok, I'll request artist to remake the `fall loop` animation base on this frame ( frame 39 ):
image
But maybe the current static version is already better, can be used as first version directly.


avaer:
It's not had, but the fall loop has only one frame?

avaer:
I don't think we should be copying the existing hack. The animation was never good in the first place.

I was expecting a new set of animations that make a lot more sense in fast combat.

avaer:
Additionally, I do think when we redo these jump animations, we should have a good fall loop in it (not static).

avaer:
Gonna close this since I don't think it's worth working on or shipping until we have good animations that are not based on the existing.

gonnavis:
Got it, I'll apply art resources to do a new animation and give some ideas of my own.

avaer:
Thanks! The code for unjump is probably usable too.

Some other points about jumping which might be related:
- The plan for naruto run is that it greatly increases your jump height and gravity, so you can easily jump over tall obstacles with it.
- I think we will have a "stomp" attack, which immediately brings you out of the jump by pressing C in midair.

Both of these will need to interact with the jumps in some way so it is worth factoring it into the animations and animations style.

The above is also the reason the existing Mixamo jump animation is really not good, since it is slow and relaxed, and not really compatible with these kinds of fast gameplay moments.

gonnavis:
Hello @avaer , texta has already made one new jump animation, how about it?

https://user-images.githubusercontent.com/10785634/170889857-703a13bc-a877-4a0e-b046-d96da5c4fdb2.mp4

PR #3082, opened 2022-05-29
Pick up tree animation: Add one click logic.

gonnavis:
Fix: https://github.com/webaverse/app/pull/3010#issuecomment-1130310052 
>And finally the throw should not require hold, but only single tap to complete the throw

Related: https://github.com/webaverse/app/pull/3010#issuecomment-1139039858
>because of one click logics' structure is a blocker of this pr, so removed it first.
Will re-add it after refactored animation system.

I'm trying to refactor the animation system which each `action`/`motion` is a class instance and has `animationEnd`/`finish` event.

But I want to try another way in parallel that, because of we know the duration of `pickUpThrow` and other animations,
so `animationEnd`/`finish` event maybe not a must need.
Just check if `useTime` larger than known duration to run `_endUse()` is ok.
This is what this PR do.


avaer (approved):
LGTM 👍

PR #3097, opened 2022-05-31
One click play a full attack animation & continue combo. ( 2nd try )

gonnavis:
Follow: https://github.com/webaverse/app/pull/2804 , https://github.com/webaverse/app/pull/2804#issuecomment-1096711630 , https://github.com/webaverse/app/pull/3082

- Apply the same one-click logic ( check known animation duration in `update()` ) as https://github.com/webaverse/app/pull/3082 to combo attacks, instead of `animationEnd` events.
- Update to continuous combo attack animations, which compatible with this logic.
- Use `avatar.useAnimations.pickUpThrow.duration` instead of `const throwAnimationDuration`.


msm0823:
The avatar is unable to move in street scene. You can see even the dash function is not working properly. Can't pick the sword from street scene.



https://user-images.githubusercontent.com/104460803/171660599-cf36e67a-5a1a-4263-88c7-836e5cafdb72.mp4

msm0823:
After switching the scene, I'm able to equip Bow from the tree house scene. The functionality is not working on the bow weapon.

https://user-images.githubusercontent.com/104460803/171660834-6b2e0d9a-b0ce-4c07-a917-97363f2e6c4f.mp4

gonnavis:
@msm0823 Thanks for testing!
Sorry, I have not emphasised that, the one click logic is only applied to sword and silsword in this PR.
So it's correct that the bow and other weapons keep the existing behavior.

msm0823:
> @msm0823 Thanks for testing! Sorry, I have not emphasised that, the one click logic is only applied to sword and silsword in this PR. So it's correct that the bow and other weapons keep the existing behavior.

Got it! Thanks for the clarification!

PR #3098, opened 2022-06-01
apply scale to all selected physics objects

### LINKED ISSUE #3096: app crashes while trying to edit certain objects in world tokens
Right now, After theo's fix to world tokens issue, Most of the items work fine(we are able to edit them), But certain items gets the app crashed when we click on them.Throws these errors. For example, skybox, application/light in street scn, plaza in plaza.scn.

world1
world2

https://user-images.githubusercontent.com/35926530/171255128-4fbe7733-3725-447e-973a-0039da27cb16.mp4

mwmwmw:
Fixes #3096 

This PR fixes two issues. 

1) selecting objects without physics would crash because the zero index was hardcoded.
2) editing objects with more than one physics object would only update the scale of the first object.

### DIFF
--- a/src/components/general/world-objects-list/WorldObjectsList.jsx
+++ b/src/components/general/world-objects-list/WorldObjectsList.jsx
@@ -106,8 +106,9 @@ export const WorldObjectsList = () => {
     const updatePhysics = () => {
 
         const physicsObjects = selectedApp.getPhysicsObjects();
-        physicsManager.setGeometryScale( physicsObjects[0].physicsId, selectedApp.scale );
-
+        physicsObjects.forEach( ( physicsObject ) => {
+            physicsManager.setGeometryScale( physicsObject.physicsId, selectedApp.scale );
+        });
     };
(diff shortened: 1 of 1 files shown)


alisaad673:
Seems to work fine..

alisaad673:
The issue seems to be fixed but there is some issue with the scale. 
When we increase the scales of glb objects, sometimes it works fine, but in some cases the scale isn't updating as it can be seen in these two examples:
-plaza in plaza.scn
-Lootpod in spacepod.scn

https://user-images.githubusercontent.com/35926530/171464216-00adea03-d4cc-4bbd-9a80-f7a99cd7925d.mp4


https://user-images.githubusercontent.com/35926530/171464626-915509f4-166e-4052-ad95-9963d3cfb7f6.mp4

avaer (approved):

avaer:
@alisaad673 Can you open an issue for that? It seems not related to this PR.

PR #3099, opened 2022-06-01
Add new jump animation.

gonnavis:
https://user-images.githubusercontent.com/10785634/171347815-476aa148-7dea-4ba9-a399-981184663870.mp4

Used new three stages ( jump / falling / landing ) jump animation files made by textametron( @plankatron ).

Reduced jump air time from 999ms to 682ms ( measured when jumping from flat ground ).
Because we want shorter air time, but higher jump height. I think can't be based on real -9.8 gravity anymore.
So I increased the velocity of the jump from 6 to 15, and adjusted the gravity to 4 times the value.
But can't keep such a high gravity value all the time, otherwise will cause the character to fall too fast when jumping from higher to lower, so I reset the gravity value back to normal after 682ms.

I tried that, can't simply blend running with current landing, we'll do a specific run jump/land version later.


avaer:
This is a draft. Should it be reviewed?

gonnavis:
Updated, ready for review now.

- Aesthetic jump movement by sinusoidally `moveCharacterController`, instead of change `gravity`.
- Add `fallLoop` & `land` actions.
- Update `jump`/`fly` logics, prevent trigger jump when flying.
- Use `landing.fbx` for non-movement landing, use/blend `landing 2.fbx` for lending with walk/run.

https://user-images.githubusercontent.com/10785634/174206624-f9c16544-e522-4db9-aa4c-a98917924dc7.mp4

gonnavis:
@alisaad673 Fixed.

gonnavis:
> 1-The landing speed feels a little fast. Can we reduce it a little bit to make the landing look smooth?

Ok, we'll adjust the landing animation.

> 2-Can you disable jumping while landing from flying? I mean once we are flying and press f, we can still jump in the air without touching the ground.

Great catch! Sorry I haven't found this bug myself, I'll fix it.

alisaad673:
It Seems fine now.

alisaad673:
Yep. Works Fine Now.

lalalune:
Hello @alisaad673 @gonnavis is this good to pull in?

gonnavis:
> Hello @alisaad673 @gonnavis is this good to pull in?

Yes, ready for merge or more reviews.

lalalune:
Hey @gonnavis --
I merged your backflipping PR before this, however that caused conflicts here. Can you please clean up the conflicts and we can do another QA pass? Thanks.

gonnavis:
@lalalune OK, sure! I'll resolve.

gonnavis:
Hello @lalalune . Resolved conflicts. ( Because of the reverting of backflip pr, no much conflicts need to be resolved though )

lalalune:
@avaer @gonnavis how we doing with this PR? It looks awesome. Even if it's something that needs more work, practically we might want to pause it to get some other things further along.

gonnavis:
@lalalune Let me retarget to `master` branch instead of `z-targeting` now and see if any new conflicts.

PR #3106, opened 2022-06-03
Improve use bow when walking animations. ( 2nd try )

gonnavis:
Follow: https://github.com/webaverse/app/pull/2903

Hello @avaer , sorry that, I found I made a big mistake/regression in the final codes of that old PR https://github.com/webaverse/app/pull/2903
So the result is not the same as the video I posted in that old PR.
This maybe why you feel "the top is stiff" https://github.com/webaverse/app/pull/2903#pullrequestreview-968549619
Now in this PR, I fixed the mistake/regression, made some more improvements, and made it compatible with z-targeting branch.

Old PR wrong result, top very stiff:

https://user-images.githubusercontent.com/10785634/171775146-061af342-8be0-448b-b195-3a21c2819dda.mp4

This new PR:

https://user-images.githubusercontent.com/10785634/171775273-865233a2-9018-4449-bb02-1b4535153541.mp4

How about this new result, is it still stiff or ok?

---

https://github.com/webaverse/app/pull/2903#pullrequestreview-968549619
> I also think the slow walk draw system is easy to break (e.g. do naruto run while drawing the bow and it will run in place).

I prevented do naruto run while drawing in the PR, now should the very same slow walk logics as crouch.
And I'll request art assets of bow with running, so should not need do slow walk at all afterwards.
But for this PR, I want focus on bow with walking first, is it ok?


avaer:
Keep in mind the most important part of this is the mouse aim. It's not shown in any of these demos.

PR #3148, opened 2022-06-20
New Readme

lalalune:
Adds new readme and some images


lalalune:
@avaer this is ready for review

PR #3149, opened 2022-06-21
VR new

Loryhoof:


io-manager.js
```diff
@@ -195,10 +190,45 @@ const _updateIo = timeDiff => {
...
+            game.menuMouseDown();
+          } else {
+            game.menuMouseUp();
+          }
+
+          if (buttons[1] >= 0.5) {
+            game.menuActivateDown();
           }
+
+          if (buttons[5] >= 0.5 && !ioManager.lastButtons[5]) {
```
avaer, line 206: The buttons should not be queried directly like this. It should be mapped per controller.

avaer:
I noticed that the player position is not the headset position, so player-position based things like loading a dynamic world do not work.

webaverse-render-pass.js
```diff
@@ -28,24 +29,49 @@ class WebaverseRenderPass extends Pass {
...
+        
+        // renderer.clear();
+        // renderer.render(scene, camera);
+        // renderer.clear();
+        // renderer.render(sceneHighPriority, camera);
+        // renderer.clear();
+        // renderer.render(sceneLowPriority, camera);
+        //console.log(renderTarget);
+        renderer.clear();
+        renderer.setRenderTarget( renderer.getRenderTarget() );
```
avaer, line 55: Is this completely bypassing the postprocessing? This seems the wrong place to hook VR. It should be dowstream of this.

xr-manager.js
```diff
@@ -0,0 +1,210 @@
...
+
+      localMatrix.copy(xrCamera.matrix)
+        .premultiply(dolly.matrix)
+        .decompose(localVector, localQuaternion, localVector2);
+
+      //newVec.set(localPlayer.position.x, localVector.y, localPlayer.position.z);
+      //console.log(localPlayer.position.y, xrCamera.position.y);
+      
+      localPlayer.avatar.setLocalAvatarPose([
+        [localVector.toArray(), localQuaternion.toArray()],
```
avaer, line 202: Don't create new objects each frame, especially in VR.

PR #3160, opened 2022-06-22
WASM version of Refactor Animation System. [WIP]

gonnavis:
Follow: https://github.com/webaverse/app/pull/3116
Sibling PR: https://github.com/webaverse/app-wasm/pull/63


avaer:
@gonnavis if this is the branch we are working on it needs a rebase.
If you have a better branch, please kindly close this!

gonnavis:
Should restored all existing animations into this WASM version, and added some new ones such as dash attack and new fly animations.

https://user-images.githubusercontent.com/10785634/183092906-7c80d03f-de72-41d6-ad0c-844736d3c1f0.mp4

avatars/animationHelpers.js
```diff
@@ -37,34 +38,35 @@ import {
...
-const localQuaternion6 = new Quaternion();
+// const localQuaternion = new Quaternion();
+// const localQuaternion2 = new Quaternion();
 
-const identityQuaternion = new Quaternion();
+// const identityQuaternion = new Quaternion();
+
+const isDebugger = false; // Used for debug only codes.Don’t create new data structures on the avatar, to not add any more gc sweep depth in product codes.
 
 let animations;
```
avaer, line 59: Is there any way we can delete all of this local animations initialization?

It seems all of this can be done in WASM now and the JS doesn't even need to parse it?

avaer: If there is some code that needs to read animations in JS then maybe we can create an API for doing that, but don't parse all animations in JS for all avatars.

avaer: This would also make this file much smaller and easier to reason about.

gonnavis: Got it, I'll try to delete them.

gonnavis: >Is there any way we can delete all of this local animations initialization?
It seems all of this can be done in WASM now and the JS doesn't even need to parse it?

Yeah, don't need now. Removed the code for parsing `AnimationClip`s and creating `interpolant`s. https://github.com/webaverse/app/pull/3160/commits/f1649da5ff689c237233d1012b25b9c4f83bce8b

>If there is some code that needs to read animations in JS then maybe we can create an API for doing that, but don't parse all animations in JS for all avatars.

Now using `animationsJson.animations` as `animation` directly, instead of parse into `animationClip`s.
And decorate `animationsJson.animations` directly too.
The only place I found is the need of changing `animation.interpolants` to `animation.tracks`. https://github.com/webaverse/app/pull/3160/commits/f1649da5ff689c237233d1012b25b9c4f83bce8b
Elsewhere work fine, the infos of `animationsJson` is enough ( and mostly the same ).

>This would also make this file much smaller and easier to reason about.

Have not make this file much smaller.
Because of the need to support `.metaversefile` configs, can't hardcode animations' names, so most codes still there.

character-controller.js
```diff
@@ -991,13 +993,18 @@ class UninterpolatedPlayer extends StatePlayer {
...
       pickUp: new InfiniteActionInterpolant(() => this.hasAction('pickUp'), 0),
       unuse: new InfiniteActionInterpolant(() => !this.hasAction('use'), 0),
       aim: new InfiniteActionInterpolant(() => this.hasAction('aim'), 0),
       aimRightTransition: new BiActionInterpolant(() => this.hasAction('aim') && this.hands[0].enabled, 0, aimTransitionMaxTime),
       aimLeftTransition: new BiActionInterpolant(() => this.hasAction('aim') && this.hands[1].enabled, 0, aimTransitionMaxTime),
       narutoRun: new InfiniteActionInterpolant(() => this.hasAction('narutoRun'), 0),
       fly: new InfiniteActionInterpolant(() => this.hasAction('fly'), 0),
+      flyDash: new BiActionInterpolant(() => {
+        const ioManager = metaversefile.useIoManager();
+        return this.hasAction('fly') && ioManager.keys.shift;
```
avaer, line 1006: In general, can we avoid testing `ioManager` in character controller?

I think for supporting NPCs we can't really check the `ioManager`, since that is tied to the local user input, and it will interfere with other characters.

avatars/animationHelpers.js
```diff
@@ -412,1069 +423,789 @@ export const loadPromise = (async () => {
...
+    const backwardFactor = 1 - MathUtils.clamp((Math.PI - Math.abs(angle)) / (Math.PI / 2), 0, 1);
+    const leftFactor = 1 - MathUtils.clamp(Math.abs(angle - Math.PI / 2) / (Math.PI / 2), 0, 1);
+    const rightFactor = 1 - MathUtils.clamp(Math.abs(angle - -Math.PI / 2) / (Math.PI / 2), 0, 1);
+    const mirrorFactorReverse = 1 - avatar.mirrorFactor;
+    const mirrorLeftFactor = avatar.mirrorFactor * leftFactor;
+    const mirrorRightFactor = avatar.mirrorFactor * rightFactor;
+    const mirrorLeftFactorReverse = mirrorFactorReverse * leftFactor;
+    const mirrorRightFactorReverse = mirrorFactorReverse * rightFactor;
+
+    physx.physxWorker.setWeight(avatar.walkForwardMotionPtr, forwardFactor);
```
avaer, line 895: I have two concerns with all of these keys:

1) this makes the avatar object bigger, not smaller, so it somewhat defeats the point of doing the animation in WASM
2) this is doing many calls to WASM per frame, per avatar. But it could just be 1.

Instead of this, we can just keep the whole binding vector in a preallocated `ArrayBuffer`, fill it in, and call `.update()` once to absorb it on the WASM side.

avatars/animationHelpers.js
```diff
@@ -412,1069 +423,789 @@ export const loadPromise = (async () => {
...
-      dst,
-      // isTop,
-      lerpFn,
-      isPosition,
-    } = spec;
-
-    _getHorizontalBlend(k, lerpFn, isPosition, dst);
+  // end util ---
+
+  const createMotions = () => {
```
avaer, line 542: Since this setup seems constant for avatars, it does not make much sense to do it in JS. It's just extra overhead, initialization, and GC pressure per avatar. The whole point of doing it in WASM is to remove this overhead from the JS side.

The node system is a great design, but the same thing could be set up more efficiently in C++ instead, and I don't expect many people will be modifying this on the JS side. This is highly entangled with the avatar system anyway.

gonnavis: Deleted all motions/nodes on JS side, moved to WASM side.

avatars/animationHelpers.js
gonnavis: 1. Deleted all `motions/nodes` on JS side, moved to WASM side.

2. `setWeight/Factor/speed` etc all on WASM side, triggered by two js call `updateAvatar` ( for number/floats ) and `updateAvatarString` ( for name/strings ), data passed through preallocated `scratchStack`. ( Should can serialize floats and strings at once, then just need one call, keep improving )

physx.js
```diff
@@ -2230,10 +2232,378 @@ const physxWorker = (() => {
...
+  }
+  w.updateAvatar = (animationAvatarPtr, values) => {
+    values.forEach((value, i) => {
+      scratchStack.f32[i] = value;
+    })
+    Module._updateAvatar(
+      animationAvatarPtr, scratchStack.ptr,
+    )
+  }
+  w.updateAvatarString = (animationAvatarPtr, strings) => {
```
avaer, line 2251: What is the purpose of this "string" method?

gonnavis: I need transfer `float` and `bool` data such as `idleWalkFactor` `jumpStart` to wasm at once, I use `scratchStack.f32` in `updateAvatar()` to do this.

Also need transfer `string` data such as `useAnimation` `emoteAnimation` to wasm at once, I use `scratchStack.u8` in `updateAvatarString()` to do this.

Ideally to merge them into one call, let `scratchStack.u8` string data follow after `scratchStack.f32` float/bool data. I'll try.

gonnavis:
Closed in favor of https://github.com/upstreet-labs/app/pull/82

character-controller.js
gonnavis: Calc `mirrorFactor` from `avatar.getAngle()` instead of `ioManager`. https://github.com/upstreet-labs/app/pull/82/commits/74a40028a891eaabed973c32bb02b52c5498bd2a

gonnavis: Update, now calculating `mirrorFactor` using original methods, from `backwardAnimationSpec`, instead of from `avatar.getAngle()` directly. https://github.com/upstreet-labs/app/pull/82/commits/58797c96e2ede64d7eafdd62e2d0ca21592829a0

PR #3165, opened 2022-06-24
nexus nft blockchain

mwmwmw:
- Hooks for web3 account, NFT contract, and chains
- Providers for account context and chain context
- util addition for uploading blobs straight from the canvas
- ObjectPreview forwards a ref (used for user-created GLB previews)
- Decoupling of component and state
- New contract ABI deployments for polygon


alisaad673:
1-After logging in, Address doesn't shows up. Though the user gets logged in to their metamask wallet and is able to mint on mumbai chain.

https://user-images.githubusercontent.com/35926530/175954416-def106f7-82a4-4e12-9697-7beb25573d1a.mp4

2-We are only able to mint on Mumbai chain for now. I guess the contract aren't deployed on polygon mainnet yet. Right? On mainnet and webaverse sidechain, It throws this error.

image


3-After minting on mumbai chain, even after the minting is successful, it still keeps showing this minting status bar.(Its quite misleading so we should remove it.)

4- The minting complete notification comes before the nft is actually minted and is available in inventory.

PR #3168, opened 2022-06-25
Z targeting multiplayer 2

lalalune:
This adds multiplayer and addresses the review by @avaer -- all blockchain code and unnecessary changes have been removed, so this just focuses on multiplayer fixes.

Known issues:
Players are at 0,0,0 until they send movement, so an afk player's avatar might just be in the ground
Player update is out of sync with vehicle, so player rubber bands around.


avatars/microphone-worklet.js
```diff
@@ -1,63 +1,62 @@
...
         if (args.emitBuffer !== undefined) {
-          emitBuffer = args.emitBuffer;
+          this.emitBuffer = args.emitBuffer;
         }
       }
     });
     this.port.start();
   }
   process(inputs, outputs) {
+    try {
```
avaer, line 34: Is this try catch needed? If there is a bug, it should be fixed. If there is no bug then we can let it crash.

avatars/animationHelpers.js
```diff
@@ -1233,7 +1233,7 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
       let defaultAnimation = 'grab_forward';
 
       const activateAction = localPlayer.getAction('activate');
-      if (activateAction.animationName) {
+      if (activateAction && activateAction.animationName) {
```
avaer, line 1236: Why would there be no `activateAction` but there is `activateTime`? This sounds like some code fails to clean up.

lalalune: It crashes, this patches it but doesn't address root. Will look into it.

avatars/avatars.js
```diff
@@ -1418,6 +1421,24 @@ class Avatar {
...
+    // Set the velocity, which will be considered by the animation controller
+    const positionDiff = localVector.copy(lastPosition)
+      .sub(currentPosition)
+      .divideScalar(timeDiffS)
+      .multiplyScalar(0.1);
+    localEuler.setFromQuaternion(currentQuaternion, 'YXZ');
+    localEuler.x = 0;
+    localEuler.z = 0;
+    localEuler.y += Math.PI;
+    localEuler2.set(-localEuler.x, -localEuler.y, -localEuler.z, localEuler.order);
```
avaer, line 1435: FYI you don't need a second euler for this. You are already mutating the existing one anyway.

avatars/avatars.js
```diff
@@ -1492,28 +1513,21 @@ class Avatar {
     this.aimLeftFactorReverse = 1 - this.aimLeftFactor;
 
     const _updateHmdPosition = () => {
```
avaer, line 1515: If this is for local players, why not check `.isLocalPlayer`? I don't think it's a good idea to bleed update logic to the caller.

avatars/avatars.js
```diff
@@ -1989,40 +2009,68 @@ class Avatar {
...
 
       this.audioRecognizer = new AudioRecognizer({
         sampleRate: audioContext.sampleRate,
       });
+
       this.audioRecognizer.addEventListener('result', e => {
         this.vowels.set(e.data);
       });
+
+      this.audioWorker = new MicrophoneWorker({
```
avaer, line 2039: Why is there a second `MicrophoneWorker`?

avatars/avatars.js
```diff
@@ -1989,40 +2009,68 @@ class Avatar {
...
+        this.microphoneWorker.addEventListener('volume', _localVolume);
+        this.microphoneWorker.addEventListener('buffer', _localBuffer);
+      }
     } else {
       this.volume = -1;
     }
   }
-  getAudioInput() {
-    return this.microphoneWorker && this.microphoneWorker.getInput();
+  getAudioInput(useMicrophone = false) {
```
avaer, line 2072: I don't think this should have an argument. If there are different nodes/different ways of inputting data to the avatar then there should be another method for grabbing an alternative node, rather than an argument switch.

avaer: But I still don't get why there are two in the first place.

character-controller.js
```diff
@@ -80,20 +82,20 @@ function makeCancelFn() {
 const heightFactor = 1.6;
 const baseRadius = 0.3;
 function loadPhysxCharacterController() {
-  const avatarHeight = this.avatar.height;
-  const radius = baseRadius/heightFactor * avatarHeight;
-  const height = avatarHeight - radius*2;
+  const avatarHeight = this.avatar?.height || 1;
```
avaer, line 85: If there is no avatar then this will create an incorrect collider since you don't know the size.

I can see why players would be initialized without avatars, but that means when the avatar pops in we need to set the capsule size correctly (we may need to remake it).

character-controller.js
```diff
@@ -25,41 +27,41 @@ import {
...
 // const localQuaternion2 = new THREE.Quaternion();
 const localMatrix = new THREE.Matrix4();
 const localMatrix2 = new THREE.Matrix4();
-const localArray3 = [0, 0, 0];
-const localArray4 = [0, 0, 0, 0];
 
 const zeroVector = new THREE.Vector3(0, 0, 0);
 const upVector = new THREE.Vector3(0, 1, 0);
 
+const controlActionTypes = ["jump", "crouch", "fly", "sit"];
```
avaer, line 63: Prefer single quotes.

avatars/avatars.js
avaer: It's to handle the local mouth flap case which should not have audio feedback.

avatars/avatars.js
avaer: Sounds like this should be `getLocalAudioInput` + `getRemoteAudioInput`.

The former would be for the microphone. The latter would also be for the case of the local hups talking.

avatars/avatars.js
avaer: Probably a better name for these workers would be `localAudioWorker` and `remoteAudioWorker`.

avatars/avatars.js
```diff
@@ -1971,12 +1985,18 @@ class Avatar {
...
   }
-
   isAudioEnabled() {
-    return !!this.microphoneWorker;
+    return !!this.audioWorker;
+  }
+  isMicrophoneEnabled() {
+    return this.localPlayer && !!this.microphoneWorker;
   }
   setAudioEnabled(enabled) {
```
avaer, line 1994: IMO this should be split into the two distinct pipeline cases, and two methods. Otherwise it's confusing that multiple workers (some of which are not used) are created.

The caller should call either `setLocalAudioEnabled` or `setRemoteAduioEnabled`.

The two methods could still share the `audioRecognizer` which is created/deleted on demand.

avatars/avatars.js
```diff
@@ -2080,21 +2128,14 @@ class Avatar {
   } */
 
   destroy() {
-    this.setAudioEnabled(false);
+    // this.setAudioEnabled(false);
```
avaer, line 2131: We probably _do_ want to destroy the audio workers along with the player.

character-controller.js
avaer: If this is handled, please link the place where it's handled.

If not it needs to be handled.

character-controller.js
```diff
@@ -249,43 +275,72 @@ class PlayerBase extends THREE.Object3D {
...
     } else {
       this.voiceEndpoint = null;
     }
     this.updateVoicer();
   }
+
   getVoice() {
-    return this.voiceEndpoint || this.voicePack || null;
+    if (!this.voiceEndpoint) console.error("this.voiceEndpoint is null")
+    return this.voiceEndpoint || this.voicePack || console.warn("no voice endpoint set");
```
avaer, line 324: This should throw, not attempt to return.

character-controller.js
```diff
@@ -296,46 +351,83 @@ class PlayerBase extends THREE.Object3D {
...
-      for (const action of this.getActionsState()) {
-        if (action.type === 'wear') {
-          usedIndexes[action.loadoutIndex] = true;
+  updatePhysics(timestamp, timeDiff) {
+    if (this.avatar) {
+      const timeDiffS = timeDiff / 1000;
+      this.characterPhysics.update(timestamp, timeDiffS);
+    }
+  }
+  handleWearUpdate(app, wear, loadoutIndex = -1, isAppUpdate = false, isPlayerUpdate = true) {
```
avaer, line 361: The arguments to this function are a mess.

When it's this many it should probably be a dict.

character-controller.js
lalalune: This ordering problem probably related to why character starting below floor when landing in world. Will look into it.

character-controller.js
```diff
@@ -347,8 +439,12 @@ class PlayerBase extends THREE.Object3D {
         if (world.appManager.hasTrackedApp(app.instanceId)) {
           world.appManager.transplantApp(app, this.appManager);
         } else {
-          // console.warn('need to transplant unowned app', app, world.appManager, this.appManager);
-          // debugger;
+          console.warn(
```
avaer, line 442: If this case is hit I recommend a `throw` so it can be debugged.

character-controller.js
avaer: Also this seems very internal to it can be made private (`#`) to ensure no external callers.

character-controller.js
```diff
@@ -144,41 +145,66 @@ class PlayerHand extends THREE.Object3D {
     this.enabled = false;
   }
 }
-class PlayerBase extends THREE.Object3D {
-  constructor() {
+class Player extends THREE.Object3D {
+  constructor({
+    name = defaultPlayerName,
+    playerId = makeId(5),
+    playersArray = new Z.Doc().getArray(playersMapName),
```
avaer, line 152: IMO `Player` is too complicated to be taking in bare network state like this.

I tried to contain the network sync code to `bindState` but IMO this makes it worse -- now anyone who wants to make a player is forced to think about network state, which I don't think is the correct abstraction.

[We already contain the player -> avatar binding in its own file](https://github.com/webaverse/app/blob/z-targeting/player-avatar-binding.js), so maybe we need a `PlayerNetworkBinding` that performs the sync functions?

If you need to handle e.g. voice or avatar swaps, `Player`'s events can be used to listen and push that on the network. Then this file wouldn't have to worry about the network at all.

character-controller.js
```diff
@@ -456,130 +550,88 @@ class PlayerBase extends THREE.Object3D {
...
+  bindState(nextPlayersArray) {
+    // latch old state
+    const oldState = this.detachState();
+
+    // unbind
+    this.unbindState();
+    this.appManager.unbindState();
+
+    this.playersArray = nextPlayersArray;
+    if (!this.playersArray) return console.warn("this.playersArray is null")
```
avaer, line 593: Probably a case that needs to be handled, depending on how the network state API is abstracted.

character-controller.js
```diff
@@ -690,34 +733,34 @@ class StatePlayer extends PlayerBase {
             if (!cancelFn.isLive()) return;
             _setNextAvatarApp(nextAvatarApp);
           } else {
-            console.warn('switching avatar to instanceId that does not exist in any app manager', instanceId);
+            _setNextAvatarApp(null)
```
avaer, line 736: I think it's a legit state error if you try to switch to a nonexistent avatar and we shouldn't ignore that.

If the user wanted to set a null avatar they could do so explicitly.

character-controller.js
```diff
@@ -690,34 +733,34 @@ class StatePlayer extends PlayerBase {
...
       this.characterPhysics.setPosition(position);
     }
   }
   getActions() {
     return this.getActionsState();
   }
   getActionsState() {
-    let actionsArray = this.playerMap.has(avatarMapName) ? this.playerMap.get(actionsMapName, Z.Array) : null;
+    if (!this.playerMap) return []
+    let actionsArray = this.playerMap.has(avatarMapName)
```
avaer, line 761: This should go through a network abstraction so that this code is not aware of Z.js.

avaer: It could be a combination of `PlayerNetworkBinding` and `this.player.state = new PlayerState()` which can handle the implementation of things like actions.

avaer: The point is to make `Player` less prone to race conditions since it is dealing in very low level asynchronous manipulation.

character-controller.js
```diff
@@ -982,113 +895,118 @@ class UninterpolatedPlayer extends StatePlayer {
...
+      };
+      const pack4 = (v, i) => {
+        packedTransform[i] = v.x;
+        packedTransform[i + 1] = v.y;
+        packedTransform[i + 2] = v.z;
+        packedTransform[i + 3] = v.w;
+      };
+      pack3(self.position, 0);
+      pack4(self.quaternion, 3);
+      packedTransform[7] = 0;
```
avaer, line 999: What's this extra?

character-controller.js
```diff
@@ -1168,94 +1088,275 @@ class LocalPlayer extends UninterpolatedPlayer {
...
-    this.playersArray.doc.transact(() => {
-      /* if (isNaN(this.position.x) || isNaN(this.position.y) || isNaN(this.position.z)) {
-        debugger;
-      } */
-      this.playerMap.set('position', this.position.toArray(localArray3));
-      this.playerMap.set('quaternion', this.quaternion.toArray(localArray4));
-    }, 'push');
-
-    this.appManager.updatePhysics();
+  packed = new Float32Array(8);
```
avaer, line 1091: We are initializing in `this` already, so initializing more here is inconsistent. Can we put all properties in the same place?

avatars/avatars.js
avaer, line 2072: In future we will want to break out the entire audio processing stack into a worker, but that also depends on breaking out the networking stack so that audio can route directly to output even if the main thread is locked up.

No need to do that in this PR but that is a good thing to be thinking about.

avaer: Workers can talk to each other directly via https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel

character-controller.js
```diff
@@ -1168,94 +1088,275 @@ class LocalPlayer extends UninterpolatedPlayer {
...
-  detachState() {
-    return null;
+
+  async initRemotePlayer(){
+      // Called on the remote player on construction
+    for (let i = 0; i < this.appManager.appsArray.length; i++) {
+      const trackedApp = this.appManager.appsArray.get(i, Z.Map);
+      const app = await this.appManager.importTrackedApp(trackedApp);
+
+      if(!app) return console.error("App not found", app);
```
avaer, line 1276: These cases should be handled, since they seem like actual data races.

character-controller.js
```diff
@@ -1266,92 +1367,99 @@ class RemotePlayer extends InterpolatedPlayer {
...
-      action,
-    });
-  }
-  removeAction(type) {
-    for (let i = 0; i < this.actions.length; i++) {
-      const action = this.actions[i];
-      if (action.type === type) {
-        this.removeActionIndex(i);
-        break;
+    this.appManager.callBackFn = (app, event, flag) => {
```
avaer, line 1443: If you wanted events you could have a real `EventTarget` but even that seems wrong. I don't know what this is supposed to do but it can be written in a better way.

app-manager.js
```diff
@@ -487,195 +492,223 @@ class AppManager extends EventTarget {
...
-    
+
     let dstTrackedApp = null;
+    const srcTrackedApp = this.getOrCreateTrackedApp(instanceId);
+
+    const contentId = srcTrackedApp.get("contentId");
+    let transform = srcTrackedApp.get("transform");
+    const components = srcTrackedApp.get("components");
+
+    if (!transform) transform = new Float32Array(11);
```
avaer, line 551: Isn't sizeof(PQS) === 10?

app-manager.js
```diff
@@ -487,195 +492,223 @@ class AppManager extends EventTarget {
...
+
+    const pack3 = (v, i) => {
```
avaer, line 553: `THREE.Vector3` has `.toArray(arr, offset)` so these methods are pointless.

app-manager.js
```diff
@@ -487,195 +492,223 @@ class AppManager extends EventTarget {
...
-          const trackedApp = this.getTrackedApp(app.instanceId);
-          if (trackedApp) {
-            app.matrixWorld.decompose(localVector, localQuaternion, localVector2);
-
-            trackedApp.set('position', localVector.toArray());
-            trackedApp.set('quaternion', localQuaternion.toArray());
-            trackedApp.set('scale', localVector2.toArray());
-
-            app.updateMatrixWorld();
+  packed = new Float32Array(11);
```
avaer, line 620: `10`?

character-hups.js
```diff
@@ -153,6 +156,7 @@ class CharacterHups extends EventTarget {
     });
     player.addEventListener('actionremove', e => {
       const {action} = e;
+      if (!action) return console.warn("action remove didn't have an actionId", e)
```
avaer, line 159: Should throw.

character-sfx.js
```diff
@@ -94,7 +94,17 @@ class CharacterSfx {
     if (!this.player.avatar) {
       return;
     }
-    
+
+    const actions = this.player.getActionsState();
+
+    //check current actions has sit action
+    let hasSitAction = false
```
avaer, line 101: Could be `.hasAction`

game.js
```diff
@@ -558,6 +723,7 @@ const _gameUpdate = (timestamp, timeDiff) => {
     if (pickUpAction) {
       const {instanceId} = pickUpAction;
       const app = metaversefileApi.getAppByInstanceId(instanceId);
+      if(!app){ return console.error ("App not found when picking up`")}
```
avaer, line 726: Should throw.

game.js
```diff
@@ -607,14 +773,37 @@ const _gameUpdate = (timestamp, timeDiff) => {
...
+
       const collision = physx.physxWorker.raycastPhysics(physx.physics, position, quaternion);
       if (collision) {
         const physicsId = collision.objectId;
-        highlightedPhysicsObject = metaversefileApi.getAppByPhysicsId(physicsId);
-        highlightedPhysicsId = physicsId;
+        highlightedPhysicsObject =
+          metaversefileApi.getAppByPhysicsId(physicsId);
+        
+        if (!highlightedPhysicsObject) return
```
avaer, line 785: Instead of early return, can we nest?

metaverse_components/pet.js
```diff
@@ -57,17 +58,23 @@ export default (app, component) => {
...
       idleAction = m.idleAction;
+    } else {
+      console.warn('unhandled wear on pet');
     }
+    localPlayer = null;
   };
+
   app.addEventListener('wearupdate', e => {
     if (e.wear) {
+      localPlayer = metaversefile.getPlayerByInstanceId(app.instanceId);
```
avaer, line 73: I would not call this `localPlayer` since the player might not be local.

util.js
```diff
@@ -943,6 +943,16 @@ export const handleDropJsonItem = async (item) => {
...
   return null;
 };
+
+export const handleBlobUpload = async (name, blob, progress) => {
+  const formData = new FormData();
+  formData.append(name, blob, name);
+  const hashes = await doUpload(`https://ipfs.webaverse.com/`, formData, progress);
+
+  const rootDirectoryHash = hashes.length > 0 ? hashes[0].hash : null;
+  return `https://ipfs.webaverse.com/ipfs/${rootDirectoryHash}/`;
```
avaer, line 953: These should be in `constants.js`. I also don't think it's related to this PR.

world.js
```diff
@@ -288,44 +25,140 @@ const _getBindSceneForRenderPriority = renderPriority => {
...
+    // Create a new instance of the websocket rtc client
+    // This comes from webaverse/wsrtc/wsrtc.js
+    this.wsrtc = new WSRTC(u, {
+      localPlayer,
+      crdtState: state,
+    });
+
+    // Handy debug function to see the state
+    // Delete this eventually. For now press escape to see world state
+    document.addEventListener('keydown', event => {
```
avaer, line 116: We should centralize these hacks.

There is already a place for it: https://github.com/webaverse/app/blob/6ab3bb7d2da38454e71c92105e6c7aad4b6ea13c/webaverse.js#L545

avaer (changes requested):
The main thing I recommend is a network abstraction of `Player`. Doing that will fix bugs now and in the future.

alisaad673:
Non Multiplayer Issues:

1-Something is wrong with the flying animation. It has this weird jitter in it.

https://user-images.githubusercontent.com/35926530/176144472-39faad03-5e92-49b8-90fd-cb93ffcfda10.mp4

2-Unable to switch any of the scenes. It throws this error

image

3-Avatar Pfp isn't loading. I guess this error is related to that.

image

4-Wrong jump animation

https://user-images.githubusercontent.com/35926530/176146093-4b10be64-3473-4262-b2bf-a73e253a0ea3.mp4

5-The equip animation is broken too. The objects get equipped by just pressing 'E'.)we don't need to hold E.

https://user-images.githubusercontent.com/35926530/176147246-a5efd4ef-d659-4030-a67b-31d114e0c820.mp4

6-While we try to uneqip items by pressing R. It first throws this error and on the second attempt the item gets unequipped.

image

7-Fox isn't working( Following the player) it keeps moving its feet in its original position

https://user-images.githubusercontent.com/35926530/176147702-57b4d3b6-2806-4b9f-b2fa-c06e1280252e.mp4

8-No sword animation

https://user-images.githubusercontent.com/35926530/176147920-be2c0bee-dfb4-423e-abf1-fa414643d606.mp4

9-Once we equip sword, another sword appears in its place. And once we unequip the sword, it leavers an instance of that sword that isn't interactable anymore

https://user-images.githubusercontent.com/35926530/176148275-b32c396a-f21c-41b0-86c5-88dbaebd266a.mp4

10-With vehicles(hovercraft and dragon too) there are certain issues:
           -Once we equip it, another instance of vehicle instantly appears at the original position. This vehicle is not equip able.
           -When we press "spacebar" to get out of vehicle, It first throws an error ,than avatar rides the vehicle standing.
           -Once we get out of vehicle by pressing "R" twice, when we try to ride it again. Everything breaks.

https://user-images.githubusercontent.com/35926530/176149765-0ad12893-17ef-447d-90fb-32ff399b7662.mp4

https://user-images.githubusercontent.com/35926530/176153950-948681d0-1dad-4a5c-86d9-d3bbf3f5e149.mp4

11-Character Select isn't working for any of the characters. No previews, Unable to equip characters. Throws this error.

image

https://user-images.githubusercontent.com/35926530/176150736-1f326015-e330-4c43-aaa4-8d35f61ae4d1.mp4

12-WHile we try to equip characters from character select,  We are unable to equip them, but they appear standing in the scene. But we aren't able to switch. Throws this error

image

https://user-images.githubusercontent.com/35926530/176152127-b325d1d1-3189-41dc-a8d4-2c78dfd93fd3.mp4

13-Trying to equip items from character tab doesn't works. No previews, unable to equip. Thows these errors.

image

https://user-images.githubusercontent.com/35926530/176151253-54e0b5ae-c508-4a7c-9c15-3cb88775b98d.mp4

14-Trying to use bow, it crashes the app. Throws this error

image

15- We have issues with potion drinking animation, Issues with aiming guns/pistol. 

Multiplayer:

1-While I try to join multiplayer by creating a room, It throws this error. And avatar is unable to move. Nothing happens

image


I wasn't able to test the multiplayer rooms as i wasn't able to get into that.

PR #3176, opened 2022-06-28
NFT Blockchain

mwmwmw:


lalalune:
@mwmwmw mostly looks good, seems like world tokens won't load without opensea api, please review this:

https://user-images.githubusercontent.com/18633264/177024112-7c179792-bf8d-4714-b996-c1f85a2f02a5.mp4

lalalune:
@supernftier @mwmwmw please review

I deleted OpenSea API key, so I can now see the old user inventory window (white window on the right side)

image

image

I have some NFTs which you can see. However they don't show up in my player inventory (tab)

image

Lastly, there is no way to log out or change wallets. The "logout" button in settings doesn't worry, and it feels like there should be an option near the login.
image

Here is a video clicking through all of it.

https://user-images.githubusercontent.com/18633264/177024503-7410176b-fd48-48e1-8d4f-e855923a1133.mp4

I think we need to rid of the wallet on the right side entirely, and move inventory into the player inventory (tab menu)

lalalune:
Closing in favor of https://github.com/webaverse/app/pull/3203

PR #3187, opened 2022-07-03
ConvAI Emotion Responses

lalalune:
This simplifies the ConvAI emotion response changes, both because the PR got trashed by some weird merges and because automatic linting added a ton of confusing stuff.

This replaces https://github.com/webaverse/app/pull/3042

Will confirm everything is as-it-should-be and we have implemented response additions for all conversation

### DIFF
--- a/ai/lore/lore-ai.js
+++ b/ai/lore/lore-ai.js
@@ -38,9 +38,6 @@ const numGenerateTries = 5;
 const temperature = 1;
 const top_p = 1;
 
-// Initializing the action generation URL from ConvAI
-const actionGenerationURl = `https://test.convai.com/getActions`
-
 class AICharacter extends EventTarget {
   constructor({
     name = defaultPlayerName,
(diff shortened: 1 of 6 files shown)


alisaad673:
Hey @lalalune.  This is throwing a ton of issues. It was working before.

1-In the automated response system(double press E and then left click): when there is no option to go ahead( means only our character speaks and chat ends) when we left click on the prompt)
It throws this error:

image

https://user-images.githubusercontent.com/35926530/177823867-cbee31ba-b87b-4ece-a540-4cc9a95c5bf3.mp4



2-In the automated chat, when there is an option to parse to next response, It throws this error and app is stuck now because we are unable to move forward or end the prompt.

image

https://user-images.githubusercontent.com/35926530/177823788-91d88c0b-17b3-4709-8773-7dd624453526.mp4





3-In the general chat, the responses are there in console but they are not appearing as voice and text prompts like it does for our character. Also it throws this error each time we are to get a response from the npc. Note that the error is being thrown only when the emote component is anything other than 'none'

image

https://user-images.githubusercontent.com/35926530/177823850-62244f8f-f2e9-4ea3-bcbb-597d6035b9eb.mp4

lalalune:
Closed in favor of https://github.com/webaverse/app/pull/3206

PR #3195, opened 2022-07-06
Don't play footsteps while character is sitting

lalalune:
This PR adds a check to see if the character is sitting before playing a footstep sound. This fixes footsteps playing while the player is on a mount or vehicle.


alisaad673:
Tested it on multiple vehicles. Works fine.

PR #3196, opened 2022-07-06
Sonic boom players for all players, not just localplayer

lalalune:
This fixes a bug where all players are creating a sonic boom app for local player when they join the world. Instead, the player is referenced into the sonic boom app, the result is sonic boom works in multiplayer.


alisaad673:
Sonic boom is already working for local player though, but  I am unable to test it in multiplayer as multiplayer rooms aren't working for me at all.

image

lalalune:
@alisaad673 this was extracted from multiplayer PR, so while it definitely works, it is dependent on multiplayer to QA -- we can wait until those PRs are merged in

character-fx.js
```diff
@@ -234,8 +234,9 @@ class CharacterFx {
 
     this.lastSSS = isSSS;
     const _updateSonicBoomMesh = () => {
-      if (this.player.isLocalPlayer && !this.sonicBoom) {
+      if ( !this.sonicBoom  && !this.player.isNpcPlayer ) {
         this.sonicBoom = metaversefile.createApp();
+        this.sonicBoom.player = this.player;
```
lalalune, line 239: make player id a component, if its null player then is local player, otherwise look itup

metaverse_modules/sonic-boom/index.js
```diff
@@ -7,7 +7,7 @@ const baseUrl = import.meta.url.replace(/(\/)[^\/\\]*$/, '$1');
 
 export default () => {
   const app = useApp();
-  const localPlayer = useLocalPlayer();
+  const localPlayer = app.player ?? useLocalPlayer();
```
avaer, line 10: If the player is being passed in, it should be a component (`playerId`). If it's `null` it can mean local, otherwise we can look it up as a remote player.

In general the interface to programming apps should be components only.

avaer (changes requested):
Suggest cleaning up the configuration of the app to go through `components`.

PR #3201, opened 2022-07-07
website has a black background at start

lalalune:
The background color of the app is currently default, which leads to a really awkward white screen while UI loads over it. This PR makes it black, which vibes nicer with the coming app loading screen etc.

### DIFF
--- a/index.css
+++ b/index.css
@@ -3,6 +3,7 @@
 }
 body {
   margin: 0;
+  background-color: #000000;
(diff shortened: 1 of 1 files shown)


lalalune:
Wasn't able to reproduce on my chrome. I think this is probably a fine edge case, it's a small aesthetic detail and one line of css. The other options around  this seem to be painful.

PR #3203, opened 2022-07-07
Z nft blockchain

lalalune:
This adds the work from supernftier and Gravecat into a new PR, removing any other dependant cruft.

This was reviewed a bit before -- I removed the sidebar and changed the styling on the account view a bit, and inventory is loaded form opensea in the user inventory window instead of the tokens.jsx sidebar


lalalune:
@supernftier can you follow up with this
1. I moved tokens from the menu on the right side to the user inventory, need to make sure that didn't break existing inventory :o: 
2. Need to make sure all chains are supported in inventory, and test with mainnet Polygon
3. Respond to any other notes from @alisaad673

lalalune:
@supernftier i cherry-picked your changes onto this branch, please review

supernftier:
About major issues from alisaad673
1. Fixed Inventory Issue - It is caused by an old smart contract on the Mumbai testnet, so we have deployed the latest version of the smart contract on the testnet.
2.  Untill now, there isn't log out button.
3. Fixed an issue with notifications issued successfully by nft. - Currently, a notification will be displayed until the NFT issuance is completed on the chain.
4. Inventory mint is working well in polygon mainnet.

alisaad673:
This is working good now. Major issues are solved. Minting is working fine on polygon mainnet and mumbai testnet.

just some minor issues and questions:
1-The notification still appears before the nft is actually minted(and appears in the inventory)

crop

2-There are these unrelated chains that we are not using. I remember grave said they only appear on local. Is that the case? Can we remove them if they are not needed?

image

3-The webaverse sidechain and etherium chain are not supported yet right?

4-What does the claims button do in inventory? When I click on it, it just throws a warning "failed minting on webaverse contract"

https://user-images.githubusercontent.com/35926530/179065744-cbea6e0e-d691-42c5-8193-87adc204734a.mp4

alisaad673:
The above issues are all fixed. Its working fine now. But, as the issue of nft minted successfully notification has been fixed, this "upload object" popup that appears after we drag and drop any asset stays until the notification of nft minted successfully appears. I guess its attached to that notification and only goes away after the notification. The issue is during the time in which nft is being minted, if the user clicks again on mint on polygon button, it can start another transaction, thus disturbing the flow. Can we disable the mint button once a transaction has started or if we can make this "upload object" popup go away right after user approves the transaction it would be great.
Also there's a conflict too. Can you resolve that.

image

https://user-images.githubusercontent.com/35926530/179220328-e8192560-4c48-495b-bd9f-d8069141f7b4.mp4

supernftier:
I fixed all issues
- Notifications appear and disappear on time.
- Disable mint while a transaction is pending.

alisaad673:
This is working Fine now.

metaverse-modules.js
```diff
@@ -21,6 +21,7 @@ const moduleUrls = {
   cameraPlaceholder: './metaverse_modules/camera-placeholder/',
   targetReticle: './metaverse_modules/target-reticle/',
   halo: './metaverse_modules/halo/',
+  silsword: './metaverse_modules/silsword/',
```
lalalune, line 24: we should reference the silsword at https://github.com/webaverse/silsword

drop-manager.js
```diff
@@ -55,6 +67,7 @@ class DropManager extends EventTarget {
       level,
       voucher,
     };
+    console.log("voucher", voucher)
```
lalalune, line 70: remove love

drop-manager.js
```diff
@@ -64,10 +77,45 @@ class DropManager extends EventTarget {
     }));
   }
   pickupApp(app) {
+    console.log("app", app)
```
lalalune, line 80: remove log

src/components/general/equipment/Equipment.jsx
```diff
@@ -318,9 +328,56 @@ export const Equipment = () => {
...
+            const inventoryItems = tokens.map((token, i) => {
+                return {
+                    name: token.name ?? "",
+                    start_url: token.url ?? "",
+                    // start_url: token.url ?? token.animation_url ?? token.collection.banner_image_url ?? "",
+                    level: token.level ?? 1,
+                    claimed: true
+                };
+            });
+            console.log("invnetory", inventoryItems)
```
lalalune, line 367: remove log

src/components/general/equipment/Equipment.jsx
```diff
@@ -557,10 +622,17 @@ export const Equipment = () => {
...
+                selectedMenuIndex={selectedMenuIndex}
                 name={selectObject ? selectObject.name : null}
+                selectObject={selectObject ? selectObject : null}
                 description={selectObject ? selectObject.description : null}
                 imageBitmap={imageBitmap}
                 onActivate={onDoubleClick(selectObject)}
+                mintEnabled={isChainSupported(selectedChain) && account.currentAddress}
+                onMint={() => {
+                    mintClaim(selectObject);
+                    console.log("mint object", selectObject)
```
lalalune, line 634: remove log

src/hooks/voucherHelpers.js
```diff
@@ -0,0 +1,22 @@
...
+export async function getVoucherFromServer() {
+    const tokenId = 1;
+
+    const expiry = Math.round(new Date().getTime() / 1000) + 1000;//timestamp
+    // const expiry = 1657335995;//timestamp
+    const nonce = ethers.BigNumber.from(ethers.utils.randomBytes(4)).toNumber();
+    // const nonce = 552311376;
+    const balance = 0;
+
+    const response = await fetch("http://172.32.13.15:8081/claim", {
```
lalalune, line 12: this needs to be broken out into a const, placed in constants.js, we probably need something like voucher.webaverse.com -- then we can do local loopback to test (add entry to /etc/hosts)

lalalune:
TODO: This should proxy through the API server to the claim server, to hide the claim server from the world.

src/components/web3/chains.jsx
```diff
@@ -0,0 +1,47 @@
+import React, {useContext, useState} from 'react';
+import {ChainContext} from '../../hooks/chainProvider';
+import {isChainSupported} from '../../hooks/useChain';
+import {CHAIN_TYPE} from '../../hooks/web3-constants';
+
+import styles from './chains.module.css';
+
+function UpArrow() {
+  return <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" viewBox="0 0 330 330">
```
avaer, line 9: We shouldn't inline assets like this... it should be a separate file.

avaer:
This got reverted due to intractable regressions.

Issue #3129, opened 2022-06-16
Z-Targeting: Crash while trying to Switch Scene from street scene 


msm0823:
Selecting the same scene causes the graphics to turn black and cause errors.

https://user-images.githubusercontent.com/104460803/174085263-48271aab-28bd-43e8-bc3f-cc514b7a33a1.mp4

https://user-images.githubusercontent.com/104460803/174085290-153e4b34-ccbb-4722-9e7b-a6ab44fc9587.mp4

alisaad673:
This is actually the case with street scene, If we try to switch to any other scene from street scene, It crashes the app and throws these long list of errors.

![switching zt](https://user-images.githubusercontent.com/35926530/174105371-9687d464-6b1b-4630-897e-e0790f852a83.PNG)
![switching zt 2](https://user-images.githubusercontent.com/35926530/174105364-10fc90c8-a7d1-4a9d-9f75-0d706dcff684.PNG)

alisaad673:
This has been due to the stats banner, since the stats banner is removed, it no longer happens.
https://github.com/webaverse/app/issues/3267

PR #3207, opened 2022-07-08
Z: Move multiplayer room handling to universe.js and make world a proper class

lalalune:
Multiplayer connection and disconnection was handled awkward between universe.js and world.js -- we don't want the multiplayer system to necessary be dependent on the lifecycle of the world object. This moves that functionality to universe.js, which also simplifies imports and calls elsewhere. Also made World a proper class object to be more inline with code elsewhere.

The player update will throw errors when joining or leaving the world, because this is going to be handled in a separate PR. The success for this PR is that you can join or leave a multiplayer room at all, since the functions we are moving are isConnected and getConnection checks.


alisaad673:
Entering and leaving the worlds works fine. One thing is as we have errors switching scenes from street scene, It works fine when checked from other scenes.
https://github.com/webaverse/app/issues/3129

lalalune:
@alisaad673 this is a separate bug we'll need to address, I believe this is happening on z-targeting and was pointed out by msm (but having a hard time finding a ticket right now)

Issue #3150, opened 2022-06-21
Z-Targeting: Screen turns white on equiping dragon


msm0823:
After equipping dragon from character menu, the screen turns white. However, it becomes normal after refreshing it.

https://user-images.githubusercontent.com/104460803/174780375-4b08137c-537b-4faf-a6af-18ffef920348.mp4

Issue #3136, opened 2022-06-17
Z-Targeting: Dragon causing errors


msm0823:
Equipping Dragon is causing errors

https://user-images.githubusercontent.com/104460803/174288737-eb641d3d-1eb6-4fe0-88f5-2a3ed3c7f8e6.mp4

https://user-images.githubusercontent.com/104460803/174288691-382522c3-838a-4377-aeb4-7b1e1b2b2b27.mp4

PR #3211, opened 2022-07-08
patch fix riding on mounts like the dragon

lalalune:
This vehicle code is going to be completely updated, but for now it fixes totally broken mounts

addresses these:
https://github.com/webaverse/app/issues/3136
https://github.com/webaverse/app/issues/3150

image


alisaad673:
Works Fine. Just has a conflict.

PR #3218, opened 2022-07-10
Fix remote player use [E]

lalalune:
Right now, when the player goes to use an item an error is thrown on the last frame of the use. This is because there is a race condition with the use action itself being removed from the network state. The easy fix for this is to null check the action before checking it's animation name. This only handles this edge case and otherwise the code is ignored.

This depends on several multiplayer dependencies, but can be tested to not break anything on single player at least.

### DIFF
--- a/avatars/animationHelpers.js
+++ b/avatars/animationHelpers.js
@@ -1233,7 +1233,9 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
       let defaultAnimation = 'grab_forward';
 
       const activateAction = localPlayer.getAction('activate');
-      if (activateAction.animationName) {
+      // activateAction can be null on remote player on the last frame the action is removed
+      // so we null check it before checkout for the animation name
+      if (activateAction && activateAction.animationName) {
         defaultAnimation = activateAction.animationName;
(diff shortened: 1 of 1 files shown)


avatars/animationHelpers.js
```diff
@@ -1233,7 +1233,9 @@ export const _applyAnimation = (avatar, now, moveFactors) => {
       let defaultAnimation = 'grab_forward';
 
       const activateAction = localPlayer.getAction('activate');
-      if (activateAction.animationName) {
+      // activateAction can be null on remote player on the last frame the action is removed
+      // so we null check it before checkout for the animation name
+      if (activateAction && activateAction.animationName) {
```
lalalune, line 1238: can we do this with an if statement on activate action above?

avaer, line 1238: Can this check be moved up to the `if` above?

```
const activateAction = localPlayer.getAction('activate');
    if (activateAction) {
      // const localPlayer = metaversefile.useLocalPlayer();
      ```

avaer (changes requested):
Suggestion.

PR #3219, opened 2022-07-10
Fixing wearing and unwearing pets in multiplayer (and bugs in single player)

lalalune:
In the current implementation, pets can be offset from their root if they move. When this is fixed, the pet jumps to where the player placed them. This PR fixes that and abstracts the local player references so pets work in multplayer.

To test: wear and unwear fox in the shadows scene. To test multiplayer we'll need to have some other PRs pulled in, however since the pet now treats all players as equal this is effectively guaranteed.

### DIFF
--- a/metaverse_components/pet.js
+++ b/metaverse_components/pet.js
@@ -18,7 +18,9 @@ export default (app, component) => {
   let walkAction = null;
   let runAction = null;
   let rootBone = null;
-  
+
+  let player = null;
+
   const petComponent = component;
   const _makePetMixer = () => {
     let petMixer, idleAction;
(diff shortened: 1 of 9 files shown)


metaverse_components/pet.js
```diff
@@ -108,15 +111,14 @@ export default (app, component) => {
           rootBone.updateMatrixWorld();
         }
         if (petMixer) { // animated pet
-          if (petSpec) { // activated pet
+          if (petSpec && player) { // activated pet
```
lalalune, line 114: is player ever null?

avaer, line 114: Should remove these checks if `player` cannot be null.

That would mean that the worn app is not part of the player's app manager, which seems it would be a bug.

avaer (changes requested):
Suggest removing guard checks.

lalalune:
Resolved comments

PR #3225, opened 2022-07-10
Add nameplates above players heads in multiplayer

lalalune:
This adds basic instanced player nameplates, along with a new nameplate metaverse_module.

May require a bit more design before final release, but looks good in screenshots and improves the experience.

Requires multiplayer to test


metaverse_modules/nameplate/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+const {useApp, useCamera, useLocalPlayer, useMaterials, useFrame, useText} =
+  metaversefile;
+
+const Text = useText();
+const gltfLoader = new GLTFLoader();
+let nameplateMesh = null;
+
+async function createNameplateMesh() {
+  if (nameplateMesh) return;
```
avaer, line 15: There are a lot of guard conditions like this. Can we program this without all of these checks and just call things in the correct order?

metaverse_modules/nameplate/index.js
```diff
@@ -0,0 +1,125 @@
...
+  const nameplateModel = await new Promise((resolve, reject) => {
+    gltfLoader.load(nameplateModelUrl, resolve, () => {}, reject);
+  });
+  const {geometry, material} = getModelGeoMat(nameplateModel);
+  nameplateMesh = new THREE.InstancedMesh(geometry, material, 1000);
+}
+
+const createNameplateInstance = () => {
+  if (!nameplateMesh) return 0;
+  if (!nameplateMesh.instanceIndex) nameplateMesh.instanceIndex = 0;
```
avaer, line 26: I think this variable is called `.count`?

https://threejs.org/docs/#api/en/objects/InstancedMesh.count

We do we need to declare it again?

metaverse_modules/nameplate/model.js
```diff
@@ -0,0 +1,64 @@
...
+          materialIndices.push(newItems[1]);
+        }
+      } else {
+        newItems = addToMaterials(newMaterials, child.material); // materials.push(child.material);
+        newMaterials = newItems[0];
+        materialIndices.push(newItems[1]);
+      }
+
+      if (child.geometry.isBufferGeometry) {
+        const tGeometry = new Geometry().fromBufferGeometry(child.geometry);
```
avaer, line 35: Why use a non-buffer geometry?

metaverse_modules/nameplate/model.js
```diff
@@ -0,0 +1,64 @@
...
+    if (mat.name === newMaterial.name) {
+      return [materials, i];
+    }
+  }
+  materials.push(newMaterial);
+  return [materials, materials.length - 1];
+};
+
+export const getModelGeoMat = model => {
+  const newGeometry = new Geometry();
```
avaer, line 16: The main point of instancing is to use a single draw call. But here we are using multiple materials anyway.

Can it be a single material model and we can get rid of this loader hack?

metaverse_modules/nameplate/model.js
```diff
@@ -0,0 +1,64 @@
...
+          newMaterials = newItems[0];
+          materialIndices.push(newItems[1]);
+        }
+      } else {
+        newItems = addToMaterials(newMaterials, child.material); // materials.push(child.material);
+        newMaterials = newItems[0];
+        materialIndices.push(newItems[1]);
+      }
+
+      if (child.geometry.isBufferGeometry) {
```
avaer, line 34: Why can't this operate on BufferGeometry? Regardless, I don't think we need this code at all (see above).

avaer (changes requested):
The new app needs cleanup.

metaverse_modules/nameplate/model.js
ultrabot2022: In case there are several children meshes in the glb model, I had to generate single geometry and matrial to implement MultiMeshInstance by merging them. So at first I tried to merge BufferGeometries into single BufferGeometry keeping children's position, but after merging their position was moved to zero point (0, 0, 0). As a result, the imported glb file wasn't similar to the original glb file and there was no way to keep their position when using BufferGeometry.
So totally, I decided to use legacy geometry to control faces and matrix to help keeping children's position. I'd like to get any hint how to keep position when using BufferGeometry. Thanks.

metaverse_modules/nameplate/model.js
ultrabot2022: Merging materials is only possible in very specific situations. The materials basically need to be identical, except for the 'color'.
So the correct single material can't be got when merging the materials, I think.

metaverse_modules/nameplate/model.js
ultrabot2022: I had to use legacy geometry to keep children's position when merging several meshes.

metaverse_modules/nameplate/index.js
ultrabot2022: I used such conditions to implement the correct MultiMeshInstance function. And separating the features is good for programming, I think. Please let me know if any other solutions.

metaverse_modules/nameplate/model.js
lalalune: @ultrabot2022 if the model has more than one material we should remove that. Basically, we should assume that everyone has the same nameplate with a single mesh and single material

metaverse_modules/nameplate/model.js
lalalune: The nameplate mesh should have a single model

metaverse_modules/nameplate/index.js
lalalune: Lets assume single mesh for this case, we will require all future nameplates to be single mesh, single material

PR #3226, opened 2022-07-10
Fix error where Remote Player tries to sync a null avatar when initialized

lalalune:
This fixes an error where the remote player tries to sync before their avatar has been created. This PR removes the syncAvatar from the constructor of the remove player, and adds a throw to check if this behavior happens. The remote player syncs their avatar from the avatar variable anyways.

Without this PR, an error will throw on the remote player. With this PR, there are no longer any errors.

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -1281,6 +1281,8 @@ class RemotePlayer extends InterpolatedPlayer {
     
     this.appManager.bindState(this.getAppsState());
     this.appManager.loadApps();
+    
+    this.syncAvatar();
   }
 }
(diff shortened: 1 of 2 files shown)


character-controller.js
```diff
@@ -696,8 +696,6 @@ class StatePlayer extends PlayerBase {
           }
         }
       }
-    } else {
```
avaer, line 699: IMO avatar syncing should not be attempted if there is no avatar instance id to switch to.

What about else `throw`? This seems like a legitimate bug if we need to do this, so it's best to call it out and at least get a stack trace.

PR #3228, opened 2022-07-10
Update scene selector to work with multiplayer and multiplayer rooms

lalalune:
Rebases the scene selector commits which were removed as part of a larger revert. Updated the scene selector UI so that the player can go to multiplayer, fixes the 'undefined' room bug and encourages user to create as room instead of default to Erithor.

### DIFF
--- a/src/components/editor-mode/scene-menu/SceneMenu.jsx
+++ b/src/components/editor-mode/scene-menu/SceneMenu.jsx
@@ -32,10 +32,7 @@ export const SceneMenu = ({ multiplayerConnected, selectedScene, setSelectedScen
     const [ speechEnabled, setSpeechEnabled ] = useState( false );
     const [ sceneInputName, setSceneInputName ] = useState( selectedScene );
     const [ scenesList, setScenesList ] = useState( origSceneList );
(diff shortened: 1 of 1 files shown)


src/components/editor-mode/scene-menu/SceneMenu.jsx
```diff
@@ -96,32 +96,27 @@ export const SceneMenu = ({ multiplayerConnected, selectedScene, setSelectedScen
...
-        //     const text = await res.text();
-        //     console.warn( 'error creating room', res.status, text );
-
-        // }
-
-    };
+        const sceneName = selectedScene.trim();
+        const data = null; // Z.encodeStateAsUpdate( world.getState( true ) );
+
+        const roomName = (Math.random().toString(36) + '00000000000000000').slice(2, 10);
```
avaer, line 102: Should use https://github.com/webaverse/app/blob/e298191a0cb0765e0fbdc563f1f7250b88323623/util.js#L584

avaer (changes requested):
Just one edit.

lalalune:
This is fixed

PR #3231, opened 2022-07-13
delete addapp

supernftier:


avaer:
Needs more research. I suspect this might splice the app across app managers.

alisaad673:
The duplicate items issue seems fixed, but app crashes while we click on any of the the world objects:

image
image

PR #3239, opened 2022-07-14
Separate avatar velocity from avatar update loop

lalalune:
Right now we are updating the avatar velocity every frame.

The problem is that if we don't receive updated velocity information this frame we can get hiccups. The solution I have found is to update the velocity on remote avatars when the observer updates on transform calls, since any change in motion would necessitate a change in velocity. Also, this velocity is used to update the appearance of the avatar, nothing to do with position.

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -1490,7 +1490,7 @@ class Avatar {
     localEuler.setFromQuaternion(currentQuaternion, 'YXZ');
     localEuler.set(0, -(localEuler.y + Math.PI), 0);
     positionDiff.applyEuler(localEuler);
-    this.velocity.copy(positionDiff).add(this.lastVelocity).divideScalar(2);
+    this.velocity.copy(positionDiff);
     this.lastVelocity.copy(this.velocity);
     this.direction.copy(positionDiff).normalize();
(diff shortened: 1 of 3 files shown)


avatars/avatars.js
```diff
@@ -1475,6 +1481,25 @@ class Avatar {
...
+  setVelocity(timeDiffS, lastPosition, currentPosition, currentQuaternion) {
+    // Set the velocity, which will be considered by the animation controller
+    const positionDiff = localVector.copy(lastPosition)
+      .sub(currentPosition)
+      .divideScalar(Math.max(timeDiffS, 0.001))
+      .multiplyScalar(0.1);
+    localEuler.setFromQuaternion(currentQuaternion, 'YXZ');
+    localEuler.set(0, -(localEuler.y + Math.PI), 0);
+    positionDiff.applyEuler(localEuler);
+    this.velocity.copy(positionDiff).add(this.lastVelocity).divideScalar(2);
```
avaer: I don't understand how this velocity is computed.

It's the average of these two points?

lalalune:
Updated to set isLocalPlayer on the avatar at construction time

PR #3240, opened 2022-07-14
Add update function and call on remote player

lalalune:
Previously we had commented out the update function because it did not exist. This PR adds an update function and uncomments the existing call (while changing it from updateAvatar to update)


character-controller.js
```diff
@@ -939,6 +939,22 @@ class InterpolatedPlayer extends StatePlayer {
       quaternion: this.quaternionInterpolant.get(),
     };
   }
+  update(timestamp, timeDiff) {
+    this.updateInterpolation(timeDiff);
+
+    const mirrors = metaversefile.getMirrors();
+    if(!this.avatar) return console.warn("no avatar");
```
avaer, line 946: `throw`?

lalalune: i will test, if its legit then we will remove the warning

avaer (changes requested):
Small change.

PR #3241, opened 2022-07-14
Destroy remote players when players-manger is unbound

lalalune:
The adds a handler function to destroy any remote players who might be in the scene when the user goes to leave the room.

### DIFF
--- a/players-manager.js
+++ b/players-manager.js
@@ -32,7 +32,9 @@ class PlayersManager {
     return this.playersArray;
   }
   unbindState() {
+    if(this.unbindStateFn != null) {
       this.unbindStateFn();
+    }
       this.playersArray = null;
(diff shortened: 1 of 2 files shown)


lalalune (commented):
This should be called as deleteRemotePlayers, not inside unbindState

players-manager.js
```diff
@@ -20,6 +20,16 @@ class PlayersManager {
   unbindState() {
     const lastPlayers = this.playersArray;
     if (lastPlayers) {
+      const playerSpecs = lastPlayers.toJSON();
```
avaer, line 23: This should be in a method like `clearRemotePlayers`.

avaer (changes requested):
Small change.

PR #3251, opened 2022-07-14
Vertical swimming.

patriboz:
Implemented vertical swimming animations on keys SPACE and C.

### DIFF
--- a/ai/lore/lore-model.js
+++ b/ai/lore/lore-model.js
@@ -1,2 +1,2 @@
 // export * from '../../lore/lore-model.js';
-export * from 'http://127.0.0.1:8080/lore-model.js';
+export * from 'http://localhost/lore-model.js';
--- a/ai/lore/lore-model.js
+++ b/ai/lore/lore-model.js
@@ -1,2 +1,2 @@
 // export * from '../../lore/lore-model.js';
-export * from 'http://localhost/lore-model.js';
+export * from 'http://127.0.0.1:8080/lore-model.js';
--- a/avatars/animationHelpers.js
+++ b/avatars/animationHelpers.js
@@ -1253,8 +1253,6 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
       const f = MathUtils.clamp(swimTimeS / 0.2, 0, 1);
       // if (isPosition) console.log(f);
 
-      
-
       if (!isPosition) {
         localQuaternion2.fromArray(v2);
         localQuaternion3.fromArray(v3);
@@ -1262,13 +1260,14 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
         localQuaternion5.fromArray([0,0,0,1]);
         localQuaternion6.fromArray([1,0,0,0]);
         
-        
         // // can't use idleWalkFactor & walkRunFactor here, otherwise "Impulsive breaststroke swim animation" will turn into "freestyle animation" when speed is fast,
         // // and will turn into "floating" a little when speed is slow.
         // localQuaternion3.slerp(localQuaternion4, walkRunFactor);
         // localQuaternion2.slerp(localQuaternion3, idleWalkFactor);
-        localQuaternion3.slerp(localQuaternion4, avatar.sprintFactor);
-        localQuaternion2.slerp(localQuaternion3, avatar.movementsTransitionFactor);
+        if(!avatar.swimmingOnSurfaceState || (avatar.swimmingOnSurfaceState && avatar.horizontalMovementsTransitionFactor > 0)) {
+          localQuaternion3.slerp(localQuaternion4, avatar.sprintFactor);
+          localQuaternion2.slerp(localQuaternion3, avatar.movementsTransitionFactor);
+        }
 
         if(boneName === 'Hips') {
           if(avatar.swimUpFactor > 0 && avatar.horizontalMovementsTransitionFactor === 0) {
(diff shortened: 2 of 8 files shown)


ai/lore/lore-model.js
```diff
@@ -1,2 +1,2 @@
 // export * from '../../lore/lore-model.js';
-export * from 'http://localhost/lore-model.js';
\ No newline at end of file
+export * from 'http://127.0.0.1:8080/lore-model.js';
```
lalalune, line 2: please remove this from PR

avatars/animationHelpers.js
```diff
@@ -1225,9 +1228,10 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
       animationTrackName: k,
       dst,
       isPosition,
+      boneName,
       isFirstBone,
     } = spec;
-
+
```
lalalune, line 1234: This looks like VSCode auto formatting.

I know we have a bunch of red squigglies, but can we remove these / paste them back in for the sake of making a really simple PR and then we can do formatting separately?

avatars/animationHelpers.js
```diff
@@ -1249,17 +1253,49 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
...
+            localQuaternion2.slerp(localQuaternion6, fU);
+            localQuaternion2.slerp(localQuaternion7, fD);
+          }
+        }
+
+
         dst.slerp(localQuaternion2, f);
+        
+
+
```
lalalune, line 1298: no need for these spaces

scenes/water.scn
```diff
@@ -49,15 +46,8 @@
...
+      "scale": [100, 1, 100],
       "start_url": "/practice/water-main2/"
-    },
-    {
-      "position": [
-        0,
-        0,
-        0
-      ],
-      "start_url": "/practice/dual-contouring-terrain-main/"
```
lalalune, line 60: do we need these changes in pr?

avatars/animationHelpers.js
```diff
@@ -1249,17 +1253,49 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
...
         // // and will turn into "floating" a little when speed is slow.
         // localQuaternion3.slerp(localQuaternion4, walkRunFactor);
         // localQuaternion2.slerp(localQuaternion3, idleWalkFactor);
         localQuaternion3.slerp(localQuaternion4, avatar.sprintFactor);
         localQuaternion2.slerp(localQuaternion3, avatar.movementsTransitionFactor);
+
+        
+
+        // Swinging Hips for Breaststroke
+        if(boneName === 'Hips' && avatar.sprintFactor === 0 && movementsTimeS > 0) {
```
lalalune, line 1277: since we are checking if bonename === hips below, could we embed this boolean check inside the other one?

character-controller.js
```diff
@@ -977,16 +977,28 @@ class UninterpolatedPlayer extends StatePlayer {
       emote: new BiActionInterpolant(() => this.hasAction('emote'), 0, crouchMaxTime),
       movements: new InfiniteActionInterpolant(() => {
         const ioManager = metaversefile.useIoManager();
```
lalalune, line 979: We could probably move the useIoManager call up to the top of the function and then re-use it

character-controller.js
```diff
@@ -977,16 +977,28 @@ class UninterpolatedPlayer extends StatePlayer {
       emote: new BiActionInterpolant(() => this.hasAction('emote'), 0, crouchMaxTime),
       movements: new InfiniteActionInterpolant(() => {
         const ioManager = metaversefile.useIoManager();
-        return  ioManager.keys.up || ioManager.keys.down || ioManager.keys.left || ioManager.keys.right;
+        return  ioManager.keys.up || ioManager.keys.down || ioManager.keys.left || ioManager.keys.right || ioManager.keys.space || ioManager.keys.ctrl;
       }, 0),
       movementsTransition: new BiActionInterpolant(() => {
+        const ioManager = metaversefile.useIoManager();
+        return  ioManager.keys.up || ioManager.keys.down || ioManager.keys.left || ioManager.keys.right || ioManager.keys.space || ioManager.keys.ctrl;
+      }, 0, crouchMaxTime),
```
lalalune, line 985: is crouchmaxtime right for swimming?

lalalune (changes requested):
I have added a review. Mostly cosmetic and PR cleanups.

character-controller.js
gonnavis: `crouchMaxTime` used more like a default value in existing codes.
But it should be better to rename to `defaultMaxTime`.

gonnavis:
@patriboz Vertical swimming works fine.
But if already swim up on surface of water, should turn back to floating or horizontal swimming, instead of keep swim vertically.

And I think should stick on surface until press `C` to dive.

Reference: [https://youtu.be/prreEsPw1Qg](https://youtu.be/prreEsPw1Qg?t=915)

You may need chat with @tcm390 on how to check if on surface.

scenes/water.scn
patriboz: No, we don't need these changes in water.scn. I made them to have more space for swimming while testing. Will remove them.

character-controller.js
patriboz: I will change `crouchMaxTime` to `defaultMaxTime`.

patriboz:
@gonnavis Yes, you are right. I will change the animation accordingly and talk to @tcm390 about this.

character-controller.js
patriboz: I tried to move it up and reuse it, but when doing so, `metaversefile` somehow is only an empty `EventTarget` object and `useIoManager()` isn't defined. So I'll keep it as it is for now.

gonnavis:
- Vertical state to surface state need smooth transition too.
- If on surface not moving, need play floating animation instead of swimming ( even holding `space` key ).
- I feel the sinusoidal motion too "vertically", need turn back to "horizontally" a little.

https://user-images.githubusercontent.com/10785634/179241028-470ea0ab-8c77-4d8e-9c20-6583e92482ec.mp4

image

gonnavis:
Looks like the "sinusoidal motion too vertically" issue caused by out of sync.
It's ok at start, but getting worse after some time, even will "impulse at most vertical pose".

Can try to sync, but maybe good to delete "sinusoidal motion" feature in this PR.
Let this PR focus on vertical swimming due to `space` `C` and camera angle.

PR #3268, opened 2022-07-16
Bake mobs animations to textures on loadtime

jakezira:
- Bake animations to textures on loadtime
- Removes per instance animation indexing


avaer (approved):
Overall, this looks good to me.

It will be hard to test until mobs are restored via the new LOD tasker.

PR #3287, opened 2022-07-19
Lodz

memelotsqui:
Restore from lod.js file, and event dispatch when a chunk block is added


lalalune:
@avaer how we doing with this, did you pull in or is this still relevant?

memelotsqui:
@lalalune this one is no longer elevant, I will be closing the pr :)

Issue #3261, opened 2022-07-15
Unable to hide UI anymore


madjin:
Clicking Hide button and ctrl+h (previous hotkey) doesn't seem to work in latest build

madjin:
tried on chrome win10 by clicking the hide ui button
![image](https://user-images.githubusercontent.com/32600939/179373073-9b5edc90-42bd-4570-beae-35fd5dc5609b.png)
![image](https://user-images.githubusercontent.com/32600939/179373079-001217df-12c0-4a3f-b1f6-2016882147d6.png)

lalalune:
Fixed by https://github.com/webaverse/app/pull/3289

PR #3289, opened 2022-07-19
Hide UI

lalalune:
This revives an old commit which added functionality to hide the UI. Now the "hide" button works again.

Fixes https://github.com/webaverse/app/issues/3261

image

### DIFF
--- a/src/AvatarIcon.jsx
+++ b/src/AvatarIcon.jsx
@@ -109,7 +109,7 @@ const CharacterIcon = () => {
   );
 };
 
-export const AvatarIcon = () => {
+export const AvatarIcon = ({ className }) => {
     const { state, setState } = useContext( AppContext );
 
     const handleCharacterBtnClick = () => {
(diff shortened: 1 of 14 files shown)


alisaad673:
Works fine. Just these minor issues:
1-When logged in, the address doesn't goes away while we hide the UI

https://user-images.githubusercontent.com/35926530/179811729-f5f67dbb-1a3b-4c19-b813-fc8b63eb2914.mp4

2- This little ctrl+h makes the UI look out of sync. Sometimes it isn't even visible due to background. Can we remove it? or either ad it in to the hide button someway

image

lalalune:
@alisaad673 I have fixed these. I think that we should pull it in but maybe make a ticket to add a toast to notify the user that they can press ctrl+h to show UI again. For now it's probably only power users anyways.

alisaad673:
Sure. Will make one.

PR #3291, opened 2022-07-19
Peek buffer (JS)

aryasalehi:
Occlusion culling for chunks ( Terrain Optimization )
In the image bellow I'm drawing the culled chunks 
image

the next step is to port this algo to wasm and add more filters

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -13,11 +13,29 @@ export const rarityColors = {
   legendary: [0xAD00EA, 0x32002D],
 };
 
-export const PEEK_FACE_INDICES = [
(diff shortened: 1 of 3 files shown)


constants.js
```diff
@@ -13,6 +13,12 @@ export const rarityColors = {
   legendary: [0xAD00EA, 0x32002D],
 };
 
+export const PEEK_FACE_INDICES = [
```
avaer, line 16: IMO this should be dynamically computed to make it clear how we arrived at this constant.

avaer (approved):

PR #3304, opened 2022-07-20
experimental 2D

Loryhoof:


avaer:
This is pretty cool, did we want to bring this mode to production?

PR #3145, opened 2022-06-20
Add cryptoavatars on character select component using CA API.

luixxma:
Added CryptoAvatars characters on CharacterSelect component.

Support pagination, ownership, opensource & collection filter.

Characters that you don't own and aren't open source are blocked from being selectable.

This is a video showing funcionalities.
https://drive.google.com/file/d/1KdzDJeUh8VUdghGT1SC0CFUz0WcRKpy2/view?usp=sharing

PR #3305, opened 2022-07-21
Add Cryptoavatars

lalalune:
This PR replaces this one: https://github.com/webaverse/app/pull/3145

We had some bad merge conflicts with current master. This fixes them by implementing the new NpcManager, but might still need some review (do we need the caching?)

TODO:
- [x] Fix this bug, avatars are getting added to world when created as previews
image

- [x] Selection arrow doesn't work on Cryptoavatars, treated differently
image

- [x] Design here is pretty weak, some rounded corners and margins to clean it up
image

- [x] Avatar camera gets cut off if avatar is too close
image


lalalune:
@mavisakalyan remaining bugs:

1. after merge with master, arrow is not showing up
2. refs error with the refmap system in the master branch, maybe this needs to just be redone or put back the way you had it but for all? or we can just set target character on hover and be done with the complexity?
3. some characters are showing up 180 degrees flipped and/or too tall off the ground
>> flipped direction is probably because they are gltf and not vrm, or vice versa. we should detect gltf vs vrm, otherwise we need to get the forward avatar direction and test it
>> too high off the ground... i dont know. maybe they are just too big and we make them smaller? idk

lalalune:
@mavisakalyan 

I guess we had a regression on this-- some of the characters are facing the wrong way and are not framed well. Like our camera is either not getting set or it's set to the wrong pose.

image

lalalune:
Weird border when loading CA avatars, some CSS issue

Also, sometimes a weird crash because target is null. Something to do with spine bone, or maybe it's getting deleted

image

lalalune:
@mavisakalyan please remove any changes to Claims UI or anything not related to this PR and put in a separate PR. Conflicts need to be resolved with master, should be easy. Please post a video of everyhting working and avatars changing.

luixxma:
Hi @lalalune , i have a question about disabled avastars wich cannot be selected. Its fixed? I remember a comment wich says that not owned avatars and not open source avatars could be selected. CryptoAvatarsLoader script returns an attribute of each avatar, called canBeUsed saying what avatar can be selected or not. Thank you for everything :)

Issue #3322, opened 2022-07-22
Update README with minimum requirements to run


justincolangelo:
After setting this up on both an EC2 and a DO droplet it would not load. There were errors about the service stopping. 

```
The service was stopped
The service is no longer running: write EPIPE
```

After resizing the droplet a few times I found that it needed more power to build and run.

PR #3323, opened 2022-07-22
#3322 update readme to include minimum requirements to run

justincolangelo:
Adding minimum requirements to build and run the project. `esbuild` seemed to take up a few GBs of resources and as much CPU as it could (85% at times). This would eventually error out and not build correctly so requests to the server would fail or timeout.


lalalune:
lgtm

PR #3327, opened 2022-07-22
Vertical and Surface Swimming

patriboz:
- Implemented smooth character rotations while swimming vertically. 90deg when ascending on `SPACE` and -90deg when descending on `C`. Combined with `WSAD` results in 45deg or -45deg respectively.
- Implemented surface swimming: Once reaching the surface, the character will swim at surface until `C` is pressed to dive.

### DIFF
--- a/avatars/animationHelpers.js
+++ b/avatars/animationHelpers.js
@@ -34,7 +34,7 @@ import {
 } from './constants.js';
 
 import {
-  defaultMaxTime,
+  defaultActionTransitionTime,
   // useMaxTime,
   aimMaxTime,
   // avatarInterpolationFrameRate,
@@ -833,7 +833,7 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
         const t2 = (now / 1000) % danceAnimation.duration;
         const v2 = src2.evaluate(t2);
 
-        const danceFactorS = avatar.danceFactor / defaultMaxTime;
+        const danceFactorS = avatar.danceFactor / defaultActionTransitionTime;
         const f = Math.min(Math.max(danceFactorS, 0), 1);
         lerpFn
           .call(
@@ -864,7 +864,7 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
         const t2 = Math.min(emoteTime / 1000, emoteAnimation.duration);
         const v2 = src2.evaluate(t2);
 
-        const emoteFactorS = avatar.emoteFactor / defaultMaxTime;
+        const emoteFactorS = avatar.emoteFactor / defaultActionTransitionTime;
         const f = Math.min(Math.max(emoteFactorS, 0), 1);
         lerpFn
           .call(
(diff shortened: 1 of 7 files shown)


constants.js
```diff
@@ -110,7 +110,7 @@ export const metaverseProfileDefinition = `kjzl6cwe1jw145wm7u2sy1wpa33hglvmuy6th
 
 export const audioTimeoutTime = 10 * 1000;
 
-export const crouchMaxTime = 200;
+export const defaultMaxTime = 200;
```
avaer, line 113: This is a poor rename; it's even less accurate than the previous name.

io-manager.js
```diff
@@ -242,7 +242,13 @@ const _updateIo = timeDiff => {
       }
     }
     if (localPlayer.hasAction('fly') || localPlayer.hasAction('swim')) {
-      keysDirection.applyQuaternion(camera.quaternion);
+      if(localPlayer.getAction('swim').onSurface) {
```
avaer, line 245: Io-manager should be forbidden from querying the local player, or any actions.

avatars/animationHelpers.js
```diff
@@ -1253,26 +1254,45 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
         localQuaternion2.fromArray(v2);
         localQuaternion3.fromArray(v3);
         localQuaternion4.fromArray(v4);
+        localQuaternion5.fromArray([0,0,0,1]);
```
avaer, line 1257: Is this an allocation in the frame loop?

io-manager.js
avaer: Also, this is inefficient to check the action and then get it.

avatars/animationHelpers.js
```diff
@@ -1253,26 +1254,45 @@ export const _applyAnimation = (avatar, now, moveFactors, timeDiffS) => {
...
-        localVector4.y += liftSwims;
-        // localVector3.lerp(localVector4, walkRunFactor);
-        // localVector2.lerp(localVector3, idleWalkFactor);
-        localVector3.lerp(localVector4, avatar.sprintFactor);
-        localVector2.lerp(localVector3, avatar.movementsTransitionFactor);
+        if(!avatar.swimmingOnSurfaceState || (avatar.swimmingOnSurfaceState && avatar.horizontalMovementsTransitionFactor > 0)) {
+          const liftSwims = 0.05; // lift swims height, prevent head sink in water
+          localVector3.fromArray(v3);
+          // localVector3.y += 0.21; // align Swimming.fbx's height to freestyle.fbx
+          localVector3.y += 0.03; // align Swimming.fbx's height to freestyle.fbx
```
avaer, line 1287: These hacks do not belong here.

Ideally they do not exist, because the code is fully throught through and doesn't need adjustments.

I would prefer if all of these adjustments were deleted. If they cannot be then:

- base them on the avatar height, not world height, since it scales with avatar height
- make them a constant
- break out these position hacks into separate methods

avaer: Note the surrounding code was also quite bad and was not approved, so please do not copy it.

avaer (changes requested):
Good progress, but the code quality is poor and repeats mistakes of other poor quality code around it.

constants.js
```diff
@@ -137,7 +137,7 @@ export const defaultMaxId = 8192;
 
 export const defaultMusicVolume = 0.35;
 
-export const voicePacksUrl = `https://webaverse.github.io/voicepacks/all_packs.json`;
+export const voicePacksUrl = `http://localhost/voicepacks/all_packs.json`;
```
gonnavis, line 140: Don't commit this, unless the PR is in draft/WIP state.

patriboz:
Ported to studios: https://github.com/webaverse-studios/app/pull/58

PR #3331, opened 2022-07-22
Arya oc

avaer:
Sibling PR of : https://github.com/webaverse/app-wasm/pull/69


instancing.js
```diff
@@ -37,8 +41,8 @@ const peekFaceSpecs = [
 
 const maxNumDraws = 1024;
 
-const isVectorInRange = (vector, min, max) => {
-  return (vector.x >= min.x && vector.x < max.x) && (vector.y >= min.y && vector.y < max.y) && (vector.z >= min.z && vector.z < max.z);
+const isCameraInRange2D = (vector, min, max) => {
```
avaer, line 44: I would call it `isPointInPillar`. It has nothing to do with cameras.

PR #3333, opened 2022-07-22
Fix NPC responses in chat

lalalune:
Currently, when the user walks up to an NPC and chats with them, it's broken -- there is an error since the formatting of messages was changed to send over the network.

This changes the raw message to a message object format.


alisaad673:
Works fine.

PR #3345, opened 2022-07-23
disable activate of objects that don't do anything on activate.

ultrabot2022:
Some objects show "E" when user approach to them, even they don't do anything after user tries to grab.
So added a feature that hide "E" in non-active objects.
Please check the result and give me feedback.

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -756,7 +756,6 @@ class StatePlayer extends PlayerBase {
     return this.getActionsState();
   }
   getActionsState() {
-    if (!this.playerMap) return new Z.Array();
     let actionsArray = this.playerMap.has(actionsMapName) ? this.playerMap.get(actionsMapName, Z.Array) : null;
     if (!actionsArray) {
       actionsArray = new Z.Array();
--- a/character-controller.js
+++ b/character-controller.js
@@ -756,6 +756,7 @@ class StatePlayer extends PlayerBase {
     return this.getActionsState();
   }
   getActionsState() {
+    if (!this.playerMap) return new Z.Array();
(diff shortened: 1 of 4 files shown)


lalalune:
This doesn't fix the case where I, saw, throw a blender cube into the scene. I think we should check if there is a use, grab, etc component on it, and then show the "e" if so. There might even already be a canUse function to test this.

lalalune:
@ultrabot2022 please edit your comment to reference the Github issues this fixes and record a video of it working so QA knows what to look for.

alisaad673:
@ultrabot2022 I tested this but it has a few issues:
-I am unable to equip avatars as E doesn't appears
-I am Unable to equip chest for claims
-I am unable to equip npc to follow the character. 

Can you look into it? Maybe the logic needs to be changed

https://user-images.githubusercontent.com/35926530/181256600-cf01f1f4-5cbf-4842-bec6-000ff8582092.mp4

https://user-images.githubusercontent.com/35926530/181256735-9c17befe-0f52-43f1-be0e-449e20ac1e44.mp4

ultrabot2022:
I added disable logic to only objects that have active components such as 'wear', 'sit', 'use', 'aim', 'drop' before.
But I knew that there are other type of object such as npc, vrm wear after checking the logic again. And I am afraid of whether what type of will be added in the future.
So I added "nonItem" component to non-active object like following and added disable logic to the game.js
The result can be checked in treehouse.scn

lalalune:
I think we'll need to think deeper about this. We should have some logic that evaluates whether the object is usable or not once, and store and retrieve that information -- like a canUse component or something.
By default, GLBs and FBXs should not be usable.

PR #3346, opened 2022-07-23
Remove DOM elements attached to NPCs when destroyed

ultrabot2022:
Hi @lalalune 
Please check this and give me feedback.


lalalune:
It breaks when an NPC is wearing a CharacterStats panel, as they have been before in street.scn (but may have been removed in a recent version of that).

Regardless, the problem is that there isn't a cleanup on apps associated with NPCs. I wonder if that is just the character stats needing to be updated to accomodate destroy?

lalalune:
The problem is the CharacterStats and applications are not destroyed or removed when the NPC is.

To reproduce:
1. Use the street.scn here: https://github.com/webaverse/app/blob/4f169a0d5087eff94f65328d226a9ef16a19e271/scenes/street.scn
This has character stats on an NPC
2. Join world and switch to another world -- are there errors?

Null check returning a [] doesn't solve the deeper issue which is that there are timers running with no objects attached to them, and these are apps that should be destroyed and gc'd

ultrabot2022:
https://user-images.githubusercontent.com/80860637/181392897-682a40b1-6a36-4821-9433-3efd727d026d.mp4

character-controller.js
```diff
@@ -756,10 +757,10 @@ class StatePlayer extends PlayerBase {
     return this.getActionsState();
   }
   getActionsState() {
```
lalalune, line 759: the actions state happens inside the update loop. we actually want this to throw an error if it's being accessed before it's been initialized, otherwise we're hiding a deeper issue. so we want to make sure that we're just not calling getActionsState at all, because the update function has stopped, because the object that is updating has been destroyed

character-controller.js
```diff
@@ -185,7 +185,8 @@ class PlayerBase extends THREE.Object3D {
     this.appManager.addEventListener('appremove', e => {
       if (!this.detached) {
         const app = e.data;
-        app.parent && app.parent.remove(app);
+        const parent = app.parent;
```
lalalune, line 188: This is probably unnecessary because there is always a parent

dom-renderer.jsx
```diff
@@ -9,7 +9,7 @@ const cubicBezier = easing(0, 1, 0, 1);
 const physicsScene = physicsManager.getScene();
 
 class DomItem extends THREE.Object3D {
-  constructor(position, quaternion, scale, width, height, worldWidth, render) {
+  constructor(position, quaternion, scale, width, height, worldWidth, render, domRenderEngine) {
```
lalalune, line 12: domRenderer is the same thing, you should access the singleton, no need to pass the reference in here

dom-renderer.jsx
```diff
@@ -54,6 +54,8 @@ class DomItem extends THREE.Object3D {
     physicsScene.disableActor(physicsObject);
     physicsScene.disableGeometryQueries(physicsObject);
     this.physicsObject = physicsObject;
+
+    this.domRenderEngine = domRenderEngine;
```
lalalune, line 58: remove

dom-renderer.jsx
```diff
@@ -116,6 +118,8 @@ class DomItem extends THREE.Object3D {
   destroy() {
     physicsScene.enableActor(this.physicsObject);
     physicsScene.removeGeometry(this.physicsObject);
+    this.remove(this.floatNode);
+    this.domRenderEngine.removeDom(this);
```
lalalune, line 122: domRenderer.removeDom(this)

dom-renderer.jsx
```diff
@@ -171,7 +175,7 @@ export class DomRenderEngine extends EventTarget {
     worldWidth = 1,
     render = () => (<div />),
   }) {
-    const dom = new DomItem(position, quaternion, scale, width, height, worldWidth, render);
+    const dom = new DomItem(position, quaternion, scale, width, height, worldWidth, render, this);
```
lalalune, line 178: no need

lalalune (commented):
I think that we're still avoiding the deeper fix. We need to do this without altering the actionsArray or faking an empty array there, because we want to get rid of the object that is calling the update in the first place. Removing from DOM seems to be a good part of the cleanup but no need to pass the domRendererEngine instance, domRenderer is itself a singleton

PR #3348, opened 2022-07-23
Disable running and flying of Naruto.

ultrabot2022:
In some case, The user doesn't want to act unrealistic animation such fly and superfast running (Naruto run)
So added a feature to disable them in a given scene.
Please check the result and give me feedback.

https://user-images.githubusercontent.com/80860637/181359104-8ade86ca-c30e-4cba-b277-36b5cca596d0.mp4

Should install totum package to test the result.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@pixiv/three-vrm": "./packages/three-vrm",
     "@react-three/fiber": "^7.0.29",
     "@shaderfrog/glsl-parser": "^0.1.23",
+    "axios": "^0.27.2",
     "classnames": "^2.3.1",
     "encoding-japanese": "^1.0.30",
     "ethers": "^5.6.8",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "@pixiv/three-vrm": "./packages/three-vrm",
     "@react-three/fiber": "^7.0.29",
     "@shaderfrog/glsl-parser": "^0.1.23",
-    "axios": "^0.27.2",
     "classnames": "^2.3.1",
     "encoding-japanese": "^1.0.30",
     "ethers": "^5.6.8",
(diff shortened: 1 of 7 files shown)


lalalune:
The goal is not to disable it always, but to disable it if the scene indicates it should be disabled.

ultrabot2022:
Shall I add a new field in the .scn file to indicate the naruto running and flying should be disabled, like the following?

lalalune:
yes but probably should be like
```
"settings": {
  fly: false,
  narutoRun: false
}
```

lalalune:
Here are the steps:

https://raw.githubusercontent.com/webaverse/app/2c99b58d4ae01b8e4b507ee375ce5939cd26dec0/scenes/club.scn

1. Copy the "rendersettings" application and rename the copy to "scenesettings".
2. Look where rendersettings is being initialized and loaded, and let's copy and paste all of that but handle with scene settings
3. These settings should probably be stored on the gameManager singleton, import gameManager from './game-manager.js'

avaer:
@ultrabot2022 this is not an acceptable PR description.

avaer:
Here is a more descriptive PR description: https://github.com/webaverse/dc-wasm/pull/16

Additionally, if the PR can only be understood by one specific person it is not a good PR.

package.json
```diff
@@ -18,6 +18,7 @@
     "@pixiv/three-vrm": "./packages/three-vrm",
     "@react-three/fiber": "^7.0.29",
     "@shaderfrog/glsl-parser": "^0.1.23",
+    "axios": "^0.27.2",
```
lalalune, line 21: get rid of axios dependency

lalalune (commented):
This seems to be the right pattern, but I think we can import the sceneSettingsManager as a singleton and use that instead of referencing the metaversefileApi.

@avaer might have some ideas on naming conventions here

PR #3380, opened 2022-07-27
Press 'R' to log current state to console

lalalune:
I find myself very often having to log the current state of the world, scene, localplayer, etc.

This adds a handy function to webaverse.js to log these things to the console.


avaer:
R is the drop key.

lalalune:
@avaer capital R only, but I can make it something else -- } ?

alisaad673:
@avaer right now 'p' is the drop key in master

lalalune:
changed to }

PR #3385, opened 2022-07-27
Integrate mobs physics onto new api

jakezira:


jakezira:
This needs to be merged to keep in sync: https://github.com/webaverse/totum/pull/119

Issue #3389, opened 2022-07-28
Facial expressions get auto-set(corner case)


alisaad673:
Facial expressions get auto-set(doesn't gets reset) in case we trigger another emote while one emote is in progress.

https://user-images.githubusercontent.com/35926530/181501343-011080c6-ad9c-446e-987b-0d8399e0ffca.mp4

It works fine if we wait for the emote to finish( facial expressions get reset here)

https://user-images.githubusercontent.com/35926530/181501306-a8468154-a34d-4b1b-888d-b59d142da1d0.mp4

PR #3403, opened 2022-07-28
Fix facial expressions get auto-set bug

tcm390:
related: https://github.com/webaverse/app/issues/3389

Issue:

https://user-images.githubusercontent.com/60634884/181650968-8793d554-f6f7-4b56-b75a-0f7e91d13c3c.mp4

Result:

https://user-images.githubusercontent.com/60634884/181650986-dc7fc2af-0dec-4d34-a88f-550c47109711.mp4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2898,8 +2898,6 @@
       "dependencies": {
         "node-fetch": {
           "version": "2.6.6",
-          "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.6.tgz",
-          "integrity": "sha512-Z8/6vRlTUChSdIgMa51jxQ4lrw/Jy5SOW10ObaA47/RElsAN2c5Pn8bTgFGWn/ibwzXTE8qwr1Yzx28vsecXEA==",
           "requires": {
             "whatwg-url": "^5.0.0"
           }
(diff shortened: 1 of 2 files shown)


alisaad673:
@tcm390 though this bug is solved, there is another corner case if you can look into it plz

-In case we select a facial expression lets say 'happy' and then we perform an emote that includes that facial expression lets say 'victory'. The previously set 'happy' facial expression would be reset and would go away. it doesn't happens in case the performed emote doesn't contain the previously selected facial expression.

https://user-images.githubusercontent.com/35926530/181916644-0f8d3d63-2fcf-4754-9e79-023d00b864a2.mp4

tcm390:
[maintain user previous facial expression](https://github.com/webaverse/app/pull/3403/commits/dea3d4bcbcbad515d069a496ca05f8618c5ab407)

Result:

https://user-images.githubusercontent.com/60634884/182469129-381e6f68-2a40-45dd-8dff-033ad8f44aef.mp4

avaer (changes requested):
From the looks of it, this design isn’t thinking ahead. There could be other places we trigger emotes in the future and then this code would be even more broken.

The correct way is not to store the old emote but to have an expression recompute function which you can call anytime to update the emote from the current state without any reference to past state.

Issue #3400, opened 2022-07-28
app crashes while we equip dragon after equipping silsword


alisaad673:
App is crashing if we equip silsword first and then try to equip dragon, It throws this error

![image](https://user-images.githubusercontent.com/35926530/181569093-9cda9246-ea5a-4e7c-95f9-20f7f2535dbb.png)

https://user-images.githubusercontent.com/35926530/181569136-46f176c9-3f8e-4838-9ba1-3f43438b194e.mp4

avaer:
Fixed by https://github.com/webaverse/app/issues/3400

alisaad673:
Reopening because of a regression in the previous fix.

alisaad673:
fixed with: https://github.com/webaverse/app/pull/3440

PR #3404, opened 2022-07-29
Added randomString() for instanceId

memelotsqui:
-Apps loaded from Equipment were created with an instanceId = '' (not setup)
-When searching by instance by Id, every object from the inventory matched instanceId = '', so it would return the first one it found (sword)
-Sword did not have the same functionality as mount, but had the same id, so it crashed the app trying to use sword as mount.

### DIFF
--- a/game.js
+++ b/game.js
@@ -25,7 +25,7 @@ import npcManager from './npc-manager.js';
 import raycastManager from './raycast-manager.js';
 import zTargeting from './z-targeting.js';
 import Avatar from './avatars/avatars.js';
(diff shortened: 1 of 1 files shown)


lalalune:
@memelotsqui we use makeId(5) elsewhere (also in util.js), I think that would be preferable to stay with convention.

memelotsqui:
Alright!, Will change to makeId and push changes again

avaer:
This is an undocumented fix for https://github.com/webaverse/app/issues/3400

Issue #3398, opened 2022-07-28
app gets stuck while we try to switch scene with switched npc('g' key)


alisaad673:
If we switch to the nearest npc with 'g' key, and then try to switch scene to any other scene. It throws this error and app gets stuck.

![image](https://user-images.githubusercontent.com/35926530/181526164-255c31ae-a0db-4749-83c8-78e8935ead96.png)

https://user-images.githubusercontent.com/35926530/181526195-eeccbe18-75e8-491f-8197-c55e48cc53b3.mp4

PR #3405, opened 2022-07-29
Fix 3398 -- trying to destroy remote player that doesn't exist

lalalune:
Fixes https://github.com/webaverse/app/issues/3398 by null checking remotePlayer


lalalune:
If this is happening because of NPCs then maybe we should rethink our algo here

avaer (changes requested):
IMO this should never happen. If there is no player in the specs then it should have been removed from this map by the listener.

If it's not being removed by the listener then the listener has a bug.

PR #3406, opened 2022-07-29
Fix clipped MegaHup

lalalune:
This PR fixes the clipping issues on the avatar selection canvas, and puts the avatar behind the tokens + inside the menu object so that the tokens are selectable even with the wider canvas.

So this fixes both the ugly hard clipping -- which is more apparent in the Cryptoavatars even -- and the layering when the winder is small.

Current:

image

Fixed:

image


avaer:
This makes the canvas screen sized-right? I guess that's fine, it won't cost more draw time.
It wastes more memory for `MegaHup` but I don't see any other way to clean up clipping for thicc models.

avaer (approved):

PR #3434, opened 2022-07-30
Arya bone [WIP]

avaer:


avaer:
This is no longer needed; it was used to debug the fact that we need to upgrade THREE.js to fix morphs.

PR #3435, opened 2022-07-31
Update three.js and three-vrm

lalalune:
This PR updates three.js to r143 and three-vrm to the latest, supporting VRM 1.0

Note: there are some color encoding issues with apps as a result of sRGB encoding largely being removed from three.js

A couple of objects in the street look weird but everything else looks fine. In the next PR we can completely remove sRGB and use a pure linear flow, which should give us a little bit better color overall.


avatars/avatars.js
```diff
@@ -562,11 +562,12 @@ class Avatar {
...
     this.eyeballTargetPlane = new THREE.Plane();
     this.eyeballTargetEnabled = false;
-
+    this.springBoneManager = null;
     if (options.hair) {
-      this.springBoneManager = new VRMSpringBoneImporter().import(object);
-    } else {
-      this.springBoneManager = null;
+      (async () => {
+        this.springBoneManager = await (new VRMSpringBoneImporter()).import(object);
```
avaer, line 568: I don't think this can be async. I believe the existing `three-vrm` fork specifically modifies `VRMSpringBoneImporter` to be sync, because it is unacceptable for a constructor to be async.

PR #3445, opened 2022-08-01
complete_inventory_mint

hedera07:
Mint to


src/hooks/web3-constants.js
```diff
@@ -0,0 +1,142 @@
+import _CONTRACTS from 'https://contracts.webaverse.com/config/addresses.js';
```
avaer, line 1: Why is this file under `hooks`?

Imo constants should be exposed higher up. We already have a constants file, and although this one is web3 specific, I would never look for it here in a React subdirectory.

util.js
```diff
@@ -943,6 +943,14 @@ export const handleDropJsonItem = async (item) => {
   }
   return null;
 };
+export const handleBlobUpload = async (name, blob, progress) => {
+  const formData = new FormData();
+  formData.append(name, blob, name);
+  const hashes = await doUpload(`https://ipfs.webaverse.com/`, formData, progress);
+
+  const rootDirectoryHash = hashes.length > 0 ? hashes[0].hash : null;
```
avaer, line 951: This isn't used?

src/DragAndDrop.jsx
```diff
@@ -248,16 +271,94 @@ const DragAndDrop = () => {
...
+    if (mintComplete) {
+      setTimeout(() => {
+        setMintComplete(false);
+      }, 6000);
+    }
+  }, [mintComplete]);
+
+  useEffect(() => {
+    if (error) {
+      setTimeout(() => {
```
avaer, line 284: - This seems arbitrary, and this kind of complex timing logic should happen in a popups/errors system component, not `DragAndDrop`.
- There is no cancellation so it will conflict with other errors
- This number should be a constant

src/DragAndDrop.jsx
```diff
@@ -248,16 +271,94 @@ const DragAndDrop = () => {
...
+      }
+      setPreviewImage(previewURL);
+    });
+  }
+
+  useEffect(() => {
+    if (!currentApp) return;
+
+    if (APP_3D_TYPES.includes(currentApp.appType)) {
+      setTimeout(() => {
```
avaer, line 319: Another timing issue.

util.js
```diff
@@ -943,6 +943,14 @@ export const handleDropJsonItem = async (item) => {
   }
   return null;
 };
+export const handleBlobUpload = async (name, blob, progress) => {
+  const formData = new FormData();
+  formData.append(name, blob, name);
+  const hashes = await doUpload(`https://ipfs.webaverse.com/`, formData, progress);
```
avaer, line 949: This should be a constant.

src/hooks/useNFTContract.js
```diff
@@ -0,0 +1,311 @@
...
+    return provider.getSigner(currentAccount);
+  }
+
+  const getContract = async () => { // NFTcontract
+    const simpleRpcProvider = new ethers.providers.StaticJsonRpcProvider(selectedChain.rpcUrls[0]);
+    const contract = new ethers.Contract(NFTcontractAddress, NFTABI, simpleRpcProvider);
+    return contract;
+  };
+
+  async function mintNFT(currentApp, previewImage, callback = () => {}, afterminting = () => {}) {
```
avaer, line 63: This entire code block is basically repeated in `https://github.com/webaverse/app/pull/3445/files#diff-4fe5770969890d344194764aa4b1778f5b6385e3aab7914b4ae69e9bac9a3bfeR166, so it should probably be better abstracted.

src/hooks/useNFTContract.js
```diff
@@ -0,0 +1,311 @@
...
+      const metadataHash = json_hash.split(FILE_ADDRESS)[1].split('/')[0];
+      const webaverseContract = new ethers.Contract(WebaversecontractAddress, WebaverseABI, signer);
+      // const NFTcontract = new ethers.Contract(NFTcontractAddress, NFTABI, signer);
+      const FTcontract = new ethers.Contract(FTcontractAddress, FTABI, signer);
+
+      const bigMintFee = await webaverseContract.mintFee();
+      const mintfee = BigNumber.from(bigMintFee).toNumber();
+      if (mintfee > 0) { // webaverse side chain mintfee != 0
+        const FTapprovetx = await FTcontract.approve(NFTcontractAddress, mintfee); // mintfee = 10 default
+        const FTapproveres = await FTapprovetx.wait();
```
avaer, line 232: `FTapprovers`?

src/hooks/web3-constants.js
```diff
@@ -0,0 +1,142 @@
...
+import _CONTRACT_ABIS from 'https://contracts.webaverse.com/config/abi.js';
```
avaer, line 2: If we are importing the abi... then why are we also uploading the ABI to `app`?

We need to decide what we are doing and remove dead constants.

avaer (changes requested):
This looks fine, it just needs cleanup.

src/hooks/useNFTContract.js
hedera07: FTapproveres is for checking transactionHash.
if(FTapproveres.transactionHash)

util.js
CoryITpro: As you can see, this value is used in the next line.

hedera07: As you can see, this value is used in the next line.

util.js
hedera07: const ipfsFileURL = 'https://ipfs.webaverse.com/';
const ipfsFolderURL = 'https://ipfs.webaverse.com/ipfs';

src/hooks/web3-constants.js
hedera07: It's dead constant.
I removed it.

src/hooks/web3-constants.js
hedera07: It's a dead constant.
I removed it.

src/hooks/useNFTContract.js
hedera07: mintNFT and mintfromVoucher are basically same.
But currently mintfromVoucher is for testing.
I will update this when serve drops feature and voucher server are completed.

src/DragAndDrop.jsx
hedera07: fixed as return clearTimeout.

src/DragAndDrop.jsx
hedera07: The popups/errors system component is <GenericLoadingMessage>.
But it is used different timecount(3000, 6000) for each GenericLoadingMessage>.
And fixed conflict with other erros as return clearTimeout

avaer:
Closing due to poor title and nonexistent description.

This pr has a high likelihood of needing a revert so the description is required.

Issue #3423, opened 2022-07-30
Character Diorama rotation issue


alisaad673:
With the recent fix for diorama getting out of frame, though that issue is fixed but the rotation of character in frame has become weird in both character diorama and chat diorama.

previous one:

https://user-images.githubusercontent.com/35926530/181908104-7606d086-4109-4fed-9fed-0f30ec77317a.mp4

Current One:

https://user-images.githubusercontent.com/35926530/181908106-f206535a-9c33-4d20-a090-c24f3fdb91b5.mp4

alisaad673:
This was fixed

PR #3447, opened 2022-08-01
Fixed character diorama rotation issue.

ultrabot2022:
## Describe your changes
The rotation of character in frame was weird in both character diorama and chat diorama. And when selecting another character in the character-select panel, the avatar of chat diorama wasn't changed to the selected one. So I updated related code.

1. Press Enter and input text to test Chat diorama.
2. Press Tab and then press "character select" button on the left bottom corner to test Character diorama.

3423 https://github.com/webaverse/app/issues/3423

## Screenshots and/or video

https://user-images.githubusercontent.com/80860637/182253993-603f8e9e-06f6-4fd4-baf9-8c8dfd4e679f.mp4

- [ Y ] I have performed a self-review of my code
- [ Y ] I am not adding any irrelevant code or assets
- [ Y ] I am only including the changes needed to implement the change
- [ N ] I have playtested and intentionally tried to find error cases but couldn't

### DIFF
--- a/game.js
+++ b/game.js
@@ -1673,20 +1673,21 @@ class GameManager extends EventTarget {
       outline: true,
       grassBackground: true,
       // glyphBackground: true,
+      cameraOffset: new THREE.Vector3(0, 0, -0.6),
     });
     localPlayer.addEventListener('avatarchange', e => {
       this.playerDiorama.setObjects([
         e.avatar.model,
       ]);
-      let neckBone;
+      let headBone;
       e.avatar.model.traverse(
         (object) => {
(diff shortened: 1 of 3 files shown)


lalalune:
It still looks a little forced, like it's stuck on the character's face

We want it to be in front of the character but not locked to their skeleton every frame. It looks odd. We could smooth interpolate the position of the camera, which would look better, or get the Y offset from skeleton but otherwise keep it how it was (or both!)

jakezira:
It could've worked in previous versions. Better to find out the reason where it broke and fix it rather than designing new features.

PR #3449, opened 2022-08-02
Fix npc spawning at avatar position while the app is loading

jakezira:
Hide npc rendering till physics enabled

### DIFF
--- a/npc-manager.js
+++ b/npc-manager.js
@@ -61,7 +61,7 @@ class NpcManager extends EventTarget {
     }
 
     await npcPlayer.setAvatarUrl(avatarUrl);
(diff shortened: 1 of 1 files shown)


alisaad673:
Works Fine.

avaer:
I think we want to make the whole npc app, including the character controller, to be in the correct place, not just make it invisible while loading.

I think this is more about setting the `position`/`matrixWorld` early enough directly from the initialized `app`.

jakezira:
Fixed npc position by update avatar instead of hiding on loading time.

avaer (approved):

Issue #3428, opened 2022-07-30
Floating swords if dropped while being used


alisaad673:
Steps to reproduce:
1-Equip a sword
2-Drop it while performing combat( press 'r' while using the sword)
3-Sword would be floating
4- Equip another sword from character tab and drop it the same way

Issues:
-If we drop the sword right while we are doing combat with it, it starts floating around in the scene.
-After dropping, in case we equip another sword, lets say silsword and drop it too while doing combat, the previous sword appears along with the recently dropped one.

One solution could be to disable dropping it while it is being used

https://user-images.githubusercontent.com/35926530/181915526-e2aa7561-cc64-4b52-adc6-864e1a656e8c.mp4

alisaad673:
Fixed : https://github.com/webaverse/app/pull/3526

PR #3453, opened 2022-08-02
Fix: Floating swords if dropped while being used #3428

patriboz:
Fix for: https://github.com/webaverse/app/issues/3428
The floating sword after unwearing while using it is caused by an active use-action in the moment of unwearing.
This fix checks if there is a current active use-action. If so, it will be removed and then the weapon will be dropped to the ground.

It was suggested to disable unwearing while using the sword, but I found this might decrease a fluid gameplay experience, so I decided to fix it as above.

### DIFF
--- a/app-manager.js
+++ b/app-manager.js
@@ -39,8 +39,6 @@ class AppManager extends EventTarget {
     
     this.appsArray = null;
     this.apps = [];
-
-    this.transform = new Float32Array(10);
     
     this.pendingAddPromises = new Map();
     // this.pushingLocalUpdates = false;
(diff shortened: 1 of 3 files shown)


alisaad673:
Though it fixes the issue of floating swords but it has following issues:
1-These errors are appearing in console on loading the app:

image

2-No previews for items anymore. 

image

3-Pfp isn't loading anymore

image

4-This error appears while we pick up silk from chest

image

https://user-images.githubusercontent.com/35926530/182613414-be0e2b31-b010-4736-a257-b5e9a0e5d491.mp4

5-These errors appear while we select any of the items from character panel:

image

https://user-images.githubusercontent.com/35926530/182613643-a9cbe1b0-0c22-495b-91cc-91ede99dfd87.mp4

6-Pressing the 'g' key to switch to npc, its not able to move.

https://user-images.githubusercontent.com/35926530/182614918-9a529df7-3608-4790-8dc6-f9c8fa953c4d.mp4

7-Also has this issue of sonic boom disappearing for a moment, when near to this effect.

https://user-images.githubusercontent.com/35926530/182616002-3cc8d9e4-35a8-4b17-9b08-f793dc9e60cc.mp4

8-No previews of objects here when we select them:

image

alisaad673:
It works fine now. Just one issue. Can you fix this issue fit bow too? that happens in case we drop it while using. The string remains attached to the right hand even though the bow is dropped

https://user-images.githubusercontent.com/35926530/182661139-95fa469a-4577-4d62-897c-6488fcd8e79b.mp4

.

patriboz:
Issue seems to be resolved after I reverted `app-manager.js` to this version: https://github.com/webaverse/app/blob/f63db7e9b79070869e60e52ca031866b1e375ddd/app-manager.js

Issue #3452, opened 2022-08-02
Crash while changing scene with equipped dragon


alisaad673:
**Describe the bug**
While we have equipped dragon from character panel, and then we try to change scene, it throws an error and the vrm appears stuck at spawn point

**To Reproduce**
1. Equip dragon from character panel
2. Switch scene to any other scene and error appears

**Expected behavior**
The scene should change without an error and without crashing, along with dragon

**Screenshots & Recordings**
![image](https://user-images.githubusercontent.com/35926530/182374320-54513889-9525-44b7-89aa-5adbaced6a8d.png)

https://user-images.githubusercontent.com/35926530/182374403-0b58236f-3c6c-4bb5-bf8d-eca9abd5b2e5.mp4

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome

PR #3457, opened 2022-08-02
changed from toJSON to array copy

memelotsqui:
## Describe your changes
-changed toJSON() to make a copy of the the desired array.
-toJSON was entering an infinite loop (I believe it has to do with hitTracker being a property and a parent of an object)

1-Equip a Dragon
2-Load a scene
3-Scene should not be crashing

#3452 

## Screenshots and/or video

https://user-images.githubusercontent.com/1117257/182483592-ffe0ae98-1ac5-454f-85e1-0315bfcac782.mp4

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -603,7 +603,7 @@ class StatePlayer extends PlayerBase {
   }
   bindCommonObservers() {
     const actions = this.getActionsState();
-    let lastActions = actions.toJSON();
+    let lastActions = [...actions.binding.e]
     const observeActionsFn = () => {
       const nextActions = Array.from(this.getActionsState());
       for (const nextAction of nextActions) {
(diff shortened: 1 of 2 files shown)


alisaad673:
Works fine

avaer (changes requested):
> -toJSON was entering an infinite loop (I believe it has to do with hitTracker being a property and a parent of an object)

This object is the `zjs` state. It must be JSONable, attaching things like THREE.js `Object3D` to it is illegal.

Therefore it seems like some other code is destroying the zjs state object, and we need to fix that.

avaer:
Also, please don't do this for this PR (the zjs state must be serializable so this will not be a legal solution), but there is also the `structuredClone` method: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

memelotsqui:
Alright!, right now im investigating because I believe the issue comes from HitTracker, thats the one getting called in loop, I will find a different solution to make this JSONable :)

avaer:
Needs a totally different solution, so closing.

Issue #3458, opened 2022-08-03
"sit" component not working properly (character physics issue)


platote:
**Bug description**
Everything freezes when using the "sit" component.  "TypeError: Cannot read properties of undefined (reading 'scene')"
Related to #3455, not restricted to planes.

**To Reproduce**

- Import the [plane](https://github.com/webaverse/plane) in a webaverse scene OR make a custom three.js object, with a "sit" component using useActivate() and try to use the vehicle
- See error

**Expected behavior**
The character should be just riding the vehicle. 

**Screenshots**
![image](https://user-images.githubusercontent.com/84842352/182619551-b67e20df-88dc-4d30-9824-8a4cfa7517d5.png)
Adding new screenshots, but same error as in #3455 while being unrelated problem

memelotsqui:
Thank you for reporting this issue :), I will try to fix it.

alisaad673:
Fixed

PR #3462, opened 2022-08-03
traverse on app instead of glb.scene

memelotsqui:
## Describe your changes
-Changed from app.glb.scene.traverse to app.traverse to make sure it doesnt falls into null error.

-When creating a custom app with index.js, the app no longer contains app.glb property, so when it tries to access app.glb.scene it drops the error `Cannot read properties of undefined (reading 'scene')` breaking this part of the code

-We're using Object3D.traverse (https://threejs.org/docs/#api/en/core/Object3D.traverse) to find in any of the children a skinned mesh that contains a bone with sitbone component by name (also optional set by the user), directly in app.glb.scene.

-My suggestion is to use Object3D.traverse directly on the app (that is also an object3D), instead of trying it always on the .glb.scene (that will be null when an index is defined), this will avoid future issues.

1- load any mount that uses as start_url index.js instead of a .glb file
2- app will throw error

#3458  

## Screenshots and/or video

Without Fix (app freezes)

https://user-images.githubusercontent.com/1117257/182714612-8cae9faa-0e52-492c-9404-ade85ac4fa2e.mp4

With Fix with an index.js file: (Im aware some errors are present, but this bugs/errors have to do with the plane app and not this specific bug )

https://user-images.githubusercontent.com/1117257/182714258-4795efdf-bae8-4071-bfbe-6ab0b5956ab3.mp4

With fix with a normal .glb file (hovercraft)

https://user-images.githubusercontent.com/1117257/182715251-b5b6a448-9591-4aaa-ac7f-7c04d5e98411.mp4

- [ ] I have playtested and intentionally tried to find error cases but couldn't

### DIFF
--- a/character-physics.js
+++ b/character-physics.js
@@ -202,7 +202,7 @@ class CharacterPhysics {
 
         // Patch fix to fix vehicles and mounts for now
         let rideMesh = null;
-        controlledApp.glb.scene.traverse(o => {
+        controlledApp.traverse(o => {
           if (rideMesh === null && o.isSkinnedMesh) {
             rideMesh = o;
           }


avaer (approved):
Makes sense to me, needs testing.

Issue #3396, opened 2022-07-28
Double pressing 'j'(fire) throws error & makes fire unable to disappear with 'j'


alisaad673:
While we press 'j' once, fire appears, if we press it again, it disappears. But if we press 'j' twice, simultaneously, Fire appears but throws this error and upon pressing 'j' again, fire doesn't disappears.

![image](https://user-images.githubusercontent.com/35926530/181520257-2f4348f8-b8e1-47a2-91da-23515ae15993.png)

https://user-images.githubusercontent.com/35926530/181520285-0f625c54-2d87-4457-8670-026559e84d16.mp4

alisaad673:
Fixed

PR #3468, opened 2022-08-04
add abort flag

memelotsqui:
## Describe your changes
-Added abort flag in case of destroying app before promise gets resolved

-tap 'j' once, fire will come
-tap 'j' again, fire will go
-tap 'j' twice,  fire wont appear, but no error will occur this time

#3396 

## Screenshots and/or video

https://user-images.githubusercontent.com/1117257/182752823-6f6b1c54-1aaa-4ec7-bdc4-fda15fbedf1b.mp4

### DIFF
--- a/metaverse_modules/firedrop/index.js
+++ b/metaverse_modules/firedrop/index.js
@@ -33,6 +33,8 @@ export default e => {
 
   let particleSystem = null;
   const particles = [];
+
+  let abort = false;
   e.waitUntil((async () => {
     {
       const localParticleSystem = particleSystemManager.createParticleSystem({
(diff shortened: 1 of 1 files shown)


metaverse_modules/firedrop/index.js
```diff
@@ -33,6 +33,8 @@ export default e => {
 
   let particleSystem = null;
   const particles = [];
+
+  let abort = false;
```
avaer, line 37: Rather than `abort`, we should call this `live`, which is the convention in the rest of the codebase.

The more modern way is to use https://developer.mozilla.org/en-US/docs/Web/API/AbortController

avaer (changes requested):
Recommending a small refactor.

metaverse_modules/firedrop/index.js
memelotsqui: Alright :), will use 'live' to keep convention, but can change it to AbortController if required.

avaer (approved):

PR #3489, opened 2022-08-04
App would crash If avatar does not have springBones

tcm390:
check whether avatar has springBones before updating wind effect 

Issue:

https://user-images.githubusercontent.com/60634884/182854256-560b2270-e1c3-46e0-b484-3e5428a01a8d.mp4

Result:

https://user-images.githubusercontent.com/60634884/182854299-a44c88e4-96a0-41c7-a9b2-06863c9f0edc.mp4

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -1981,7 +1981,7 @@ class Avatar {
     // update wind in simulation
     const _updateWind = () =>{
       const headPosition = localVector.setFromMatrixPosition(this.modelBoneOutputs.Head.matrixWorld);
-      // The avatar may not have spring bones, so we should make sure the avatar has springBone before update the wind effect
+      // The avatar may not have spring bones, so we should make sure the avatar has springBone before updating the wind effect
       this.springBoneManager && wind.update(timestamp, headPosition, this.springBoneManager)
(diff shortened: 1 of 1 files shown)


avatars/avatars.js
```diff
@@ -1981,7 +1981,7 @@ class Avatar {
     // update wind in simulation
     const _updateWind = () =>{
       const headPosition = localVector.setFromMatrixPosition(this.modelBoneOutputs.Head.matrixWorld);
-      wind.update(timestamp, headPosition, this.springBoneManager)
+      this.springBoneManager && wind.update(timestamp, headPosition, this.springBoneManager)
```
avaer, line 1984: Can you add a comment here to clarify that some avatars will not have spring bones?

avaer (changes requested):
Recommending a comment.

avatars/avatars.js
tcm390: Added comment

https://github.com/webaverse/app/pull/3489/files

alisaad673:
Tested it with the cyberpunk.vrm its working fine

avaer (approved):

Issue #2939, opened 2022-05-04
Profile Pic does not auto scale for different avatar types


ahadshams:
![image](https://user-images.githubusercontent.com/44798013/166647595-85b52b0c-32b8-400b-a4d0-e0c1eed31f2d.png)

@alisaad673 pls test further and expand the issue

avaer:
It should be doable to compute the head bounding box and then zoom the camera back.

We already effectively do this triangle math for the decapitation feature.

PR #3491, opened 2022-08-04
resize avatar icon

tcm390:
set `cameraOffset` based on the avatar head size in avatar-screenshotter.js

related: https://github.com/webaverse/app/issues/2939

result: https://github.com/webaverse/app/pull/3491#issuecomment-1205832633

### DIFF
--- a/avatar-screenshotter.js
+++ b/avatar-screenshotter.js
@@ -86,7 +86,7 @@ export const screenshotPlayer = async ({
   };
   const localVector = new THREE.Vector3();
   const _updateTarget = (timestamp, timeDiff) => {
-    const decapitatePosition = localVector.setFromMatrixPosition(player.avatar.modelBones.Head.savedMatrixWorld);
+    const neckPosition = localVector.setFromMatrixPosition(player.avatar.modelBones.Head.savedMatrixWorld);
     let avatarHighestPos = 0;
     let tempMesh = null;
     player.avatar.model.traverse(o => {
@@ -98,7 +98,7 @@ export const screenshotPlayer = async ({
       }
     });
     avatarHighestPos += tempMesh.position.y;
-    let headHeight = avatarHighestPos - decapitatePosition.y;
+    let headHeight = avatarHighestPos - neckPosition.y;
     const fov = 50 * ( Math.PI / 180 );
     let cameraZ = headHeight / 2 / Math.tan( fov / 2 );
     cameraZ *= 1.5; //multiply offset so that avatar does't fill the icon
--- a/avatar-screenshotter.js
+++ b/avatar-screenshotter.js
@@ -89,20 +89,28 @@ export const screenshotPlayer = async ({
     const neckPosition = localVector.setFromMatrixPosition(player.avatar.modelBones.Head.savedMatrixWorld);
     let avatarHighestPos = 0;
     let tempMesh = null;
+    
     player.avatar.model.traverse(o => {
       if (o.isMesh) {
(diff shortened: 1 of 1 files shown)


tcm390:
Result

current App | Resize |
:-------------------------:|:-------------------------:
image
| 
image

lalalune:
output looks good but @tcm390 can we rename decapitatePosition since this code isn't for decapitation?

tcm390:
[rename neckposition](https://github.com/webaverse/app/pull/3491/commits/cdfc7977faf4aedec4a7663e53c88913e6a813b5)

avatar-screenshotter.js
```diff
@@ -84,10 +84,30 @@ export const screenshotPlayer = async ({
...
     };
   };
+  const localVector = new THREE.Vector3();
   const _updateTarget = (timestamp, timeDiff) => {
+    const neckPosition = localVector.setFromMatrixPosition(player.avatar.modelBones.Head.savedMatrixWorld);
+    let avatarHighestPos = 0;
+    let tempMesh = null;
+    player.avatar.model.traverse(o => {
+      if (o.isMesh) {
+        for(let i = 0; i < o.geometry.attributes.position.array.length / 3; i++){
```
avaer, line 94: This is extremely slow.

Instead of this, can we use the physics bounding box or some other heuristic? We can't check all vertices.

avaer (changes requested):
Good idea, but this `geometry` technique won't work. We should use some other bounding box or heuristic.

tcm390:
[use bounding box to get highest point](https://github.com/webaverse/app/pull/3491/commits/4edc3fec8e5260799d16801bb979746ea654db2c)

Result |
:-------------------------:|:-------------------------:
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|
image|

alisaad673:
It has improved alot for most of the vrms. Some of the vrms such as meebits still appear a bit zoomed in. But i guess its an improvement.

avatar-screenshotter.js
```diff
@@ -84,10 +84,40 @@ export const screenshotPlayer = async ({
...
+  const boundingBox = new THREE.Box3();
   const _updateTarget = (timestamp, timeDiff) => {
+    const neckPosition = localVector.setFromMatrixPosition(player.avatar.modelBones.Head.savedMatrixWorld);
+    let avatarHighestPos = 0;
+    let tempMesh = null;
+    player.avatar.model.traverse(o => {
+      if (o.isMesh) {
+        if(!o.geometry.boundingBox){
+          const position = o.geometry.attributes.position;
+          boundingBox.setFromBufferAttribute( position );
```
avaer, line 97: This is still slow.

Also, this is available as a method on the `Geometry`: https://threejs.org/docs/#api/en/core/BufferGeometry.boundingBox

tcm390: Yes, but there are a few avatars do not have `boundingBox`, so need to compute the bounding box manually. 
This line is for the avatars that do not have the bounding box

avaer: But you can call computeBoundingBox instead.

tcm390: Hi @avaer, but I found that computeBoundingBox is not that accurate.
image

In this picture, the yellow box is the result by using `o.geometry.computeBoundingBox();` and the red ball is the result by using the method above. 

After I tested it, I found it is caused by `morphAttributesPosition` in three.js.
unknown (1)

PR #3493, opened 2022-08-04
Fixed avatar capsule to cover the whole model

jakezira:
image

### DIFF
--- a/physics-manager.js
+++ b/physics-manager.js
@@ -616,8 +616,11 @@ class PhysicsScene extends EventTarget {
         position,
         physicsId
       )
+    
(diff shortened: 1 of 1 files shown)


avaer (approved):

avaer:
@alisaad673 can you double-check sure this is good for all avatars? I'm sure @jakezira checked already but we it's good to have two people sign off.

avaer:
I'm gonna merge this anyway; if it's broken we can update/revert since this probably isn't a regression.

avaer:
@jakezira also note that I think you have a new serious avatar icon rendering bug in the top left?

Issue #3501, opened 2022-08-05
Dragon icon preview rotation regressed


alisaad673:
**Describe the bug**
The dragon icon rotation is regressed.

**To Reproduce**
1. Run the app on latest master
2. open character-panel

**Expected behavior**
It should rotate as like the sword icon. A simple side view rotation of object.

**Screenshots**
![image](https://user-images.githubusercontent.com/35926530/183089244-27904f70-e249-4cb0-9ecf-729f3012bb33.png)

alisaad673:
Irrelevant

PR #3516, opened 2022-08-06
Dragon icon preview rotation bug

tcm390:
This issue is caused by that the origin of dragon glb is not set to geometry center.  In this PR, I set bounding box center to app position in `object-spriter.js`.

related:
https://github.com/webaverse/app/issues/3501

Result:

https://user-images.githubusercontent.com/60634884/183229241-f1e52d21-1f4b-479d-a4fa-f4a2b4fab58b.mp4

### DIFF
--- a/object-spriter.js
+++ b/object-spriter.js
@@ -112,10 +112,7 @@ const createObjectSpriteInternal = (app, {
       if (physicsObjects.length > 0) {
         const physicsObject = physicsObjects[0];
         const {physicsMesh} = physicsObject;
-        const box = physicsMesh.geometry.boundingBox.clone();
-        box.getCenter = () => app.position; //set box center to app position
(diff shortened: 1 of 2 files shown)


object-spriter.js
```diff
@@ -112,7 +112,10 @@ const createObjectSpriteInternal = (app, {
       if (physicsObjects.length > 0) {
         const physicsObject = physicsObjects[0];
         const {physicsMesh} = physicsObject;
-        fitCameraToBoundingBox(sideCamera, physicsMesh.geometry.boundingBox, 1.2);
+        const box = physicsMesh.geometry.boundingBox.clone();
+        box.getCenter = () => app.position; //set box center to app position
```
avaer, line 116: No, this isn't the way to do it. The box data should be fixed, not hacked,

avaer (changes requested):
This is the wrong approach.

PR #3510, opened 2022-08-05
Sit limiter gameplay

aryasalehi:
## Describe your changes
This PR disables jump and fly when the character is sitting

## What are the steps for a QA tester to test this pull request?
When you're sitting on a vehicle try jumping or flying ( you shouldn't be able to fly or jump when you're sitting )

## Issue ticket number and link
https://github.com/webaverse/app/issues/3503

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [x] I am not adding any irrelevant code or assets
- [x] I am only including the changes needed to implement the change
- [x] I have playtested and intentionally tried to find error cases but couldn't

PR #3524, opened 2022-08-06
Fix #3503

### LINKED ISSUE #3503: When you press fly while sitting in vehicle makes the vehicle's physics unavailable
- Sit in the vehicle
- Press `f` to fly
- Press tilde to find the vehicle's collider

aryasalehi:
## Describe your changes
This is the fix for : https://github.com/webaverse/app/issues/3503
related : https://github.com/webaverse/app/pull/3510

The reason that the previous PR wasn't merged is because we want to have the ability to fly or jump out of a mount for example, but we don't want that behavior for a hoverboard
So we're going to add a a property in the metaversefile which disables/enables jump or fly for a vehicle 

Same steps as the issue

https://github.com/webaverse/app/issues/3503

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -749,6 +749,11 @@ class StatePlayer extends PlayerBase {
       this.characterPhysics.setPosition(position);
     }
   }
+  getActionsWithType(type) {
+   const actions = this.getActionsState(); 
+   const typedActions = Array.from(actions).filter(action => action.type === type);
+   return typedActions;
+  }
   getActions() {
     return this.getActionsState();
   }
--- a/character-controller.js
+++ b/character-controller.js
@@ -749,7 +749,7 @@ class StatePlayer extends PlayerBase {
       this.characterPhysics.setPosition(position);
     }
   }
-  getActionsWithType(type) {
+  getActionsByType(type) {
    const actions = this.getActionsState(); 
    const typedActions = Array.from(actions).filter(action => action.type === type);
    return typedActions;
--- a/game.js
+++ b/game.js
@@ -67,6 +67,18 @@ const _getGrabbedObject = i => {
   return result;
 };
 
+const _unwearAppIfHasSitComponent = (localPlayer) => {
+  const wearActions = Array.from(localPlayer.getActionsState()).filter(action => action.type === 'wear');
+    for (const wearAction of wearActions) {
+      const instanceId = wearAction.instanceId;
+      const app = metaversefileApi.getAppByInstanceId(instanceId);
+      const sitComponent = app.getComponent('sit');
+      if (sitComponent) {
+        app.unwear();
+      }
+    }
+}
+
 // returns whether we actually snapped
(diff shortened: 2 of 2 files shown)


game.js
```diff
@@ -67,6 +67,18 @@ const _getGrabbedObject = i => {
   return result;
 };
 
+const _unwearAppIfHasSitComponent = (localPlayer) => {
+  const wearActions = Array.from(localPlayer.getActionsState()).filter(action => action.type === 'wear');
```
avaer, line 71: Why not `player.hasAction('wear')`?

aryasalehi: Makes sense , I'll use that 👍

game.js
```diff
@@ -1398,6 +1410,9 @@ class GameManager extends EventTarget {
         type: 'fly',
         time: 0,
       };
+      if (localPlayer.hasAction('sit')) {
+       _unwearAppIfHasSitComponent(localPlayer);
```
avaer, line 1414: Spacing.

game.js
```diff
@@ -67,6 +67,18 @@ const _getGrabbedObject = i => {
   return result;
 };
 
+const _unwearAppIfHasSitComponent = (localPlayer) => {
```
avaer, line 70: Argument should be `player`.

game.js
```diff
@@ -1398,6 +1410,9 @@ class GameManager extends EventTarget {
         type: 'fly',
         time: 0,
       };
+      if (localPlayer.hasAction('sit')) {
```
avaer, line 1413: This check seems pointless?

avaer (commented):
Minor code cleanup, otherwise looks good.

game.js
aryasalehi: It's actually player.getAction('wear')

game.js
```diff
@@ -67,6 +67,16 @@ const _getGrabbedObject = i => {
   return result;
 };
 
+const _unwearAppIfHasSitComponent = (player) => {
+  const wearAction = player.getAction('wear');
```
avaer, line 71: Is this only handling the first wear action? Is that correct?

game.js
```diff
@@ -67,6 +67,16 @@ const _getGrabbedObject = i => {
...
+  const instanceId = wearAction.instanceId;
+  const app = metaversefileApi.getAppByInstanceId(instanceId);
+  const sitComponent = app.getComponent('sit');
```
avaer, line 74: This could be `hasComponent` but it doesn't matter much.

character-controller.js
```diff
@@ -749,6 +749,11 @@ class StatePlayer extends PlayerBase {
       this.characterPhysics.setPosition(position);
     }
   }
+  getActionsWithType(type) {
```
avaer, line 752: I think `getActionsByType` might be more in line with current naming schemes?

avaer (approved):

Issue #3508, opened 2022-08-05
Sword attacking itself if dropped while being used and crashing the app


alisaad673:
**Describe the bug**
While we are using the sword, in case we drop it while using it, and the left click is kept pressed, it would attack itself even though the sword is dropped already. Also when the sword vanishes, the app crashes and throws this error. This happens only if we keep standing closer to the sword.

**To Reproduce**
1. Equip a sword/silsword
2. use it by pressing left click
3. drop it with 'r' while being used, keeping the 'left click' pressed.
4. wait till the sword vanishes
5. the app would crash

**Expected behavior**
sword action should end immediately after its dropped. 

**Screenshots**
![image](https://user-images.githubusercontent.com/35926530/183125926-1948b7ab-b2d4-4a15-9d8c-b9d1e9b382a1.png)

https://user-images.githubusercontent.com/35926530/183125962-09ea0187-701f-46f0-8c9f-f0ac389c6e4d.mp4

alisaad673:
FIxed

PR #3526, opened 2022-08-06
dispatch use event false when dropping any app

memelotsqui:
## Describe your changes
-Dispatch event 'use' set to false when dropping the app

1- Grab a sword
2- Click and hold to use it
3- Release with "r" while still using it
4- You shouldnt be hitting the sword app now

#3508 

## Screenshots and/or video

https://user-images.githubusercontent.com/1117257/183260008-55e44df0-9e49-4ebf-9718-e9d9c896829c.mp4

### DIFF
--- a/game.js
+++ b/game.js
@@ -1238,6 +1238,11 @@ class GameManager extends EventTarget {
           .add(localVector2.set(0, 0.5, -1).applyQuaternion(localPlayer.quaternion)),
         dropDirection: zeroVector, */
       });
+      app.dispatchEvent({
+        type: 'use',
+        use: false,
+      });
+      localPlayer.removeAction('use');
     }
   }
   deleteSelectedApp() {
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -1238,6 +1238,11 @@ class GameManager extends EventTarget {
           .add(localVector2.set(0, 0.5, -1).applyQuaternion(localPlayer.quaternion)),
         dropDirection: zeroVector, */
       });
+      app.dispatchEvent({
```
avaer, line 1241: This looks almost correct; we should only "unuse" the app if it was being used. That code can be abstracted instead of trying to manage use event emits in the `dropSelectedApp` method.

avaer (changes requested):
Recommending refactor.

game.js
memelotsqui: Alright, I will use _endUse() instread, seems that function was already validating if the app was being used.

alisaad673:
This is working fine now.

Issue #3267, opened 2022-07-16
switching scene breaks because of NPC


lalalune:
When leaving main scene, switching breaks. This is because of the stats panel.

When the NPC is destroyed it should unwear all apps with it -- currently not happening.

![image](https://user-images.githubusercontent.com/18633264/179335455-46d804a3-39c4-4b22-be3e-a9d5e8c4212c.png)

ultrabot2022:
This issue was fixed by https://github.com/webaverse/app/pull/3346

PR #3540, opened 2022-08-07
Fix character stats panel and add back to main scene NPC

lalalune:
## Describe your changes
There was a bug where the character stats panel would not remove with the NPC and would throw errors when transitioning scenes. This has been fixed.

1. Go to main scene
2. Then switch to other scene. If it's broken you'll see something like in this issue: https://github.com/webaverse/app/issues/3267
3. If it works, you'll have no errors and scene transition will be smooth

https://github.com/webaverse/app/issues/3267
Builds on this closed PR: https://github.com/webaverse/app/pull/3346


dom-renderer.jsx
```diff
@@ -116,6 +116,8 @@ class DomItem extends THREE.Object3D {
   destroy() {
     physicsScene.enableActor(this.physicsObject);
     physicsScene.removeGeometry(this.physicsObject);
+    this.remove(this.floatNode);
```
avaer, line 119: This line doesn't seem needed, since this Object3D is not valid for use in the scene after calling `destroy()` anyway.

character-controller.js
```diff
@@ -867,11 +867,12 @@ class StatePlayer extends PlayerBase {
     });
   }
   destroy() {
-    this.unbindState();
     this.appManager.unbindState();
 
     this.appManager.destroy();
-  
+    
+    this.unbindState();
```
avaer, line 874: Can you make a detailed comment about why this needs to be done in this order? I think in general the unbind should happen first so that none of the destroy methods are changing app state during destruction.

avaer (changes requested):
Recommending refactors.

rondoor124:
https://user-images.githubusercontent.com/92342281/188236186-d8a8c297-b3b9-4b31-827d-a5cda669fc5f.mp4
I found an issue when testing this PR.

Issue #3529, opened 2022-08-06
Crouching while in midair


avaer:
It is currently possible to crouch while jumping (you can tell by the camera going down). It seems to reset when you land.

We should either make crouch work in midair or disable setting such an action while jumping.

https://user-images.githubusercontent.com/6926057/183265541-cf53db0b-9028-4d45-a5ee-c377e7c7e241.mp4

alisaad673:
Fixed

PR #3543, opened 2022-08-07
disable crouch if player is not on the ground

tcm390:
In this PR, disable crouch when player is not on the ground

related:
https://github.com/webaverse/app/issues/3529

https://user-images.githubusercontent.com/60634884/183292509-0ed675ee-46be-4776-8264-9a56855ff2d3.mp4

### DIFF
--- a/io-manager.js
+++ b/io-manager.js
@@ -245,12 +245,13 @@ const _updateIo = timeDiff => {
       keysDirection.applyQuaternion(camera.quaternion);
       _updateVertical(keysDirection);
     } else {
+      const isGrounded = localPlayer.characterPhysics.lastGrounded;
       const cameraEuler = camera.rotation.clone();
       cameraEuler.x = 0;
       cameraEuler.z = 0;
       keysDirection.applyEuler(cameraEuler);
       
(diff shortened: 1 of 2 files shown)


io-manager.js
```diff
@@ -245,12 +245,13 @@ const _updateIo = timeDiff => {
       keysDirection.applyQuaternion(camera.quaternion);
       _updateVertical(keysDirection);
     } else {
+      const isGrounded = localPlayer.characterPhysics.lastGrounded;
```
avaer, line 248: We should not re-use tracking flags like this. If you want to track if the character is grounded, we should have an API for that.

avaer (changes requested):
Recomending API refactor.

io-manager.js
tcm390: [add isGrounded API in game.js](https://github.com/webaverse/app/pull/3543/commits/f7db09a71021d68980c8022850eaf722daf85f3a)

alisaad673:
Working fine.

avaer (approved):

Issue #3531, opened 2022-08-06
Lost sound effects on equip/drop


avaer:
The sound effects for equip/drop no longer play.

PR #3545, opened 2022-08-07
lost equip sound bug

tcm390:
This issue is caused by that `isLocalPlayer` hasn't been set in character-sfx constructor (undefined)

related:
https://github.com/webaverse/app/issues/3531

result:

https://user-images.githubusercontent.com/60634884/183294177-88036462-56e6-430d-adb5-e37f0f6d7f66.mp4

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -157,7 +157,6 @@ class PlayerBase extends THREE.Object3D {
     this.name = defaultPlayerName;
     this.bio = defaultPlayerBio;
     this.characterHups = new CharacterHups(this);
-    this.characterSfx = new CharacterSfx(this);
     this.characterFx = new CharacterFx(this);
     this.characterHitter = new CharacterHitter(this);
     this.characterBehavior = new CharacterBehavior(this);
@@ -1040,7 +1039,8 @@ class LocalPlayer extends UninterpolatedPlayer {
     this.isLocalPlayer = !opts.npc;
     this.isNpcPlayer = !!opts.npc;
     this.detached = !!opts.detached;
-
+    
+    this.characterSfx = new CharacterSfx(this);
     this.characterPhysics = new CharacterPhysics(this);
   }
   async setPlayerSpec(playerSpec) {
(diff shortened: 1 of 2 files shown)


character-sfx.js
```diff
@@ -86,24 +86,25 @@ class CharacterSfx {
...
     this.currentStep = null;
     this.currentSwimmingHand = null;
     this.setSwimmingHand = true;
+    this.setEquipSound = false;
   }
   update(timestamp, timeDiffS) {
     if (!this.player.avatar) {
       return;
     }
+    if (this.player.isLocalPlayer && !this.setEquipSound) {
```
avaer, line 98: If this is a listener then it should not be listened in the main loop. This belongs in setup code.

avaer (changes requested):
Code in the wrong place.

tcm390:
https://github.com/webaverse/app/pull/3545/files

create character-sfx for localPlayer after `isLocalPlayer` has been set

character-sfx.js
tcm390: https://github.com/webaverse/app/pull/3545/files

create characterSfx for localPlayer after `isLocalPlayer` has been set (like characterPhysics)

avaer:
How about we instead ensure that isLocalPlayer is initialized early enough? it could be an argument to `PlayerBase` `constructor`.

avaer (changes requested):
Recommending refactor.

alisaad673:
Works Fine

character-controller.js
```diff
@@ -1040,7 +1039,8 @@ class LocalPlayer extends UninterpolatedPlayer {
     this.isLocalPlayer = !opts.npc;
     this.isNpcPlayer = !!opts.npc;
     this.detached = !!opts.detached;
-
+    
+    this.characterSfx = new CharacterSfx(this);
```
avaer, line 1043: Both of these should be put in PlayerBase.

avaer (changes requested):
Still have initialization code in the wrong classes.

character-controller.js
avaer: Especially the character sfx, since that should affect remote players too.

PR #3549, opened 2022-08-07
Adding actionability for "jump" action in vehicles

aryasalehi:
## Describe your changes
This PR adds the option to enable "jump" action on vehicles ( sittable apps )
"jumpable" component can be added to the .metaversefile of vehicle apps
```   
{
  "key" : "jumpable",
  "value" : true
},
```
The default value of this component is false

**if you don't add this component/set it to false you can not** jump when you're sitting on a vehicle, if you set "jumpable" to "true" : you should be able to jump when you're sitting on a vehicle and that automatically drops the vehicle

Add the component that is mentioned above to a vehicle's metaversfile and set it to true/false and test to see if the gameplay matches.

Related : https://github.com/webaverse/app/pull/3524

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -397,17 +397,14 @@ class PlayerBase extends THREE.Object3D {
       };
       _transplantNewApp();
 
-      const _disableAppPhysics = () => {
-        // don't disable physics if the app is a pet
-        if (!app.hasComponent('pet')) {
-          const physicsObjects = app.getPhysicsObjects();
-          for (const physicsObject of physicsObjects) {
-            physicsScene.disableGeometryQueries(physicsObject);
-            physicsScene.disableGeometry(physicsObject);
-          }
+      const _initPhysics = () => {
+        const physicsObjects = app.getPhysicsObjects();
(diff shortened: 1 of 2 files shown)


avaer:
If these components are mutually exclusive, shouldn't it actually be one component like `jumpBehavior`?

alisaad673:
Works fine. Also we should add this to the vehicles(mount, dragon) repos readme, in order to document it.

avaer (changes requested):
Can we limit this PR to only jump action?

I think vehicle fly is a lot more complicated (helicopter vs plane vs flyable mount), and it would also be very different from the current F god mode fly hack.

alisaad673:
Its working fine on the jump action. Since we are not catering the fly action here, pressing 'f' drops the vehicle

lalalune:
Passing to @memelotsqui to review as part of actionsManager feature

rondoor124:
This PR is waiting for @aryasalehi and @memelotsqui's review as part of actionsManager feature

Issue #3509, opened 2022-08-05
Disable using swords/bow while dancing


alisaad673:
**Describe the bug**
Using sword/silsword/bow while dancing makes no sense and the animation is weird too. So, we should disable it

**To Reproduce**
1. Equip sword/silsword/bow
2. press 'v' to dance
3. use the weapon 

**Expected behavior**
Since we don't have proper animation for using weapons while dancing, we should disable using these weapons while dancing.

**Screenshots**

https://user-images.githubusercontent.com/35926530/183127829-55cf0b02-d232-4925-b9b5-4a30ece49979.mp4

https://user-images.githubusercontent.com/35926530/183127860-b29d0a08-4167-4810-96d6-b2405acf159a.mp4

PR #3551, opened 2022-08-07
validate no use action before dance action

memelotsqui:
## Describe your changes
1- Check if objects are not currently being used to trigger 'dance' animation when user press and holds 'v' key
2- Also, In case player is dancing and presses mouse button, it will give more importance to use action (it will stop dancing and switch to use action)

1- Equip any weapon
2- Press and hold 'v' to dance
3- Click with mouse
= Player stops dancing and uses the objects he is holding

1- Equip any weapon
2- Click with mouse and hold
3- Press and hold 'v'
= Player will continue doing the object action, giving priority over dance action

#3509 

## Screenshots and/or video

https://user-images.githubusercontent.com/1117257/183310840-886af42f-aca2-4f0f-80c3-fcb6fb12fd96.mp4

https://user-images.githubusercontent.com/1117257/183310837-af6439b9-71df-4117-aa8b-273144bd7ab3.mp4

### DIFF
--- a/game.js
+++ b/game.js
@@ -387,6 +387,7 @@ const _startUse = () => {
     const useComponent = wearApp.getComponent('use');
     if (useComponent) {
       const localPlayer = playersManager.getLocalPlayer();
+      localPlayer.removeAction('dance');
       const useAction = localPlayer.getAction('use');
       if (!useAction) {
         const {instanceId} = wearApp;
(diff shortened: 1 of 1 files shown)


game.js
```diff
@@ -1293,13 +1294,16 @@ class GameManager extends EventTarget {
     if (_getGrabbedObject(0)) {
       this.menuGridSnap();
     } else {
-      localPlayer.removeAction('dance');
+      const useAction = localPlayer.getAction('use');
+      if (!useAction){
```
avaer, line 1298: Spacing.

avaer (changes requested):
This is not a good actions design.

I think we need to create an actual system that intelligently lists which actions are available and which ones cannot coincide, with an automated cleanup.

Additionally, because the dance is a hot key that has not been hooked up to any gameplay action, it is not an important bug to spend time on at the moment.

This will be fixed with a proper actions system refactor.

lalalune:
Notes from Avaer:
Needs a list of actions clearly defined
A class which is used to add them and de-interfere them

avaer:
I don’t think we need this PR though. The proper fix would be entirely unrelated.

Issue #3388, opened 2022-07-28
Formalize same UI for 'inventory' and 'From upstreet' items


alisaad673:
We need to formalize and agree upon a single UI for all items in both "inventory" panel and " Season: From upstreet" panel.

![image](https://user-images.githubusercontent.com/35926530/181499060-0b1fa6d3-c930-4948-9e49-f00b7942eec7.png)
![image](https://user-images.githubusercontent.com/35926530/181499168-c21366b4-87fe-49af-9a54-a0334063b49b.png)

PR #3567, opened 2022-08-09
Formalize same UI for 'inventory' and 'From upstreet' items

mavisakalyan:
## Describe your changes
- made both lists in the inventory and upstream to look the same have similar structure so we are consistent with the design. Only difference is the highlight.
- find some items to add to your claims list and check both lists, inventory and from upstream in the account tab. Both lists should be the same in structure with color differance.
https://github.com/webaverse/app/issues/3388
## Screenshots and/or video
Screen Shot 2022-08-09 at 19 51 06
Screen Shot 2022-08-09 at 19 51 16

### DIFF
--- a/src/components/general/equipment/equipment.module.css
+++ b/src/components/general/equipment/equipment.module.css
@@ -154,6 +154,9 @@
   flex-direction: column;
   cursor: pointer;
   user-select: none;
+  margin: 0 15px 15px 0;
+  width: 100px;
+  background-color: #333333;
 }
 .object .background,
 .object .highlight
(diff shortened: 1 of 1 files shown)


avaer:
The orange highlight is not good and was never intended as an actual design. We can do a lot better with e.g. circles or background gradients to indicate categories. The only reason it's like this is it's something I did in 5 minutes a year ago.

mavisakalyan:
Got it let me see what I can come up with.

mavisakalyan:
One question though, should the hover effect be different in color for different tabs? or should we keep one consistent design? @avaer

lalalune:
@mavisakalyan don't worry about hover for now

Issue #3565, opened 2022-08-09
Sfx issue for switched npc('g' key)


alisaad673:
**Describe the bug**
-The switched npc doesn't have the footsteps sfx while its walking
-Walking sfx plays while its running

**To Reproduce**
1. Run the app
2. Press 'g' key
3. Walk and run

**Expected behavior**
walking sfx should play when npc is walking and running sfx should play when its running

**Screenshots**

https://user-images.githubusercontent.com/35926530/183646350-a5421f46-730b-4ba6-8bd2-87194ab5e2ba.mp4

tcm390:
https://github.com/webaverse/app/pull/3422

The issue would solved by this PR

tcm390:
related: https://github.com/webaverse/app/issues/3568

Issue #3397, opened 2022-07-28
Issues with switched NPC( 'g' key)


alisaad673:
**while we press 'G' key and switch to the nearest NPC, I noticed following issues:**

### 1- "E" activator keeps appearing over the npc, while we switch to it.

https://user-images.githubusercontent.com/35926530/181524059-7faf0dfe-f149-4e68-99da-e4274c3296b4.mp4

### 2-The jump animation of switched npc is too fast. 

https://user-images.githubusercontent.com/35926530/181524248-56d890e5-38d9-41af-bcc1-7bd4f96796e7.mp4

### 3-While we press "E" on the switched NPC, it restricts the movement of npc inside a kind of loop around the avatar.

https://user-images.githubusercontent.com/35926530/181524680-aee55c9a-835a-4688-9b9b-28eb6f376b06.mp4

### 4-While we switch to npc from avatar, the avatar is left just in the state it was switched from.

https://user-images.githubusercontent.com/35926530/181525036-60f806e2-1289-4712-bde9-27ec00aab496.mp4

avaer:
I think the bug is that avatar updates for swithced-to NPCs are run twice.

jakezira:
## Problems

- Local player isn't based on `App` ==> It's missing collider capsule.
- When there's local player switch, it needs to disable collider because avatar's not interacting with himself.

jakezira:
- When npc joins the party, it should be moved to avatar's app manager from scene app manager.
- Physics tracker needs to track all app manager. Actually, manager for physics tracking is preferable ==> https://github.com/webaverse/app/issues/3614

jakezira:
## Party System

- https://github.com/webaverse/app/pull/3625
- https://github.com/webaverse/totum/pull/130

- Video: https://www.loom.com/share/262313dc551c4e84b44d6a8d3d0530cc

PR #3569, opened 2022-08-09
npc update twice bug

tcm390:
related: 
https://github.com/webaverse/app/issues/3565
https://github.com/webaverse/app/issues/3397

issue:

https://user-images.githubusercontent.com/60634884/183740945-97315a59-3825-495a-9ee1-5f5bc14533af.mp4

Result:

https://user-images.githubusercontent.com/60634884/183775298-59ef6e2c-85e8-4079-b8d3-7b4486bd98da.mp4

### DIFF
--- a/npc-manager.js
+++ b/npc-manager.js
@@ -66,7 +66,7 @@ class NpcManager extends EventTarget {
     if (!detached) {
       this.npcs.push(npcPlayer);
     }
-
+    
     return npcPlayer;
   }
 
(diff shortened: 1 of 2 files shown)


alisaad673:
The previous issues seem fixed. 
Just these Issues:

1-Talking to the npc is broken, throws this error:
image

https://user-images.githubusercontent.com/35926530/184180516-75e2433b-7f1f-4e35-9167-a27f0a88cf08.mp4

2-If we are in sonic boom and switch to npc the main character stays in the state of naruto run, same if we are in naruto run with npc and we switch to main player, the npc remains in the state of naruto run(standing)

https://user-images.githubusercontent.com/35926530/184180894-ae66e311-b74a-4d83-a021-82df4b43905b.mp4

npc-manager.js
```diff
@@ -142,58 +188,18 @@ class NpcManager extends EventTarget {
...
-            /* if (isNaN(npcPlayer.position.x)) {
-              debugger;
-            } */
-            npcPlayer.updateAvatar(timestamp, timeDiff);
-            /* if (isNaN(npcPlayer.position.x)) {
-              debugger;
-            } */
-          }
-        };
-        world.appManager.addEventListener('frame', frame);
```
avaer, line 195: Because this was moved I think the npc will crash when deleted from the scene.

avaer (changes requested):
The whole approach needs a rethink.

PR #3576, opened 2022-08-10
Bump totum - fix GLB animations

avaer:
Intake https://github.com/webaverse/totum/pull/125


avaer:
This update also fixes HTML web page rendering (`classroom.scn`).

PR #3409, opened 2022-07-29
Avatar optimizer 2 [WIP]

avaer:
Replaces https://github.com/webaverse/app/pull/2805

PR #3577, opened 2022-08-10
Avatar optimizer 3

avaer:
Replaces https://github.com/webaverse/app/pull/3409.

Depends on:
- THREE.js r143 update (based here).
- https://github.com/webaverse/totum/pull/127


avaer:
Ok, this should be done. All 3 avatar optimizers produce correct results, and are running off the main thread now.

avaer:
This was merged in the labs!

PR #3591, opened 2022-08-11
Pet character physics ( WIP )

aryasalehi:
This PR is an attempt to add physx character physics to pets

### DIFF
--- a/metaverse_components/pet.js
+++ b/metaverse_components/pet.js
@@ -11,13 +11,9 @@ const localQuaternion3 = new THREE.Quaternion();
 const localEuler = new THREE.Euler();
 const localMatrix = new THREE.Matrix4();
 
-const downQuaternion = new THREE.Quaternion(
-  0.7071068, 0, 0, 0.7071068
-);
+const downQuaternion = new THREE.Quaternion(-0.7071067811865476, 0, 0, 0.7071067811865476);
 
-let lastZDirection = 0;
-
-const physicsIds = [];
+const lastDirection = new THREE.Vector3();
 
 export default (app, component) => {
   const {
@@ -32,7 +28,6 @@ export default (app, component) => {
 
   // app.rotation.y += Math.PI / 2;
   const physics = usePhysics();
-
   const petHeight = 0.01;
   const petWidth = 1.04;
 
@@ -52,28 +47,10 @@ export default (app, component) => {
     stepOffset,
     app.position
   );
+  // const debugCapsuleMesh = characterController.children[0];
+  // debugCapsuleMesh.visible = true;
 
-  // const capsuleGeometry = new CapsuleGeometry(
-  //   radius - contactOffset,
-  //   radius - contactOffset,
-  //   capsuleHeight,
-  //   contactOffset,
-  //   stepOffset
-  // );
-  // const debugCapsuleMesh = new THREE.Mesh(
-  //   capsuleGeometry,
-  //   new THREE.MeshBasicMaterial()
-  // );
-
-  const debugCapsuleMesh = characterController.children[0];
-  debugCapsuleMesh.visible = true;
-
-  app.add(debugCapsuleMesh);
-
-  // const debugCapsuleId = physics.addGeometry(debugCapsuleMesh);
-  // physicsIds.push(debugCapsuleId);
-
-  // const physicsObject = physicsScene.addCapsuleGeometry(app.position, app.quaternion, 0.2, 0.05);
+  // app.add(debugCapsuleMesh);
 
   let petSpec = null;
   let petMixer = null;
@@ -174,7 +151,6 @@ export default (app, component) => {
     if (collisionToGround) {
       position.y = collisionToGround.point[1];
     }
-    // position.y = 0;
 
     const distance = app.position.distanceTo(position);
     return { distance, position };
@@ -217,12 +193,13 @@ export default (app, component) => {
       fallLoopStartTimeS = 0;
       startFallingPosition.set(0, 0, 0);
       lastGravityH = 0;
-    } 
+    }
 
     // components
     const _updateAnimation = () => {
       // making sure the nose of the pet hits the physics object in front of it
-      app.position.z -= (petWidth / 2) * lastZDirection;
+      app.position.x -= (petWidth / 2) * lastDirection.x;
+      app.position.z -= (petWidth / 2) * lastDirection.z;
 
       if (petComponent) {
         if (rootBone) {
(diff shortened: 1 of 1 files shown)


aryasalehi:
Fixed the y positioning for direction.
Now we're able to use pets with different ground positions ( before it only worked with y=0 grounds )
Though the quaternion when you walk up a ramp is not perfect ; I will fix that next

aryasalehi:
Fixed the physics raycast quaternion and added a height follow limiter ( for y )
( HEIGHT_FOLLOW_LIMIT_Y is used for limiting the difference between the players position ( with the new Y ) and the pet Y position )

**This is ready for QA/Testing**

Uploading 2022-08-13 19-41-17.mp4…

metaverse_components/pet.js
```diff
@@ -9,8 +11,47 @@ const localQuaternion3 = new THREE.Quaternion();
 const localEuler = new THREE.Euler();
 const localMatrix = new THREE.Matrix4();
 
+const downQuaternion = new THREE.Quaternion(-0.7071067811865476, 0, 0, 0.7071067811865476);
+
+const lastDirection = new THREE.Vector3();
+
+const HEIGHT_FOLLOW_LIMIT_Y = 0.9;
+const PET_WIDTH = 1.04;
```
avaer, line 19: This constant needs to be declared or computed on the pet metaversefile components. It should not be hardcoded.

metaverse_components/pet.js
```diff
@@ -9,8 +11,47 @@ const localQuaternion3 = new THREE.Quaternion();
 const localEuler = new THREE.Euler();
 const localMatrix = new THREE.Matrix4();
 
+const downQuaternion = new THREE.Quaternion(-0.7071067811865476, 0, 0, 0.7071067811865476);
+
+const lastDirection = new THREE.Vector3();
+
+const HEIGHT_FOLLOW_LIMIT_Y = 0.9;
```
avaer, line 18: This constant should be removed.

metaverse_components/pet.js
```diff
@@ -9,8 +11,47 @@ const localQuaternion3 = new THREE.Quaternion();
...
+    usePhysics,
+    useCleanup,
+    useLocalPlayer,
+    useRemotePlayers,
+    getPlayerByAppInstanceId,
+    useActivate,
+  } = metaversefile;
+
+  const physics = usePhysics();
+  const petCapsuleHeight = 0.01;
```
avaer, line 33: Constant to remove.

metaverse_components/pet.js
```diff
@@ -9,8 +11,47 @@ const localQuaternion3 = new THREE.Quaternion();
...
+
+  const physicsScene = physicsManager.getScene();
+  const heightFactor = 1;
+  const baseRadius = 0.2;
+  const radius = (baseRadius / heightFactor) * petCapsuleHeight;
+  const capsuleHeight = petCapsuleHeight - radius * 2;
```
avaer, line 39: All of these constants should be deleted.

metaverse_components/pet.js
```diff
@@ -90,45 +141,106 @@ export default (app, component) => {
...
     }
   });
-  
+
   const smoothVelocity = new THREE.Vector3();
   const lastLookQuaternion = new THREE.Quaternion();
   const _getAppDistance = () => {
-    const position = localVector.copy(player.position);
-    position.y = 0;
+    const position = player.position.clone();
```
avaer, line 148: This is creating garbage during the frame loop.

metaverse_components/pet.js
```diff
@@ -90,45 +141,106 @@ export default (app, component) => {
...
+
+    // setting the height of the y as the height of the ground
+    const collisionToGround = physics.raycast(position, downQuaternion);
+    if (collisionToGround) {
+      position.y = collisionToGround.point[1];
+    }
+
     const distance = app.position.distanceTo(position);
-    return distance;
+    return { distance, position };
```
avaer, line 157: Garbage creation during the frame loop.

metaverse_components/pet.js
```diff
@@ -90,45 +141,106 @@ export default (app, component) => {
...
+      fallLoopStartTimeS = nowS;
+      startFallingPosition.copy(app.position);
+      localVector.set(0, 0, 0);
+    }
+    const t = nowS - fallLoopStartTimeS;
+    const h = 0.5 * physicsScene.getGravity().y * t * t;
+
+    const displacement = localVector.set(0, h - lastGravityH, 0);
+    smoothVelocity.add(displacement);
+    const flags = physicsScene.moveCharacterController(
```
avaer, line 182: This seems a duplication of character movement code, so should be charged with npc characters.

metaverse_components/pet.js
```diff
@@ -90,45 +141,106 @@ export default (app, component) => {
...
+
+              const direction = diff.normalize();
               const maxMoveDistance = distance - minDistance;
               const moveDistance = Math.min(speed * timeDiff, maxMoveDistance);
-              moveDelta.copy(direction)
-                .multiplyScalar(moveDistance);
+              moveDelta.copy(direction).multiplyScalar(moveDistance);
               app.position.add(moveDelta);
-              app.quaternion.slerp(localQuaternion.setFromUnitVectors(localVector2.set(0, 0, 1), direction), 0.1);
+              app.quaternion.slerp(
```
avaer, line 234: There is no need to slerp anything here. This should be resolved by the abstracted shared character controller, based on follow and look settings on it. Quaternions should not be computed manually here.

avaer (changes requested):
It’s a good idea to move the pet to character controllers but I don’t think this is the way.

This code is confused between the previous pet hack and actual character controller code, merging the problems of both.

Additionally there are many hardcoded constants here which will break for most pets.

Imo the maintainable + forward thinking way is for pets and npcs to share the character controller logic, settings, targets, and (soon) path finding 🙏. But there is no need to have a more buggy/less abstract version of character controlling being implemented in the pets component.

avaer:
Further, if we figure out the abstract character controller here, it can be a drop-in usage for mobs too, solving many mobs bugs in the process. And pets, characters, and mobs would all share intelligent capabilities we are developing here.

These all seem similar enough to be using the same controller code.

metaverse_components/pet.js
```diff
@@ -90,45 +141,106 @@ export default (app, component) => {
...
+
   const smoothVelocity = new THREE.Vector3();
   const lastLookQuaternion = new THREE.Quaternion();
   const _getAppDistance = () => {
-    const position = localVector.copy(player.position);
-    position.y = 0;
+    const position = player.position.clone();
+
+    // setting the height of the y as the height of the ground
+    const collisionToGround = physics.raycast(position, downQuaternion);
```
avaer, line 151: This seems like it can be entirely resolved by having correct character controller capsule and offset settings without any need for an external raycast.

metaverse_components/pet.js
aryasalehi: agree ❤

metaverse_components/pet.js
aryasalehi: This code is for finding the ground y position ( it's used for follow functionality )

aryasalehi:
I'll add a shared character controller class ( for avatars/pets/mobs ) and come back to this PR later

PR #3592, opened 2022-08-11
Fix bone chain accuracy

jakezira:


avatars/avatars.js
```diff
@@ -1252,6 +1252,14 @@ class Avatar {
       }
     }
     modelBones.Root.updateMatrixWorld();
+    // Offset skinned meshes to skeleton reference node's transform
+    const refObject = avatar.skeleton.bones[0].parent;
+    for (let mesh of avatar.skinnedMeshes) {
+      mesh.position.copy(refObject.position);
```
avaer, line 1258: Don't we want to use the parent `matrixWorld` decomposition rather than its local position?

Issue #2995, opened 2022-05-13
Unable to Equip Lamp Darkness Scene


msm0823:
I'm unable to equip lamp in the darkness scene. There is also no way of going near the lamp.

https://user-images.githubusercontent.com/104460803/168344798-5528cc13-8575-45cb-87c2-5dced5ad794e.mp4

alisaad673:
Related: https://github.com/webaverse/app/issues/2449

PR #3604, opened 2022-08-12
darkness bug

tcm390:
Sibling: https://github.com/webaverse/totum/pull/128

In darkness scene, we cannot equip the group object (lantern and fairy) because the physic is updated twice in app-manager (group app and subApp). So I add `isGroupComponent` for the subApp.
And In darkness scene, we could not equip the pet because the app is in the `world.appManager.apps` But we searched the app in `localPlayer.appManager.apps`.  https://github.com/webaverse/app/blob/ec8d75f24aed0787a518fbf0e665ced093deee93/metaversefile-api.js#L974

related: https://github.com/webaverse/app/issues/2995

result:

https://user-images.githubusercontent.com/60634884/184435203-2d3ef08e-7df6-42ae-87b2-3b2fe9bdf792.mp4

### DIFF
--- a/metaverse_components/pet.js
+++ b/metaverse_components/pet.js
@@ -68,8 +68,7 @@ export default (app, component) => {
   };
   app.addEventListener('wearupdate', e => {
     if (e.wear) {
-      // player = getPlayerByAppInstanceId(app.instanceId);
-      player = e.player;
+      player = e.player
       if (app.glb) {
         const {animations} = app.glb;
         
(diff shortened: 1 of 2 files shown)


metaverse_components/pet.js
```diff
@@ -68,7 +68,8 @@ export default (app, component) => {
   };
   app.addEventListener('wearupdate', e => {
     if (e.wear) {
-      player = getPlayerByAppInstanceId(app.instanceId);
+      // player = getPlayerByAppInstanceId(app.instanceId);
```
lalalune, line 71: just get rid of this line, we dont need to keep bad code unless it is useful

PR #3610, opened 2022-08-12
Fix avatar physx on switch

jakezira:
Reverted this: https://github.com/webaverse/app/commit/1213f06032a462267cb00c7115c9de9ae7d97b22?diff=split#r81065400 and fixed physics geometry query on avatar switch.


jakezira:
It became messy due to rebase. Make new PR from latest master again.

PR #3612, opened 2022-08-12
Fixed duplicated npc physics update

jakezira:


avaer:
@jakezira is this fixed in the party system?

jakezira:
> @jakezira is this fixed in the party system? If so, we can probably close it.

Yes, it's fixed.

zenkale:
Closing as merged in labs

PR #3621, opened 2022-08-14
revert avatars updates to velocity

lalalune:
## Describe your changes
Reverts some changes to accommodate multiplayer. Since we're changing velocity approach elsewhere we should revert this to not get entangled in separate bugs.

Characters should still run around smoothly, since this primarily affects single player avatar velocity. There should be no noticeable impact.


avaer:
PR was from the wrong repo, can you remake?

Issue #3566, opened 2022-08-09
Equipped fox won't disappear on being damaged by weapons


alisaad673:
**Describe the bug**
In case we have equipped fox and we try to damage it with weapons, it gets damaged but it wont disappear like other objects do.

**To Reproduce**
1. Go to shadows.scn
2. equip fox
3. Equip a word from character tab or equip a gun by switching scn to treehouse
4. attack the fox

**Expected behavior**
It should disappear after being fully damaged

**Screenshots**

https://user-images.githubusercontent.com/35926530/183665279-97f094db-d8d6-45ea-9086-2d15bcfa3732.mp4

patriboz:
Fix: https://github.com/webaverse/app/pull/3633

patriboz:
Updated Fix: https://github.com/webaverse/app/pull/3633

PR #3633, opened 2022-08-15
Fix pet destroy

patriboz:
Fix for https://github.com/webaverse/app/issues/3566

Added eventListener for die-events to pet apps within localPlayer.wear.

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -395,17 +395,6 @@ class PlayerBase extends THREE.Object3D {
       };
       _addAction();
 
-      const _enablePetDestroy = () => {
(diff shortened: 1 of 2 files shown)


avaer (changes requested):
Why is the character controller in charge of pet destruction?

It should be put here: https://github.com/webaverse/app/blob/master/metaverse_components/pet.js

patriboz:
> Why is the character controller in charge of pet destruction?
> 
> It should be put here: https://github.com/webaverse/app/blob/master/metaverse_components/pet.js

Done

avaer (approved):

PR #3635, opened 2022-08-15
Double jump bug

tcm390:
This issue is caused by that the `doubleJump` action is not removed while flying. In this PR, `doubleJump` is added to the controlActionTypes so that it would be removed while flying.

related: 
https://github.com/webaverse/app/issues/3639

result:

https://user-images.githubusercontent.com/60634884/184732739-b3aa0006-4c7e-4a95-b19a-cb76a2a03e2f.mp4

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -570,6 +570,7 @@ const controlActionTypes = [
   'fly',
   'sit',
   'swim',
+  'doubleJump'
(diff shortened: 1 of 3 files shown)


rondoor124:
I checked the code changes, it seems great to merge.

PR #3637, opened 2022-08-15
use avatar facing direction to set player quaternion

tcm390:
In character-physics.js, we use sitBone to set the player's quaternion. But I found that diorama and sonic boom would be broken if we sit on the dragon. I'm wondering if we could just use avatar facing direction to set player quaternion?

Issue: (hovercraft does not have problem but dragon would break the diorama and sonic boom)

https://user-images.githubusercontent.com/60634884/184726781-831d11ae-797a-415f-9185-d889c8b3f41a.mp4

Result:

https://user-images.githubusercontent.com/60634884/184726814-1566fef0-b9f6-4c36-8875-b095e664cbf3.mp4

### DIFF
--- a/character-physics.js
+++ b/character-physics.js
@@ -191,26 +191,25 @@ class CharacterPhysics {
         localVector2
       );
       localQuaternion.copy(this.character.quaternion);
+      // avatar facing direction
+      if (velocityAvatarDirection) {
+        const horizontalVelocity = localVector5.set(
+          this.velocity.x,
+          0,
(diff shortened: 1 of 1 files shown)


rondoor124:
I find one issue when the user rides the vehicle. 

https://user-images.githubusercontent.com/92342281/187803503-2b0372e3-09fe-4167-8894-95110a125758.mp4

Related: https://github.com/webaverse/app/pull/3647

Issue #3171, opened 2022-06-26
Add Rain


lalalune:
Pen is here:
![image](https://user-images.githubusercontent.com/18633264/175827777-f950ee3d-0bd1-492d-9564-ef73739c559d.png)

https://codepen.io/moonmoonmoon/pen/dymbvvW

lalalune:
@Loryhoof 
request from Uncanny Alley

PR #3650, opened 2022-08-17
rain effect

belopot:
## Describe your changes
Made "rain" effect in metaverse_modules and added it into prototype scene for preview

QA tester can see the rain effect in prototype scene

[#3171](https://github.com/webaverse/app/issues/3171)

## Screenshots and/or video
image

- [*] I have performed a self-review of my code
- [*] I am not adding any irrelevant code or assets
- [*] I am only including the changes needed to implement the change
- [*] I have playtested and intentionally tried to find error cases but couldn't

### DIFF
--- a/metaverse_modules/rain/.metaversefile
+++ b/metaverse_modules/rain/.metaversefile
@@ -1,0 +1,3 @@
+{
+  "start_url": "index.js"
+}
--- a/metaverse_modules/rain/index.js
+++ b/metaverse_modules/rain/index.js
@@ -1,0 +1,146 @@
+import * as THREE from "three";
+import metaversefile from "metaversefile";
+const { useApp, useFrame } = metaversefile;
+
+//Amount of drops
+const AMOUNT = 20000;
+//Falling speed of drops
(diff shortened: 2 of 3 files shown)


avaer (commented):
This is a good start, and it's technically sound, but:

- the effect could be much more intense (it doesn't need to look realistic, but it should not be tame)
hqdefault
- for artistic reasons we should make the terminal velocity closer to actual rain (https://www.google.com/search?q=erminal+velocity+of+rain)
- we should add ground patter
- to the last point, we will want to feed in character capsules + heightfield to the shader. the former can be added now, at least for the main character, and we can temporarily set the collision threshold for ground patter at y=0. in the procgen system we have a full heightfield texture, so it will be easy to replace that with a sample.

- finally, prototype.scn is deliberately plain for testing, so we should either create a new scene for this or simply move this rain to a more complex scene.

- not related now: these shaders should be part of a larger procgen weather system coinciding with the biomes.

avaer:
Looks like this won't be finished.

Issue #3645, opened 2022-08-17
Avatar grunts breaks speech phoneme recognition


avaer:
In `street.scn`, if you use any emote or the naruto run (sonic boom), after the character makes the grunt, the speech phoneme recognition (T to talk) is permanently broken -- it will glitch out into some sort of low volume zero-reset mode where the mouth barely moves.

PR #3651, opened 2022-08-17
Avatar grunts breaks speech phoneme recognition

tcm390:
The issue is caused by that both `audioWorker` and `microphoneWorker` are setting the value of `this.volume`

related:
https://github.com/webaverse/app/issues/3645

Result:

https://user-images.githubusercontent.com/60634884/185251426-803f8f20-16d5-4c41-9b26-8f746012d6e0.mp4

### DIFF
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -993,7 +993,7 @@ class Avatar {
     this.lastNeedsHeadTarget = false;
     this.lastHeadTargetTime = -Infinity;
 
-    this.manuallySetMouth=false;
+    this.disableAudioWorkerSetVolume=false;
   }
   static bindAvatar(object) {
     const model = object.scene;
@@ -2038,7 +2038,7 @@ class Avatar {
       const _volume = e => {
         // the mouth is manually overridden by the CharacterBehavior class which is attached to all players
         // this happens when a player is eating fruit or yelling while making an attack
-        if (!this.manuallySetMouth && this.isGrunting) {
+        if (!this.disableAudioWorkerSetVolume) {
           this.volume = e.data;
         }
       }
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -2038,7 +2038,7 @@ class Avatar {
       const _volume = e => {
         // the mouth is manually overridden by the CharacterBehavior class which is attached to all players
         // this happens when a player is eating fruit or yelling while making an attack
-        if (!this.manuallySetMouth) {
+        if (!this.manuallySetMouth && this.isGrunting) {
           this.volume = e.data;
         }
       }
--- a/avatars/avatars.js
+++ b/avatars/avatars.js
@@ -993,7 +993,7 @@ class Avatar {
     this.lastNeedsHeadTarget = false;
     this.lastHeadTargetTime = -Infinity;
 
-    this.disableAudioWorkerSetVolume = true;
+    this.enableAudioWorkerSetVolume = false;
   }
   static bindAvatar(object) {
     const model = object.scene;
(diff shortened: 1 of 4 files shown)


avaer (changes requested):
If volume goes from 1 -> 0 in one frame won't this cause it to get stuck? That's probably an even worse effect than the bug it is fixing.

How about choose which one to emit when it's enabled, not just by chance when it's zero?

tcm390:
[disable audio worker after grunt end](https://github.com/webaverse/app/pull/3651/commits/2d7e702001a61e38479d372ce4a6d6b00aae6787)

> If volume goes from 1 -> 0 in one frame won't this cause it to get stuck? That's probably an even worse effect than the bug it is fixing.
> 
> How about choose which one to emit when it's enabled, not just by chance when it's zero?

avaer:
> Can we disable audio worker after the grunt effect is completed?
No I don't think that's right, creating/destroying workers is a pretty heavy operation to run on every sound effect...

tcm390:
> > Can we disable audio worker after the grunt effect is completed?
> > No I don't think that's right, creating/destroying workers is a pretty heavy operation to run on every sound effect...

[disable audioWorker to set the volume when isGrunting is false](https://github.com/webaverse/app/pull/3651/commits/3413f9b23fd4fefccb5d280f6bd21b8282ec0517)

Got it. Is it okay to use `isGrunting` in avatar.js to decide whether `audioWorker` could set volume or not.

avaer:
> Got it. Is it okay to use isGrunting in avatar.js to decide whether audioWorker is allowed to set volume or not?

Probably not, because I think multiple grunts can happen at a time. At the very least I think this would need to be a stack number, not a boolean.

And if this is related to `manuallySetMouth`, shouldn't both be in the same class? Currently it's spread around so the code is confusing to read.

tcm390:
> And if this is related to `manuallySetMouth`, shouldn't both be in the same class? Currently it's spread around so the code is confusing to read.

[disable audioworker set volume after grunting](https://github.com/webaverse/app/pull/3651/commits/e495dbc63b728fe81e2ad3401b7378e1edd72d70)

Got it. Renamed `characterBehavior.manuallySetMouth` to `characterBehavior.disableAudioWorkerSetVolume`. And used `characterBehavior.disableAudioWorkerSetVolume` to disable audioWorker after emitting grunt.

avatars/avatars.js
```diff
@@ -993,7 +993,7 @@ class Avatar {
     this.lastNeedsHeadTarget = false;
     this.lastHeadTargetTime = -Infinity;
 
-    this.manuallySetMouth=false;
+    this.disableAudioWorkerSetVolume=false;
```
avaer, line 996: It is not good practice to have double negatives in variable names. The variable name should be the positive version, `enable`.

avaer, line 996: The default seems to be `true` when the grunt ends... so why is this false. Probably a consequence of choosing a poor variable name.

tcm390: Got it. [set disableAudioWorkerSetVolume initial value to true](https://github.com/webaverse/app/pull/3651/commits/cb447d56a76c188fa794dded6a04371269565571)

avaer:
It looks like this is still using the `disable` name: https://github.com/webaverse/app/pull/3651/files#diff-fcb2093ac11d9b04a179d3dedbef5c0aae4d14b7f71db90e483becca4170b7ccR2041

tcm390:
> It looks like this is still using the `disable` name: https://github.com/webaverse/app/pull/3651/files#diff-fcb2093ac11d9b04a179d3dedbef5c0aae4d14b7f71db90e483becca4170b7ccR2041

Sorry Changed it.
[change disableAudioWorkerSetVolume to enableAudioWorkerSetVolume](https://github.com/webaverse/app/pull/3651/commits/82d734e63e5371d383d5c3a6ea61399211339287)

rondoor124:
I have QA it and seems like work fine.

Issue #3652, opened 2022-08-18
Fix the format of file resulting from 'ctrl+s' to make it usable as scene files


zenkale:
**Describe the bug**
The file that results when we press to save a scene,"ctrl+s"  has a different format as compared to format of .scn files. That makes the resulting file unusable as a scene file.

**To Reproduce**
1. Go to any of the scenes
2. press ctrl+h
3. Change the extension of the .json file to .scn file
4. try to use it in webaverse by pasting in scenes folder

**Expected behavior**
The format should be same as the .scn files so that the exact scenes can be saved and be usable next time.

**Screenshots**

Format of file resulting from ctrl+s:

![image](https://user-images.githubusercontent.com/111510128/185467180-bac91079-fe3d-4b34-980e-627a2cbc88b6.png)

Format of our scn file:

![image](https://user-images.githubusercontent.com/111510128/185467296-5f2660f4-78fc-4650-91af-88acae4633a4.png)

avaer:
Completely agree, I think the Ctrl-S format is just out of date.

lalalune:
@avaer fixed
https://github.com/webaverse/app/pull/3653

PR #3653, opened 2022-08-18
Change ctrl s to download .scn file instead of json

lalalune:
## Describe your changes
Saves a .scn when user presses ctrl s, which can be easily re-uploaded. Requested by creators to speed up development workflow and use the app as an editor.

Press ctrl s
Scene should be saved
Verify that scene works by dragging back into window to load

https://github.com/webaverse/app/issues/3652

### DIFF
--- a/game.js
+++ b/game.js
@@ -1711,12 +1711,31 @@ class GameManager extends EventTarget {
     return localPlayer.setVoiceEndpoint(voiceId);
   }
   saveScene() {
-    const scene = world.appManager.exportJSON();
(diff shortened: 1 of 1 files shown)


avaer (commented):
This is overall good, but given that `appManager` is _already_ performing a formatting job, IMO it makes sense to move this formatting to `appManager.exportJSON`.

I don't think there are any other users of that method (it is not how we sync anything).

lalalune:
Closed in favor of Upstreet PR

PR #3257, opened 2022-07-15
Sword damage based on sword damageBox instead of localPlayer's position. ( 3rd try )

gonnavis:
Follow: https://github.com/webaverse/app/pull/2965 , https://github.com/webaverse/app/pull/3040

Sibling PRs: 
Add `damageBox` size/position infos to silsword: https://github.com/webaverse/silsword/pull/8
Add `vincibility` component to prevent sword's damageBox hit street: https://github.com/webaverse/street/pull/6

And fixed this regression issue https://github.com/webaverse/app/issues/3233

PR #3661, opened 2022-08-22
Melee Combat Improvement

patriboz:
This PR introduces a few major changes:

**Node based animations (avatars/animationHelpers.js)**
@gonnavis has implemented a WASM based animation system, which uses nodes. Instead of interpolating angles of joints on the three.js side, this system crossfades between nodes via physxworker.

**Melee combat: Weapon collision based on weapon hitbox (character-hitter.js & game.js)**
Using a weapon will now query collisions based on weapon hitbox instead of object distance. Combos are defined as a component of weapons and can then be performed with successive mouse clicks.

**Simple NPC combat AI (npc-manager.js)**
NPCs that wear a weapon will follow and attack/perform combos after activation.

[Video](https://youtu.be/pUWqaE64FwQ)


gonnavis:
Please make a standalone "player can get hit" pr to master. ( Maybe combined with "simple NPC combat AI" codes, because of if not, we should have no good way to test "player can get hit" feature, but the npc ai logics need keep very simple ).

patriboz:
> Please make a standalone "player can get hit" PR to master. ( Maybe combined with "simple NPC combat AI" codes, because of if not, we should have no good way to test "player can get hit" feature, but the npc ai logics need keep very simple, can improve it in another PR ).
> 
> `WASM animation` and `sword damageBox` are already made in other PRs #3160 , #3257, and `WASM animation` is still a work in progress.

Yes, that makes sense. I'll make a new PR with the standalone feature.

PR #3663, opened 2022-08-24
Patriboz/melee combat

patriboz:
This PR introduces the following features:

**Melee combat: LocalPlayer can get hit. Weapon collision based on weapon hitbox (character-hitter.js & game.js)**
Using a weapon will now query collisions based on weapon hitbox instead of object distance. Combos are defined as a component of weapons and can then be performed with successive mouse clicks.

**Simple NPC combat AI (npc-manager.js)**
NPCs that wear a weapon will follow and attack/perform combos after activation.

[Video](https://youtu.be/FT3A1zAJs98)


game.js
```diff
@@ -998,22 +1044,35 @@ const _gameUpdate = (timestamp, timeDiff) => {
...
-        _endUse();
+        _endUse(player);
       }
 
+      if(useAction.animationCombo.length > 0) {
+        const animation = getAnimationDuration(useAction.animationCombo[useAction.index]);
+        if(animation.length > 0) {
+          const animationDuration = animation[0].duration;
+          const useTime = player.actionInterpolants.use.get();
+          if (useTime / 1000 >= 0.6) {
```
patriboz, line 1064: This is a hack and I'm not sure if we want to do it this way, but if we let the whole animation run if `(useTime / 1000 >= animationDuration)`, combat will be quite slow.

constants.js
```diff
@@ -140,7 +140,7 @@ export const defaultMaxId = 8192;
 
 export const defaultMusicVolume = 0.35;
 
-export const voicePacksUrl = `https://webaverse.github.io/voicepacks/all_packs.json`;
+export const voicePacksUrl = `http://localhost/voicepacks/all_packs.json`;
```
gonnavis, line 143: If ready for review, pls turn off PR's "draft" state, and clean these testing codes.

scenes/test.scn
```diff
@@ -0,0 +1,40 @@
...
+      "quaternion": [ 0, 0.7071067811865475, 0, 0.7071067811865475 ],
+      "type": "application/npc",
+      "content": {
+        "name": "Ann",
+        "avatarUrl": "/avatars/ann.vrm",
+        "voice": "Maud Pie",
+        "bio": "13/F Lisk Witch (black mage). She is 100% quirky and is always creating crazy new potions. She has a dark streak and is known to torment her teachers when they give her bad grades.",
+        "wear": [
+          {
+            "start_url": "http://localhost/silsword/"
```
gonnavis, line 32: Add a new `.scn` such as `combat.scn`, put into `scenes.json`.
Using formal links instead of `localhost`.

gonnavis:
I encountered error when npc first try to use sword, any idea ?
image
image

patriboz:
> I encountered this error when npc first try to use sword, any idea ?
> 
> image image

I added logic for weapons without combos, so this should be fixed now.

avaer:
These combo animations are still not finished (I think it's been bad like this for a year now).

zenkale:
@patriboz I have identified 3 issues:
1-once we have a sword/silsword and we click for attack, One click makes a full swing happen(whole animation plays before player stops). Shouldn't it be like before? The attack should only be as long as we are clicking the left click? This make the combat very weird.

https://user-images.githubusercontent.com/111510128/188224982-c4b9545c-974f-4b98-bb23-b26396f6e0f0.mp4

2-Since we click once to attack, full attack animation plays before character is free again, we have no control over our character. We can not move it, we can not change its direction. As you can see in the video below, once we click attack it once, we are unable to change the direction, walk, run until that animation is finished.

https://user-images.githubusercontent.com/111510128/188225023-73a39dfa-0a6f-43cd-8982-33d2aaedbecf.mp4

3-Once we have equipped an object and we drop it, after that the activator "E" doesn't appears on the equip able objects. As it can be seen in the video.

https://user-images.githubusercontent.com/111510128/188225096-ccf3614e-59bc-4737-a5b6-0664cc9bea80.mp4

patriboz:
@zenkale 
**Issues 1 & 2:** In this branch the sword still has the (old) combo as one animation (3 swings). This is the correct behavior for a single animation, it is just weird that it takes so long until the controls are enabled again. We already split this combo into 3 separate animations, similar to silsword, I will implement these animations after my current priority tasks.

**Issues 3 & 4:** I will work on these issues.

Thanks!

lalalune:
@patriboz where we at with this?

Issue #3573, opened 2022-08-09
SMG shooting animation


avaer:
The `uzi` in `treehouse.scn` behaves as a strange unfinished combination of pistol and machine gun.

- the recoil should be constant when firing
- the smoke should come out of the tip

Issue #3153, opened 2022-06-21
QA Checklist for Z-targetting branch


alisaad673:
**UI:**

- Character panel
- Login
- World tokens
- Settings
- mic toggle
- voicer
- Multiplayer room
- Avatar quality settings
- AI settings

**Gameplay:**

- Scene switching
- walk/run/dash/fly/jump/crouch/dance
- combat(swords, pistol, bow)
- Drinking/Eating
- vehicles
- Avatar switching
- equipping items
- Dropping items
- Switching items
- multiplayer
- Chat
- Emotions(character panel)
- Tokens select(character panel)
- Equipping tokens
- Voice packs
- Voice endpoints
- Npcs
- Chat with npcs with goose and a121 AI
- Open AI key in settings and Chat with npcs
- Drag and drop avatar
- Drag and drop glbs
- World editor(Positions, quaternions, scale)
- Stats panel
- Character select( Equipping vrms)
- Character Previews
- Character intro dialogues in Character select panel
- Terrain
- Silk Grass
- Plants
- Silkworm Creatures
- Focus Mode( middle click)
- Fixed Focus mode( Double Click E)
- Z-lock
- Wind
- Drop( p)
- Claims inventory
- Respawning and switching to npc( G)

**Login:**

- Discord login
- metamask login
- Discord assets
- Metamask assets
- Drag and drop assets

Console log:
- No new Errors or warnings

- record performance benchmarks
- ensure no performance drops

Will keep adding more things

Issue #2855, opened 2022-04-21
SMGs need recoil


avaer:
Our pistol activates avatar recoil but our machine pistols/guns do not (there is only recoil on the first shot).

I think we need a new animation mode specifically for this recoil case.

avaer:
We can probably flag this as an action with `ik === 'smg'`, as opposed to the current `ik === 'pistol'`.

PR #3666, opened 2022-08-26
fix recoil effect for smg/uzi

intelliverse777:
## Describe your changes
I have added recoil effect for SMG and Uzi.
- Pick up SMG or Uzi
- Left click to shoot.
issue [3573](https://github.com/webaverse/app/issues/3573)
issue [2855](https://github.com/webaverse/app/issues/2855)

## Screenshots and/or video

https://user-images.githubusercontent.com/103969701/186907221-8231867e-ebc9-43ed-a6be-c892f5e9a731.mp4

- [ ] I have playtested and intentionally tried to find error cases but couldn't
- [x] I have completed the entire [QA checklist](https://github.com/webaverse/app/issues/3153) on my PR

### DIFF
--- a/character-physics.js
+++ b/character-physics.js
@@ -377,17 +377,17 @@ class CharacterPhysics {
     _updateFakeHands();
 
     const _updatePistolIkAnimation = () => {
-      const kickbackTime = 300;
       const kickbackExponent = 0.05;
       const fakeArmLength = 0.2;
-
+      const smgUse = !!useAction && useAction.ik === 'smg';
+      const kickbackTime = smgUse ? 100 : 300;
       const pistolUse = !!useAction && useAction.ik === 'pistol';
-      // console.log('got use action', !!pistolUse, useAction?.ik);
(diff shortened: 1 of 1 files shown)


avaer:
This logic seems a big mess, confusing pistols and smgs. Can it be cleaned up?

intelliverse777:
> This logic seems a big mess, confusing pistols and smgs. Can it be cleaned up?

I thought It needed to add recoil effect by using recoil effect of the pistol.
Could you please explain the recoil of the SMG that you want?
Thank you. ^^

avaer:
> I thought It needed to add recoil effect by using recoil effect of the pistol.

Yeah that's probably not the best way to do it.

I think you should think of each weapon IK as a different class of animation which is defined declaratively.

It's not good to mix all of this code in `if` statements and local variables. The better way would be to have an object which has keys like `pistol`, `smg`, etc, and maps to a mixer function or mixer class which will handle the animation for that type of weapon. And then all of the state would be isolated to local members and an `update` function. Does that make sense?

PR #3667, opened 2022-08-26
sonic boom trail issue

tcm390:
issue:

https://user-images.githubusercontent.com/60634884/186944727-c8fa2823-4bd4-4a87-89e5-1dc4a932ea8c.mp4

Result:

https://user-images.githubusercontent.com/60634884/186944754-8b493e04-9399-4488-9d61-70f9c12695f3.mp4


avaer:
The functionality looks good but:

- This PR changes textures (why is it not another PR?)
- The code is unreadable due to having so many shaders so cannot be reviewed
- Need to check that there is no performance regressions

avaer:
Actually I'd like to merge this anyway. But it's marked as a Draft. Could you clarify @tcm390?

tcm390:
> The functionality looks good but:
> 
> * This PR changes textures (why is it not another PR?)
> * The code is unreadable due to having so many shaders so cannot be reviewed
> * Need to check that there is no performance regressions

Got it would check the performance regression

tcm390:
I used h panel to test it. I think there is no performance regressions(inspect the index.js) and also it eliminate the hiccup on the first time use of sonicBoom

Result:

https://user-images.githubusercontent.com/60634884/187049307-307cea8c-2982-4d36-9fc3-0f14b6e1c9ff.mp4

old one:


https://user-images.githubusercontent.com/60634884/187049311-a52a45cf-e060-4457-b80b-374416973b0c.mp4

avaer:
`H` can never be used to test performance, because the probes reduce performance.

Issue #3669, opened 2022-08-28
Wind API is bad


avaer:
This API convention doesn't follow any of the React, metaversefile, or Webaverse conventions:

```
  getWinds() {
    return world.winds;
  },
  setWinds(wind) {
    world.winds.push(wind);
  },
  ```

avaer:
This probably requires totum updates.

PR #3671, opened 2022-08-28
Wind API is bad

tcm390:
Can we just set winds array in `metaversefile-api.js` like mirrors?

related: https://github.com/webaverse/app/issues/3669

result:

https://user-images.githubusercontent.com/60634884/187091366-c22c478b-c4e7-4393-83b6-2f96022ca63c.mp4

### DIFF
--- a/metaversefile-api.js
+++ b/metaversefile-api.js
@@ -354,6 +354,7 @@ let recursion = 0;
 let wasDecapitated = false;
 // const apps = [];
 const mirrors = [];
+const winds = [];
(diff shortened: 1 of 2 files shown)


avaer:
This isn't really an improvement. If the wind is controllable it should not have bare methods like `setWinds`. If anything it should go through a manager like everything else.

`useEnvironmentManager`?

tcm390:
Closed. Handle this issue by this PR: https://github.com/webaverse/app/pull/3673

Issue #3659, opened 2022-08-21
Redo sword grunts


avaer:
The sword grunts were implemented in the `silsword` app, which is the wrong place: https://github.com/webaverse/silsword/blob/df1cbb2cfb9a8f7eb9967d9159003d6c4e73a84e/index.js#L698

The character shape will change, which it did upstream, and now the silsword crashes, so I removed the grunt.

The grunt should occur, like the footsteps and jump, in response to actions on the avatar, implemented in the character controller system.

@tcm390 https://github.com/webaverse/silsword/blame/df1cbb2cfb9a8f7eb9967d9159003d6c4e73a84e/index.js#L698

avaer:
https://github.com/webaverse/silsword/commit/0b5316160295ca25aff0711fbbcd1edf04c5d4e1

patriboz:
Fix: https://github.com/webaverse/app/pull/3694

PR #3674, opened 2022-08-30
equip sound + sword grunt + sword sound 

tcm390:
1. Play the equip/drop sound in `metaverse_components/wear.js`  instead of `character-sfx.js`
(Should we just only play the equip sound for localPlayer?)
2. Because some apps like sword do not have the file `index.js` to play the sword whoosh, can we create a metaverse components `use` for the app to handle the sound event?
3. In animation-baker.js, baked the sword velocity to slash animation. And play the sword grunt and emit meleewhoosh event according to the velocity. (instead of using constant)
4. And for the redo sword animation, we haven't had the new sword animation so I just change sword animation to animationCombo in this PR: https://github.com/webaverse/sword/pull/2

Sibling:
https://github.com/webaverse/sword/pull/2

related:
https://github.com/webaverse/app/issues/3659

Result:

https://user-images.githubusercontent.com/60634884/187808725-518c0de0-fa6d-43c2-aff2-49e407a16fe5.mp4

### DIFF
--- a/animations-baker.js
+++ b/animations-baker.js
@@ -553,6 +553,8 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
           bones.push(bone);
         }
       });
+      // console.log('got bones', bones.map(b => b.name));
+      const rootBone = bones.find(b => b.name === 'mixamorigRightShoulder');; // not really a bone
       const leftHandBone = bones.find(b => b.name === 'mixamorigLeftHand');
       const rightHandBone = bones.find(b => b.name === 'mixamorigRightHand');
       const epsilon = 0.2;
@@ -617,6 +619,7 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
             bone.quaternion.copy(bone.initialQuaternion);
           }
         }
+        rootBone.updateMatrixWorld(true);
         const fbxScale = 100;
         const leftHand = new THREE.Vector3().setFromMatrixPosition(leftHandBone.matrixWorld).divideScalar(fbxScale);
         const rightHand = new THREE.Vector3().setFromMatrixPosition(rightHandBone.matrixWorld).divideScalar(fbxScale);
@@ -635,6 +638,7 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
         lastRightHandPos.set(rightHand.x, rightHand.y, rightHand.z);
       }
 
+      
       console.log('got', leftHandDeltas, rightHandDeltas, maxLeftDeltaIndex, maxRightDeltaIndex);
       
       return {
--- a/animations-baker.js
+++ b/animations-baker.js
@@ -553,8 +553,6 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
           bones.push(bone);
         }
       });
-      // console.log('got bones', bones.map(b => b.name));
-      const rootBone = bones.find(b => b.name === 'mixamorigRightShoulder');; // not really a bone
       const leftHandBone = bones.find(b => b.name === 'mixamorigLeftHand');
       const rightHandBone = bones.find(b => b.name === 'mixamorigRightHand');
       const epsilon = 0.2;
@@ -619,7 +617,6 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
             bone.quaternion.copy(bone.initialQuaternion);
           }
         }
-        rootBone.updateMatrixWorld(true);
         const fbxScale = 100;
         const leftHand = new THREE.Vector3().setFromMatrixPosition(leftHandBone.matrixWorld).divideScalar(fbxScale);
         const rightHand = new THREE.Vector3().setFromMatrixPosition(rightHandBone.matrixWorld).divideScalar(fbxScale);
--- a/animations-baker.js
+++ b/animations-baker.js
@@ -639,7 +639,7 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
       }
 
       
-      console.log('got', leftHandDeltas, rightHandDeltas, maxLeftDeltaIndex, maxRightDeltaIndex);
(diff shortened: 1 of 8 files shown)


rondoor124:
I checked the code changes, it seems okay to merge.

character-sfx.js
```diff
@@ -342,8 +351,49 @@ class CharacterSfx {
...
+    // combo
+    const dispatchComboSoundEvent = (soundIndex) =>{
+      this.dispatchEvent(new MessageEvent('meleewhoosh', {
+        data: {
+          index: soundIndex
+        },
+      }));
+    }
+    const _handleCombo = () => {
+      if (this.player.hasAction('use') && this.player.getAction('use').behavior === 'sword' && this.player.getAction('use').animationCombo) {
```
avaer, line 364: Why look it up 3 times per `if`?

character-sfx.js
```diff
@@ -342,8 +351,49 @@ class CharacterSfx {
...
+          const handDeltas = this.player.getAction('use').boneAttachment === 'leftHand' ? animationIndices.rightHandDeltas : animationIndices.leftHandDeltas;
+          const maxDeltaIndex = this.player.getAction('use').boneAttachment === 'leftHand' ? animationIndices.maxRightDeltaIndex : animationIndices.maxLeftDeltaIndex;
+          
+          const ratio = (timeSeconds - this.swordComboStartTime) / animation.duration;
+          if (ratio <= 1 && !this.alreadyPlayComboSound) {
+            const index = Math.floor(ratio * handDeltas.length);
+            if (index > maxDeltaIndex) {
+              this.alreadyPlayComboSound = true;
+              this.playGrunt('attack');
+              const soundIndex = this.player.avatar.useAnimationIndex * 4 + Math.floor(Math.random() * 4);
```
avaer, line 384: Why `4`? Don't include random constants in the middle of the code.

character-sfx.js
tcm390: [clean code](https://github.com/webaverse/app/pull/3674/commits/9954dfb8d6a95d9adbf0fc9cb796d8e5a7e9b722)

character-sfx.js
tcm390: got it
[remove constsant in combo sound selection](https://github.com/webaverse/app/pull/3674/commits/9af4f0458f09df5eeb9d736a41cba5c5a50b9956)

metaverse_components/use.js
```diff
@@ -0,0 +1,45 @@
+import * as THREE from 'three';
```
avaer, line 1: Why break out sound only into the component? There is a lot of `use` behavior we can move to the `use` component, but this simply makes the code worse for now since it fragments the behavior even more.

If we are doing to break out `use` into a component, we should do it for all use behavior, not just sounds.

metaverse_components/use.js
```diff
@@ -0,0 +1,45 @@
...
+    if(using){
+      const index = e.data.index;
+      const soundRegex = new RegExp(`^combat/sword_slash${index}-[0-9]*.wav$`);
+      const soundCandidate = soundFiles.combat.filter(f => soundRegex.test(f.name));
+      const audioSpec = soundCandidate[Math.floor(Math.random() * soundCandidate.length)];
+      sounds.playSound(audioSpec);
+    }
+  }
+  const _unwear = () => {
+    if (component.behavior === 'sword') {
```
avaer, line 29: This should not be an `if`, but a declarative object.

character-sfx.js
```diff
@@ -342,8 +351,50 @@ class CharacterSfx {
...
+    // combo
+    const dispatchComboSoundEvent = (soundIndex) =>{
+      this.dispatchEvent(new MessageEvent('meleewhoosh', {
+        data: {
+          index: soundIndex
+        },
+      }));
+    }
+    const _handleCombo = () => {
+      if (this.player.hasAction('use') && this.player.getAction('use').behavior === 'sword') {
```
avaer, line 364: Why is this tripled?

character-sfx.js
```diff
@@ -62,22 +72,29 @@ const _getActionFrameIndex = (f, frameTimes) => {
...
-class CharacterSfx {
+class CharacterSfx extends EventTarget{
   constructor(player) {
+    super();
     this.player = player;
 
     this.lastJumpState = false;
     this.lastStepped = [false, false];
     this.lastWalkTime = 0;
+
```
avaer, line 83: Bad spacing.

character-sfx.js
```diff
@@ -62,22 +72,29 @@ const _getActionFrameIndex = (f, frameTimes) => {
...
+   
+    
+    this.lastSwordComboName = null;
```
avaer, line 86: There are now so many random state keys on this class that it needs to be broken up further.

character-sfx.js
```diff
@@ -342,8 +351,50 @@ class CharacterSfx {
...
+              dispatchComboSoundEvent(soundIndex);
+            }
+          }
+        }
+        this.lastSwordComboName = comboAnimationName;
+      }
+      
+    };
+
+    _handleCombo();
```
avaer, line 396: This is poor quality unstructured code. The rest of the file has the exact same problem of being random unstructured `if` statements with a huge state vector, but it's gotten to the point where it is difficult for anyone to review this or even understand what is going on here.

IMO this file needs a complete redesign under some sort of structured coding theory before anyone can review it.

animations-baker.js
```diff
@@ -534,6 +534,122 @@ const {CharsetEncoder} = require('three/examples/js/libs/mmdparser.js');
     });
     // console.log('got animations', animations);
 
+    const comboAnimationNames = [
```
avaer, line 537: This code is completely unabstracted and unreadable.

avaer: There are no comments or functions here, so it's not clear what this is supposed to do or why it would be correct.

avaer (changes requested):
The code quality on this PR and associated files is too poor to properly review. There is little function abstraction, with massive code blocks, repeated lookups with little variable reuse, and massive state vectors on classes that are randomly accessed.

Most of the files this touches need a structured rewrite before this can even be reviewed for correctness.

PR #3686, opened 2022-09-02
Lazy load inventory

NPatel10:
## Describe your changes
In this PR Below changes are added
1. Inventory, Season, Land and Account panels will load lazily. Background of these panels will load at the start but the items of these panels will only load upon opening those panels. Reason for this is to keep the opening animation of these panels. Caching mechanism was already there and no changes have been done to it.
2. [Optional] Currently all the items in the inventory, Season and Account panels are being load simultaneously so in the second commit this behavior has been changed. Now only predefined number of items will load at the same time and rest will wait in the queue for previous ones to complete their loading process.

## Further Improvements
1. Currently upon receiving SpriteSheet an interval starts to render those sprite into the canvas. When the amount of items increases these intervals keeps adding up and disturbs the main rendering because they are not in sync. Maybe a system can be develop keep all these canvases and render them on certain tick.
2. Character Selection and Character Panel also has same behavior of loading at the start so it can be altered.
3. Inventory does not has scrolling currently.

1. Open the scene `https://local.webaverse.com/`
4. Press `Tab` and load the inventory. Inventory will start to load. previously it will be loaded from start.
5. Press `Tab` to close the inventory.
6. Press `Tab` again to open inventory and check that items are already loaded from cache and do not load second time.
7. If there are more then 5 items in the inventory then first 5 will load first then next 5 then another 5 and so on.

N/A

- [x] I have completed the entire [QA checklist](https://github.com/webaverse/app/issues/3153) on my PR

### DIFF
--- a/spritesheet-manager.js
+++ b/spritesheet-manager.js
@@ -19,49 +19,43 @@ class AssetLoadManager {
 
   maxLoaderCount = 5;
   loaderPool = [];
-  intervalID = 0;
-
-  schedulerInterval = 10;
 
   constructor() {
     for (let i = 0; i < this.maxLoaderCount; i++) {
-      this.loaderPool.push(new AssetLoader())
+      this.loaderPool.push(new AssetLoader());
     }
   }
 
-  poll() {
-    this.intervalID = setInterval(() => {
-      if (this.getQueue.length > 0) {
-        for (let i = 0; i < this.loaderPool.length; i++) {
-          if (this.loaderPool[i].isIdle && this.getIndex < this.getQueue.length - 1) {
-            this.loaderPool[i].load(this.getQueue[++this.getIndex]);
-          }
+  load() {
+    if (this.getQueue.length > 0) {
+      for (let i = 0; i < this.loaderPool.length; i++) {
+        if (this.loaderPool[i].isIdle && this.getIndex < this.getQueue.length - 1) {
+          (async () => {
+            await this.loaderPool[i].load(this.getQueue[++this.getIndex]);
+            this.load();
+          })();
         }
-      } else if (this.pushQueue.length <= 0) {
-        clearInterval(this.intervalID);
-        this.intervalID = 0;
       }
 
       if (this.getIndex > -1) {
         this.getQueue.splice(0, this.getIndex + 1);
         this.getIndex = -1;
       }
-
-      const t = this.pushQueue
+    } else if (this.pushQueue.length > 0) {
+      const t = this.pushQueue;
       this.pushQueue = this.getQueue;
       this.getQueue = t;
-    }, this.schedulerInterval);
+    }
   }
 
-  putInQueue(promise) {
-    this.pushQueueLock = this.pushQueueLock.then(() => {
+  async putInQueue(promise) {
+    this.pushQueueLock = (async () => {
+      await this.pushQueueLock;
       this.pushQueue.push(promise);
-      if (this.intervalID === 0) {
-        this.poll();
-      }
+      this.load();
     });
 
-    return this.pushQueueLock;
+    return this.pushQueueLock();
   }
 }
 
@@ -100,7 +94,10 @@ class SpritesheetManager {
 
     if (!spritesheet) {
       spritesheet = await new Promise(async (resolve) => {
-        this.loadManager.putInQueue(() => this.getSpriteSheetForAppUrlInternal([ appUrl, opts ]).then(resolve))
+        this.loadManager.putInQueue(async () => {
+          const spriteSheet = await this.getSpriteSheetForAppUrlInternal([ appUrl, opts ]);
+          resolve(spriteSheet);
+        });
       });
 
       this.spritesheetCache.set(appUrl, spritesheet);
--- a/spritesheet-manager.js
+++ b/spritesheet-manager.js
@@ -2,6 +2,11 @@ import {createObjectSprite} from './object-spriter.js';
 import offscreenEngineManager from './offscreen-engine-manager.js';
 
 class SpritesheetManager {
+  maxConcurrentLoading = 5;
+  currentlyLoading = 0;
+  isLoaderRunning = false;
+  waitQueue = [];
+
   constructor() {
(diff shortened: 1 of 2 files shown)


spritesheet-manager.js
```diff
@@ -29,13 +34,43 @@ class SpritesheetManager {
...
+  }
+
+  async loadSpriteSheetFromQueue() {
+    if (this.currentlyLoading >= this.maxConcurrentLoading) return;
+
+    this.isLoaderRunning = true;
+
+    for (let i = 0; i < this.maxConcurrentLoading; i++) {
+      this.currentlyLoading++;
+      this.loadSpriteSheet((i+1)).then(() => {
```
avaer, line 56: Since this is already async (for apparenlt no reason), this could be `await`.

spritesheet-manager.js
```diff
@@ -29,13 +34,43 @@ class SpritesheetManager {
...
+    if (spritesheet) return spritesheet
+
+    return new Promise((resolve) => {
+      this.waitQueue.push({ url: appUrl, opts, resolve })
+      if (!this.isLoaderRunning) this.loadSpriteSheetFromQueue()
+    })
+  }
+
+  async loadSpriteSheetFromQueue() {
+    if (this.currentlyLoading >= this.maxConcurrentLoading) return;
```
avaer, line 50: For this counted dispatch technique, my philosophy is you have to decide who is doing the counting and the dispatching check. Is it the caller or the method?

Here the check seems to be split across the caller and the method, which is not good. Should either do all of the checking here or outside (possibly in its own method).

avaer: Talking about this: https://github.com/webaverse/app/pull/3686/files#diff-7c16ad233a6efe359517c683cb8107f3f2e01df21e1b3084350926627486656aR64

NPatel10: Very helpful insight. I updated the logic accordingly and now there is only one method which handles checks.

spritesheet-manager.js
```diff
@@ -29,13 +33,35 @@ class SpritesheetManager {
...
   async getSpriteSheetForAppUrlAsync(appUrl, opts) {
-    let spritesheet = this.spritesheetCache.get(appUrl);
-    if (!spritesheet) {
-      spritesheet = await this.getSpriteSheetForAppUrlInternal([appUrl, opts]);
-      this.spritesheetCache.set(appUrl, spritesheet);
+    const spritesheet = this.spritesheetCache.get(appUrl);
+
+    if (spritesheet) return spritesheet;
+
+    return new Promise((resolve) => {
```
avaer, line 42: IMO not the best to have polymorphic functions like this; it's more complicated to read and make strongly typed.

It could just `await` the result and return it at the top level.

spritesheet-manager.js
```diff
@@ -29,13 +33,35 @@ class SpritesheetManager {
...
+      this.waitQueue.push({ url: appUrl, opts, resolve });
+      this.loadSpriteSheetFromQueue();
+    });
+  }
+
+  loadSpriteSheetFromQueue() {
+    if (this.currentlyLoading >= this.maxConcurrentLoading || this.waitQueue.length <= 0) return;
+
+    const loopCount = Math.min(this.maxConcurrentLoading, this.waitQueue.length);
+    for (let i = 0; i < loopCount; i++) {
```
avaer, line 52: It could just be a while.

spritesheet-manager.js
```diff
@@ -29,13 +33,35 @@ class SpritesheetManager {
...
+      this.currentlyLoading++;
+      const data = this.waitQueue.splice(0, 1)[0];
+
+      this.getSpriteSheetForAppUrlInternal([ data.url, data.opts ]).then((spritesheet) => {
```
avaer, line 56: For consistency, we aim for `async` wrapping where possible rather than callbacks. I believe VSCode also pushes you to do it.

spritesheet-manager.js
```diff
@@ -29,13 +33,35 @@ class SpritesheetManager {
...
+    if (spritesheet) return spritesheet;
+
+    return new Promise((resolve) => {
+      this.waitQueue.push({ url: appUrl, opts, resolve });
+      this.loadSpriteSheetFromQueue();
+    });
+  }
+
+  loadSpriteSheetFromQueue() {
+    if (this.currentlyLoading >= this.maxConcurrentLoading || this.waitQueue.length <= 0) return;
```
avaer, line 49: Another early return.

spritesheet-manager.js
```diff
@@ -29,13 +33,35 @@ class SpritesheetManager {
...
   }
+
   async getSpriteSheetForAppUrlAsync(appUrl, opts) {
-    let spritesheet = this.spritesheetCache.get(appUrl);
-    if (!spritesheet) {
-      spritesheet = await this.getSpriteSheetForAppUrlInternal([appUrl, opts]);
-      this.spritesheetCache.set(appUrl, spritesheet);
+    const spritesheet = this.spritesheetCache.get(appUrl);
+
+    if (spritesheet) return spritesheet;
```
avaer, line 40: In the rest of the codebase, early returns are not preferred (instead we nest ifs), and curlies are non-optional.

NPatel10: I updated the code and changed the logic for loading assets. Now it creates predefined number of loaders and poll them to check whether they are idle or not. If they are then assign them tasks to load the asset. If there is no loader which is idle then then loading of the assets wait in the queue.

I also changed my code style of early return to nested ifs.

Worked on the race conditions and try to mitigate them.

spritesheet-manager.js
```diff
@@ -1,8 +1,73 @@
...
+
+  schedulerInterval = 10;
+
+  constructor() {
+    for (let i = 0; i < this.maxLoaderCount; i++) {
+      this.loaderPool.push(new AssetLoader())
+    }
+  }
+
+  poll() {
```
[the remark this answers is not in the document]
NPatel10: Removed the polling mechanism and added async wrapper

PR #3712, opened 2022-09-15
Tilde grab bugfixes

patriboz:
This PR fixes the following issues when editing scenes:
- grabbed object transform out of sync
- collisions
- pivot point (apparent when entering 3rd person mode)

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -1136,38 +1136,39 @@ class LocalPlayer extends UninterpolatedPlayer {
     });
   }
   grab(app, hand = 'left') {
-    let position   = null,
-        quaternion = null;
+    if(this instanceof LocalPlayer) {
+      let position   = null, quaternion = null;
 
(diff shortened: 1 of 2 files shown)


jakezira (commented):
Need to optimize more.

character-controller.js
```diff
@@ -1136,38 +1136,39 @@ class LocalPlayer extends UninterpolatedPlayer {
     });
   }
   grab(app, hand = 'left') {
-    let position   = null,
-        quaternion = null;
+    if(this instanceof LocalPlayer) {
```
jakezira, line 1139: - Can player call grab when it's not local player?
- Even though if it's the case, you can check this with different code. Please reference how it did on existing code.

patriboz:
Made a new branch and PR: https://github.com/webaverse/app/pull/3714

PR #3717, opened 2022-09-17
Fix addConvexGeometry not setting the transform properly.

cyberneticocult:
Meshes that were using convexMesh type would have their highlighted mesh in edit mode appear in the wrong location while the actual collision hull in the physics scene would be in the right location.

physicsManager.addConvexGeometry() uses the original mesh passed as a parameter to set the transform. It should be the physicsMesh built by convertMeshToPhysicsMesh() that's used to set the transform.

### DIFF
--- a/physics-manager.js
+++ b/physics-manager.js
@@ -299,9 +299,9 @@ class PhysicsScene extends EventTarget {
   
     const physicsObject = _makePhysicsObject(
       physicsId,
-      mesh.position,
-      mesh.quaternion,
-      mesh.scale
+      physicsMesh.position,
+      physicsMesh.quaternion,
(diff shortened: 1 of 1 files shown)


lalalune:
ty for this, will review

memelotsqui:
I do believe that this pr is necessary, as no other collider uses the base mesh position/scale/rotation. They are all using the physics mesh data, and it is indeed fixing when importing hull meshes. The closest example to look at would be 
addGeometry(mesh), and it is actually using physicsMesh data:

image

Also it will be necessary for the totum omi_collider pr support to work with hulls: https://github.com/webaverse/totum/pull/154

PR #3723, opened 2022-09-22
Patriboz/fix transform issues

patriboz:
With this PR I intend to fix:
- a teleport issue, that was caused by `physicsScene.raycast(localVector, localQuaternion)` colliding with the grabbed object on the first frame, although physical actors are being disabled before.
- placing grabbed object based on the bounding box of its physical mesh.

Current master: https://youtu.be/uhOdCu5TQI8
This PR: https://youtu.be/Udk5qs859gk

### DIFF
--- a/character-controller.js
+++ b/character-controller.js
@@ -1136,31 +1136,39 @@ class LocalPlayer extends UninterpolatedPlayer {
     });
   }
   grab(app, hand = 'left') {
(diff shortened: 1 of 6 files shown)


patriboz:
This PR was merged into studios, so I close it here. https://github.com/webaverse-studios/app/pull/29

PR #3742, opened 2022-11-21
fix: fixed broken developer documentation link in README

JoeyKhd:
## Describe your changes

Link to developer documentation appeared to be broken.


lalalune (approved):

lalalune:
Good catch @JoeyKhd !

We'll be doing a 0.6.0 release and pushing that here soon which is a lot more developer friendly :)

